PR #2, opened 2025-09-11
docs: polish readme

sebastianbenz:
* add configuration section
* move user-data-dir to new concepts section

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ prototype!
 
 `chrome-devtools-mcp` exposes content of the browser instance to the MCP clients
 allowing them to inspect, debug, and modify any data in the browser or DevTools.
-Avoid sharing sensitive or personal information that you do want to share with
+Avoid sharing sensitive or personal information that you don't want to share with
 MCP clients.
 
 ## Requirements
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -128,20 +129,22 @@ For example, to launch the system-installed Chrome Canary pass `--channel=canary
...
+        "--headless=true",
+        "--isolated=true",
       ]
     }
   }
 }
 ```
 
-### User data directory
+The following options are supported when running the Chrome DevTools MCP server from an MCP client:
```
OrKoN: let's include any doc improvements in the CLI. Otherwise, this largely duplicates the CLI documentation.

sebastianbenz: Thought about this as well, not all CLI commands make sense for a MCP client. The CLI docs don't have links etc. I think in this case we should duplicate (the important) options to provide better guidance.

OrKoN: which ones do not make sense?

sebastianbenz: version, help

OrKoN: I think that is clear that those should not be used. I would suggest we do not duplicate the documentation. I'd suggest we put links into `Concepts` to explain the background information for those who might need it. For example, the duplicated list already misses executablePath and logFile that could be very relevant for MCP clients.

I think links are supported well enough by modern terminals:

Screenshot 2025-09-11 at 13 13 48

Alternatively, let's generate this list based on the CLI description?

sebastianbenz: > Alternatively, let's generate this list based on the CLI description?

Let's do that then

PR #3, opened 2025-09-11
ci: preconfigure CI jobs

Lightning00Blade:


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1 @@
+blank_issues_enabled: false
```
Lightning00Blade, line 1: Forces the use of templates

.github/dependabot.yml
```diff
@@ -0,0 +1,29 @@
+version: 2
```
Lightning00Blade, line 1: Until we this repo is private it will have no affect, after it will get picked up.

PR #8, opened 2025-09-11
ci: actual release-please

OrKoN:
disable npm publishing for now.


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,18 @@
...
+      - main
+
+permissions: read-all
+name: release-please
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: googleapis/release-please-action@v4
```
Lightning00Blade, line 13: Nit: use sha

OrKoN: we are not using it in puppeteer either. IIRC it was not dependabot updatable?

PR #20, opened 2025-09-15
ci: add node matrix

OrKoN:


.github/workflows/run-tests.yml
```diff
@@ -40,3 +43,13 @@ jobs:
 
       - name: Run tests
         run: npm run test
+
+  # Gating job for branch protection.
+  test-success:
```
Lightning00Blade, line 48: Can you add a name please

PR #37, opened 2025-09-16
feat: validate uids

OrKoN:
- generate snapshot IDs.
- change uid to be a string.
- validate that uid is not coming from a stale snapshot.

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -21,8 +21,14 @@ import path from 'node:path';
 import {listPages} from './tools/pages.js';
 
 export interface TextSnapshotNode extends SerializedAXNode {
-  id: number;
+  id: string;
   children: TextSnapshotNode[];
+}
+
+export interface TextSnapshot {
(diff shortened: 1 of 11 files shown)


src/McpContext.ts
```diff
@@ -222,35 +237,38 @@ export class McpContext implements Context {
...
+      idToNode.set(nodeWithId.id, nodeWithId);
       return nodeWithId;
     };
 
     const rootNodeWithId = assignIds(rootNode);
-    this.#textSnapshot = rootNodeWithId;
-    return rootNodeWithId;
+    this.#textSnapshot = {
+      root: rootNodeWithId,
+      snapshotId: String(this.#nextSnapshotId++),
```
Lightning00Blade: Having he syntax as this was a little bit hard to read can we extract it into a var
```ts
#nextSnapshotId = 0

... 
createTextSnapshot(){
const snaphshotId = this.#nextSnapshotId++;
}
```

PR #52, opened 2025-09-18
fix: support node20

OrKoN:
Initially, we started with the support of Node22+ but since Node20 is an LTS that is still in the maintenance mode it makes sense to support it.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -20,7 +20,7 @@ jobs:
           - windows-latest
           - macos-latest
         node:
-          - 22.12.0
+          - 20
           - 22
           - 23
           - 24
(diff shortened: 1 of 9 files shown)


OrKoN:
@jackfranklin what do you think we should do about the failing insight in node 20?

.github/workflows/run-tests.yml
```diff
@@ -34,19 +35,34 @@ jobs:
         uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
         with:
           cache: npm
-          node-version: ${{ matrix.node }}
+          node-version: 22
```
Lightning00Blade: Why do we need this?

OrKoN: Added a comment

Lightning00Blade (approved):

PR #73, opened 2025-09-23
fix: expose timeout when Emulation is enabled

Lightning00Blade:


src/McpResponse.ts
```diff
@@ -117,6 +117,9 @@ export class McpResponse implements Response {
     if (networkConditions) {
       response.push(`## Network emulation`);
       response.push(`Emulating: ${networkConditions}`);
+      response.push(
+        `Navigation timeout set to ${context.getNavigationTimeout()}`,
```
```suggestion
        `Navigation timeout set to ${context.getNavigationTimeout()} ms.`,
```

PR #78, opened 2025-09-23
chore: publish to MCP registry

OrKoN:
Includes server.json and the GitHub Actions steps.

### DIFF
--- a/.github/workflows/publish-to-npm-on-tag.yml
+++ b/.github/workflows/publish-to-npm-on-tag.yml
@@ -10,7 +10,7 @@ permissions:
   contents: read
 
 jobs:
-  publish:
+  publish-to-npm:
     runs-on: ubuntu-latest
     steps:
       - name: Check out repository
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-to-npm-on-tag.yml
```diff
@@ -38,3 +38,13 @@ jobs:
       - name: Publish
         run: |
           npm publish --provenance --access public
+
+      - name: Install MCP Publisher
```
Lightning00Blade, line 68: Please do it in a seprate Job so we can restart it if it fails.

OrKoN: it needs to happen after npm publish per documentation.

Lightning00Blade: We can always use an `needs: publish`

OrKoN: That's a good point.

Issue #87, opened 2025-09-23
Failure getting console messages


sebastianbenz:
MCP client (Gemini CLI) tried to close the page after getting console messages resulting in an error:

```
 
╭──────────────────────────────────────────────────────────────────╮
│  > Show the console messages for https://developers.chrome.com.  │
╰──────────────────────────────────────────────────────────────────╯

 ╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
 │ ✓  navigate_page (chrome-devtools MCP Server) {"url":"https://developers.chrome.com"}                                                │
 │                                                                                                                                      │
 │    # navigate_page response                                                                                                          │
 │    ## Pages                                                                                                                          │
 │    0: https://developer.chrome.com/ [selected]                                                                                       │
 ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
 ╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
 │ ✓  list_console_messages (chrome-devtools MCP Server) {}                                                                             │
 │                                                                                                                                      │
 │    # list_console_messages response                                                                                                  │
 │    ## Console messages                                                                                                               │
 ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯                      ✦ I navigated to https://developers.chrome.com and it appears there are no console messages on the page.                                                        ╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
 │ x  close_page (chrome-devtools MCP Server) {"pageIdx":0}                                                                             │                       │                                                                                                                                      │                       │    MCP tool 'close_page' reported tool error for function call: {"name":"close_page","args":{"pageIdx":0}} with response:            │
 │    [{"functionResponse":{"name":"close_page","response":{"error":{"content":[{"type":"text","text":"No page                          │                       │    selected"}],"isError":true}}}}]                                                                                                   │
 ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
✦ I was unable to close the page. It seems that there is no page selected.

╭───────────────╮
│  > Try again  │
╰───────────────╯

✦ I will try again.
 ╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
 │ x  navigate_page (chrome-devtools MCP Server) {"url":"https://developers.chrome.com"}                                                │
[...]

PR #90, opened 2025-09-23
fix: forbid closing the last page

OrKoN:
McpContext currently assumes that there is a page. Eventually we might support closing the browser by closing the last page but right now we do not really have responses that indicate that the browser was closed. So this PR helps the LLM understand that it is okay to keep the last page running (and there are platform-specific differences as to what happens if the last page is closed).

Ref https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/87

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -131,6 +131,16 @@ export class McpContext implements Context {
     this.#consoleCollector.addPage(page);
     return page;
   }
+  async closePage(pageIdx: number): Promise<void> {
+    if (this.#pages.length === 1) {
+      throw new Error(
+        'Unable to close the last page in the browser. It is fine to keep the last page open.',
+      );
+    }
+    const page = this.getPageByIdx(pageIdx);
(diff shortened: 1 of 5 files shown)


OrKoN:
@sebastianbenz  would it be useful as an intermediate solution?

src/McpContext.ts
```diff
@@ -131,6 +131,16 @@ export class McpContext implements Context {
     this.#consoleCollector.addPage(page);
     return page;
   }
+  async closePage(pageIdx: number): Promise<void> {
+    if (this.#pages.length === 1) {
+      throw new Error(
```
sebastianbenz, line 136: Should we throw an error if this is OK behavior?

OrKoN: I think it is fine, the model will get error message explaining the situation.

OrKoN:
Good idea, added.

sebastianbenz (approved):

PR #114, opened 2025-09-24
chore: add gemini extension json

OrKoN:


gemini-extension.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "chrome-devtools-mcp",
+  "version": "0.2.6",
```
sebastianbenz, line 3: Do we need to auto update this as well?

sebastianbenz (approved):

gemini-extension.json
OrKoN: yes, we would need to, right now we cannot auto-update anything during the release but also gemini installs directly from github and does not seem to use the versioning for anything.

OrKoN: gemini-cli also cannot install from a branch so I need to land this to be able to test.

PR #119, opened 2025-09-24
Chore: invalid sample json fix in readme

abhinavRai23:


OrKoN:
Thanks for flagging this! I have published the change in https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/122

Issue #125, opened 2025-09-24
Gemini shows error when trying to close the last page


sebastianbenz:
Gemini CLI will try to close the last open page and will report an error if it fails:

<img width="1096" height="122" alt="Image" src="https://github.com/user-attachments/assets/03af51f6-fd26-49bf-8a58-95f009ebce0a" />

Given that this is not an error, we probably should rethink throwing an exception in this case.

OrKoN:
It rather looks like an instruction following problem on the side of the MCP client. Does the following work better https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/127 ?

PR #127, opened 2025-09-24
fix: avoid reporting page close errors as errors

OrKoN:
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/125

### DIFF
--- a/src/tools/pages.ts
+++ b/src/tools/pages.ts
@@ -5,7 +5,7 @@
  */
 
 import z from 'zod';
-import {defineTool} from './ToolDefinition.js';
+import {CLOSE_PAGE_ERROR, defineTool} from './ToolDefinition.js';
 import {ToolCategories} from './categories.js';
 
 export const listPages = defineTool({
@@ -58,7 +58,15 @@ export const closePage = defineTool({
       ),
   },
   handler: async (request, response, context) => {
-    await context.closePage(request.params.pageIdx);
+    try {
+      await context.closePage(request.params.pageIdx);
+    } catch (err) {
+      if (err.message === CLOSE_PAGE_ERROR) {
+        response.appendResponseLine(err.message);
+      } else {
+        throw err;
+      }
+    }
     response.setIncludePages(true);
   },
 });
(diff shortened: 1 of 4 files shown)


src/tools/pages.ts
```diff
@@ -58,7 +58,15 @@ export const closePage = defineTool({
       ),
   },
   handler: async (request, response, context) => {
-    await context.closePage(request.params.pageIdx);
+    try {
+      await context.closePage(request.params.pageIdx);
+    } catch (err) {
+      if (err.message.startsWith('The last open page cannot be closed.')) {
```
sebastianbenz: This seems a bit brittle. Should we extract the partial string into a constant?

OrKoN: Agree, although it does not matter with the test coverage, updated anyway.

OrKoN:
@sebastianbenz could you please try it with gemini-cli before we land it?

sebastianbenz:
I, of course, could not reproduce the exact situation. I tried explicilty triggering it by calling "close page"  twice when only one page was open:

```

╭────────────────────╮
│  > Close the page  │
╰────────────────────╯

✦ Got it. Closing the page.
 ╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
 │ ✓  close_page (chrome-devtools MCP Server) {"pageIdx":1}                                                                             │
 │                                                                                                                                      │
 │    # close_page response                                                                                                             │
 │    ## Pages                                                                                                                          │
 │    0: about:blank [selected]                                                                                                         │
 ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
✦ Page closed.

╭────────────────────╮
│  > Close the page  │
╰────────────────────╯

✦ I can't close the last page.

Using: 1 MCP server (ctrl+t to view)
```

After the first call to close the page, the tab was closed and the `about:blank` page appeard (which is as expected). For the second call I would've expected the message to be: "No active page", but that's more a minor nit.

sebastianbenz (approved):
Thank you!

Issue #136, opened 2025-09-25
Pagination for network requests


OrKoN:
See https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/107

PR #107, opened 2025-09-24
feat: Add pagination and request type filtering to list_network_requests

OzTamir:
## Summary

This PR enhances the `list_network_requests` tool with two major features:
- **Pagination support** to handle large numbers of network requests efficiently
- **Request type filtering** to allow filtering by specific resource types

## Motivation

Playing around with  this MCP, I often found that calls to `list_network_requests` would frequently cause the LLM (in my case - Claude Code) hit token limits, resulting in errors like:
```
Error: MCP tool "list_network_requests" response (450039 tokens) exceeds maximum allowed tokens (25000). Please use pagination, filtering, or limit parameters to reduce the response size.
```

This made the tool unusable for pages with many network requests, which is common in modern web applications. The new pagination and filtering capabilities solve this by allowing users to:
- Limit the number of requests returned per call
- Filter to only specific types of requests they're interested in
- Navigate through results in manageable chunks

### Pagination Support
- Added `pageSize` parameter (max 100, defaults to 20) to control the number of requests returned
- Added `pageToken` parameter for pagination navigation
- Implemented pagination logic in `McpResponse.ts` and `networkUtils.ts`
- Added comprehensive pagination metadata in responses (nextPageToken, previousPageToken, startIndex, endIndex, total)

### Request Type Filtering
- Added `requestType` parameter to filter requests by resource type
- Supports filtering by 12 different resource types: document, stylesheet, image, media, font, script, xhr, fetch, prefetch, websocket, preflight, other
- Supports both single type and array of types for filtering
- Added robust type validation and sanitization

### Implementation Details
- Created new `networkUtils.ts` with comprehensive utility functions for filtering and pagination
- Updated `McpResponse.ts` to handle the new parameters and pass them to the pagination logic
- Enhanced `ToolDefinition.ts` to support the new parameter structure
- Added extensive test coverage for both features

### Documentation
- Updated tool reference documentation to reflect the new parameters
- Added clear descriptions for all new parameters with examples

## Testing
- Added comprehensive unit tests for pagination logic
- Added tests for request type filtering functionality
- Updated existing network tool tests to cover new functionality
- All tests pass and maintain backward compatibility

## Backward Compatibility
- All new parameters are optional, maintaining full backward compatibility
- Existing calls to `list_network_requests` without parameters continue to work as before
- Default behavior returns all requests when no pagination or filtering is specified

PR #145, opened 2025-09-25
feat: Add pagination list_network_requests

OzTamir:
## Summary

This PR enhances the `list_network_requests` tool with pagination support to handle large numbers of network requests efficiently. See [this issue](https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/136).

## Motivation

In my experience, the `list_network_requests` tool frequently hits LLM token limits on pages with many requests, making it unusable for modern web applications. I wanted to add pagination to allow agents to be more flexible and manage token limits on their own.

### Pagination Support
- Added `pageSize` parameter to limit requests per call
- Added `pageToken` parameter for navigation between pages
- Added pagination metadata in responses (nextPageToken, previousPageToken, startIndex, endIndex, total)

### Implementation Details
- **New utility**: `src/utils/pagination.ts` - generic pagination function
- **Enhanced McpResponse**: Added pagination options to `setIncludeNetworkRequests()`
- **Updated network tool**: Added pagination parameters to schema
- **Offset-based pagination**: Uses numeric tokens, handles invalid tokens gracefully

## Testing
- Comprehensive test coverage for pagination scenarios
- Tests for first page, subsequent pages, invalid tokens, and edge cases
- All existing tests continue to pass

## Backward Compatibility
- If no pagination parameters are provided, the MCP will return all requests (same as before)

### DIFF
--- a/src/tools/network.ts
+++ b/src/tools/network.ts
@@ -5,8 +5,8 @@
  */
 
 import z from 'zod';
-import { defineTool } from './ToolDefinition.js';
-import { ToolCategories } from './categories.js';
+import {defineTool} from './ToolDefinition.js';
+import {ToolCategories} from './categories.js';
 
 export const listNetworkRequests = defineTool({
   name: 'list_network_requests',
--- a/tests/McpResponse.test.ts
+++ b/tests/McpResponse.test.ts
@@ -3,10 +3,10 @@
  * Copyright 2025 Google LLC
  * SPDX-License-Identifier: Apache-2.0
  */
-import { describe, it } from 'node:test';
+import {describe, it} from 'node:test';
 import assert from 'assert';
 
-import { getMockRequest, html, withBrowser } from './utils.js';
+import {getMockRequest, html, withBrowser} from './utils.js';
 
 describe('McpResponse', () => {
   it('list pages', async () => {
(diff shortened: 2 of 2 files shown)


OzTamir:
@OrKoN this is the first PR out of the two in #107 (with your comments fixed 🙂) - let's get this through and I will build the request filtering on top of this and create an additional PR later.

src/tools/network.ts
```diff
@@ -15,9 +15,27 @@ export const listNetworkRequests = defineTool({
...
+  schema: {
+    pageSize: z
+      .number()
+      .int()
+      .positive()
+      .optional()
+      .describe(
+        'Maximum number of requests to return. When omitted, returns all requests.',
+      ),
+    pageToken: z
```
OrKoN: is there a reason to use a pageToken instead of a pageIdx? and make it non opaque so that client can jump to a particular page? I have done some testing and it seems the client gets often confused if you are asking to fetch a specific page with this API.

OrKoN: alternatively, could we change pageToken to be a number and indicate that it is not opaque and it means the index of the first item of the page?

OzTamir: Fair point - updated to use `pageIdx`.

OrKoN:
Thanks!

OrKoN (approved):

Issue #148, opened 2025-09-25
Problem to locate chrome.exe on windows 64 bits.


Prestaplugins:
**Describe the bug**
I'm trying to use chrome-devtools MCP with Cursor on windows 10 64 bits. I spent my full day trying to make it work, but I failed.
The problem is that it never finds chrome.exe, even when I use --executablePath option.

**To Reproduce**
Use cursor and windows 64 bits.
Add the MCP server using mcp.json like this :
```
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": [
        "chrome-devtools-mcp@latest",
        "--",
        "--executablePath",
        "C:\\Program Files (x86)\\Google\\Chrome\\Application\\chrome.exe"
      ]
    }
  }
}
```

Try this prompt : 
Using chrome-devtools MCP, check LCP for https://web.dev and stop on the 1st error. Show me the error.

Check the result :
"Could not find Google Chrome executable for channel 'stable' at 'C:\Program Files\Google\Chrome\Application\chrome.exe'."

**Expected behavior**
Use the executablePath option to locate chrome.exe

**Chrome version:**
140.0.7339.208

<img width="1575" height="594" alt="Image" src="https://github.com/user-attachments/assets/438f1d0e-726e-4113-a5df-26339247d097" />

sebastianbenz:
Can you please try escaoping the backlash in the path? `"C:\\Program Files (x86)\\Google\\Chrome\\Application\\chrome.exe"`?

Prestaplugins:
Yes of course I did, that's github who was removing it ;-)

sebastianbenz:
Hah, that would've been too easy. Does it work if you launch chrome-devtools mcp from the commandline?

Prestaplugins:
`npx chrome-devtools-mcp@latest -- --executablePath "C:\Program Files (x86)\Google\Chrome\Application\chrome.exe"`

Give this result :
chrome-devtools-mcp exposes content of the browser instance to the MCP clients allowing them to inspect,
debug, and modify any data in the browser or DevTools.
Avoid sharing sensitive or personal information that you do want to share with MCP clients.

It doesn't launch Chrome.

OrKoN:
Chrome is launched when the first tool call is being sent. Can you try the alternative format for executablePath: 

```
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": [
[...]

PR #150, opened 2025-09-25
fix: do not set channel if executablePath is provided

OrKoN:
Fixes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/148

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,0 +1,89 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import yargs from 'yargs';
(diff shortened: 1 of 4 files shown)


src/cli.ts
```diff
@@ -0,0 +1,89 @@
...
+};
+
+export function parseArguments(version: string, argv = process.argv) {
+  const yargsInstance = yargs(hideBin(argv))
+    .scriptName('npx chrome-devtools-mcp@latest')
+    .options(cliOptions)
+    .check(args => {
+      // We can't set default in the options else
+      // Yargs will complain
+      if (!args.channel && !args.browserUrl && !args.executablePath) {
```
OrKoN, line 66: Addition of `!args.executablePath` is the only functional change. The rest is moving code around for testing.

Issue #137, opened 2025-09-25
Filtering by resourceType() for network requests


OrKoN:
See https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/107

PR #162, opened 2025-09-26
feat: add network request filtering by resource type

OzTamir:
### Summary
This PR build upon #145 to also adds filtering by resource type to `list_network_requests` (Also see: #137 and #107).

### Motivation
Agents often need specific request types (e.g., scripts, stylesheets, images). Filtering reduces noise and improves performance.

### Changes
- **New parameter**: `resourceType` (array) to filter by resource types
- **Supported types**: all resource types supported by Puppeteer
- **Backward compatible**: when omitted, returns all requests

Filtering runs before pagination, so pagination applies to the filtered results.

### DIFF
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -7,18 +7,18 @@ import type {
   ImageContent,
   TextContent,
 } from '@modelcontextprotocol/sdk/types.js';
-import type { ResourceType } from 'puppeteer-core';
+import type {ResourceType} from 'puppeteer-core';
 
-import { formatConsoleEvent } from './formatters/consoleFormatter.js';
+import {formatConsoleEvent} from './formatters/consoleFormatter.js';
 import {
   getFormattedHeaderValue,
   getShortDescriptionForRequest,
   getStatusFromRequest,
 } from './formatters/networkFormatter.js';
-import { formatA11ySnapshot } from './formatters/snapshotFormatter.js';
-import type { McpContext } from './McpContext.js';
-import type { ImageContentData, Response } from './tools/ToolDefinition.js';
-import { paginate, type PaginationOptions } from './utils/pagination.js';
+import {formatA11ySnapshot} from './formatters/snapshotFormatter.js';
+import type {McpContext} from './McpContext.js';
+import type {ImageContentData, Response} from './tools/ToolDefinition.js';
+import {paginate, type PaginationOptions} from './utils/pagination.js';
 
 export class McpResponse implements Response {
   #includePages = false;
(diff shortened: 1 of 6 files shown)


src/tools/network.ts
```diff
@@ -33,11 +56,25 @@ export const listNetworkRequests = defineTool({
...
+        z.enum(
+          FILTERABLE_RESOURCE_TYPES as unknown as [
+            ResourceType,
+            ...ResourceType[],
+          ],
+        ),
+      )
+      .optional()
+      .describe(
+        'Filter requests to only return requests of the specified resource types. When omitted, returns all requests.',
```
```suggestion
        'Filter requests to only return requests of the specified resource types. When omitted or empty, returns all requests.',
```

Let's change the logic to return everything for the empty array. I tested with the MCP inspector and it sends an empty array by default for this. I suspect that models might try to do this as well. Since there is no use cases for returning no requests with resourceType = [], let's change the logic to treat this as equal to all available resource types. Note let's special case it to not use actual fitlering logic in case more resource types are added in Puppeteer. I left suggestions to implement this but let's also add a test.

src/tools/network.ts
```diff
@@ -33,11 +56,25 @@ export const listNetworkRequests = defineTool({
       .describe(
         'Page number to return (0-based). When omitted, returns the first page.',
       ),
+    resourceType: z
```
```suggestion
    resourceTypes: z
```

src/McpResponse.ts
```diff
@@ -179,13 +193,25 @@ Call browser_handle_dialog to handle it before continuing.`);
 
     response.push(...this.#getIncludeNetworkRequestsData(context));
 
-    if (this.#includeNetworkRequests) {
-      const requests = context.getNetworkRequests();
+    if (this.#networkRequestsOptions?.include) {
+      let requests = context.getNetworkRequests();
+
+      // Apply resource type filtering if specified
+      if (this.#networkRequestsOptions.resourceTypes) {
```
```suggestion
      if (this.#networkRequestsOptions.resourceTypes?.length) {
```

src/tools/network.ts
```diff
@@ -33,11 +56,25 @@ export const listNetworkRequests = defineTool({
...
+      .array(
+        z.enum(
+          FILTERABLE_RESOURCE_TYPES as unknown as [
```
Lightning00Blade: Let's type the Array correctly in the definition and not use satisfies
```ts
const FILTERABLE_RESOURCE_TYPES: Readonly[
  ResourceType,
  ...ResourceType[],
]> = [
...
];
```

src/tools/network.ts
OzTamir: Good call - added your suggestion + testcase.

src/tools/network.ts
OzTamir: Fixed 👍🏻

OrKoN (approved):

PR #173, opened 2025-09-26
Add Visual Studio installation instructions

madskristensen:
Added installation instructions for Visual Studio.

> Note, I'm a PM on the Visual Studio team and this way of installing MCP servers into VS is the recommended way


sebastianbenz:
Thanks for adding this! The link doesn't work for me, I saw that it redirects to a vs code specific url pattern which my mac doesn't seem to handle (even though I have vs code installed). This might be an issue with my machine, but in any case I'd recommend also describing the manual steps required (similar to what we did for cursor).

madskristensen:
@sebastianbenz It's not for VS Code but for Visual Studio - the C#/C++ IDE for Windows. You need Visual Studio installed for the protocol handler to work

sebastianbenz (approved):
Sorry! Please ignore my ignorance.

PR #179, opened 2025-09-27
Fix: update error message to use take_snapshot instead of browser_snapshot

yasirusman85:
Updated error message to reflect correct tool name (take_snapshot) instead of outdated browser_snapshot.

Fixes confusion for clients like LangChain.


OrKoN:
Thanks for the PR. Fixed via https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/205

PR #202, opened 2025-09-29
feat: increase mobile devices emulate

openmind-why:
Hello, my website is a mobile SSR project. When I was using MCP to check website performance, I found that I couldn't set mobile UA, which triggered my server PC UA redirection rule. Therefore, I think it is necessary to add this feature to provide the ability to check mobile projects

### DIFF
--- a/src/tools/ToolDefinition.ts
+++ b/src/tools/ToolDefinition.ts
@@ -4,12 +4,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import type {Dialog, ElementHandle, Page} from 'puppeteer-core';
+import type { Dialog, ElementHandle, Page } from 'puppeteer-core';
 import type z from 'zod';
 
-import type {TraceResult} from '../trace-processing/parse.js';
+import type { TraceResult } from '../trace-processing/parse.js';
 
-import type {ToolCategories} from './categories.js';
+import type { ToolCategories } from './categories.js';
 
 export interface ToolDefinition<Schema extends z.ZodRawShape = z.ZodRawShape> {
   name: string;
@@ -44,7 +44,7 @@ export interface Response {
   setIncludePages(value: boolean): void;
   setIncludeNetworkRequests(
     value: boolean,
-    options?: {pageSize?: number; pageIdx?: number; resourceTypes?: string[]},
+    options?: { pageSize?: number; pageIdx?: number; resourceTypes?: string[] },
   ): void;
   setIncludeConsoleData(value: boolean): void;
   setIncludeSnapshot(value: boolean): void;
@@ -73,8 +73,11 @@ export type Context = Readonly<{
   saveTemporaryFile(
     data: Uint8Array<ArrayBufferLike>,
     mimeType: 'image/png' | 'image/jpeg',
-  ): Promise<{filename: string}>;
+  ): Promise<{ filename: string }>;
   waitForEventsAfterAction(action: () => Promise<unknown>): Promise<void>;
+  // Added for multi-page device emulation support
+  createPagesSnapshot(): Promise<Page[]>;
+  getPages(): Page[];
 }>;
 
 export function defineTool<Schema extends z.ZodRawShape>(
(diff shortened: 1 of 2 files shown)


OrKoN:
Thanks for the PR! We will review it soon after we discuss if this is the way we want to expose the emulation options.

openmind-why:
> 您好，我的网站是一个移动端SSR项目。在使用MCP检查网站性能时，发现无法设置移动端UA，从而触发了服务器端PC端UA重定向规则。因此，我认为有必要添加此功能，提供检查移动端项目的功能。

src/tools/ToolDefinition.ts
```diff
@@ -73,8 +73,11 @@ export type Context = Readonly<{
   saveTemporaryFile(
     data: Uint8Array<ArrayBufferLike>,
     mimeType: 'image/png' | 'image/jpeg',
-  ): Promise<{filename: string}>;
+  ): Promise<{ filename: string }>;
   waitForEventsAfterAction(action: () => Promise<unknown>): Promise<void>;
+  // Added for multi-page device emulation support
+  createPagesSnapshot(): Promise<Page[]>;
```
starkshade, line 79: ## ⚠️ Code Issue

**Lines:** 79-79

```typescript
createPagesSnapshot(): Promise<Page[]>;
```

Returning a mutable array (`Page[]`) from a method that implies a 'snapshot' and is part of a `Readonly` context can lead to unintended modifications of the array itself by consumers.

src/tools/ToolDefinition.ts
```diff
@@ -73,8 +73,11 @@ export type Context = Readonly<{
...
+  getPages(): Page[];
```
starkshade, line 80: ## ⚠️ Code Issue

**Lines:** 80-80

```typescript
getPages(): Page[];
```

Returning a mutable array (`Page[]`) from a `get` method within a `Readonly` context can allow consumers to modify the internal array, potentially leading to unexpected side effects and breaking the encapsulation of the `Context`'s state.

src/tools/emulation.ts
```diff
@@ -4,17 +4,47 @@
...
+import { ToolCategories } from './categories.js';
+import { defineTool } from './ToolDefinition.js';
 
 const throttlingOptions: [string, ...string[]] = [
   'No emulation',
   ...Object.keys(PredefinedNetworkConditions),
 ];
 
+// common use device
+const deviceOptions: [string, ...string[]] = [
```
starkshade, line 20: ## 💡 Improvement Needed

**Line:** 20

The `deviceOptions` array is hardcoded. To ensure the list of available devices remains current with `puppeteer-core`'s `KnownDevices`, consider dynamically generating this array at runtime or build time. This would reduce maintenance overhead and ensure new devices are automatically included (or explicitly excluded if a curated list is still desired).

src/tools/emulation.ts
```diff
@@ -68,9 +98,124 @@ export const emulateCpu = defineTool({
...
+    if (device === 'No emulation') {
+      // apply desktop mode to all pages
+      for (const pageToEmulate of pagesToEmulate) {
+        await pageToEmulate.setViewport({
+          width: 1920,
+          height: 1080,
+          deviceScaleFactor: 1,
+          isMobile: false,
+          hasTouch: false,
+          isLandscape: true,
```
starkshade, line 181: ## 💡 Improvement Needed

**Line:** 181

Setting `isLandscape: true` for desktop mode is a specific default. For broader applicability, consider omitting this property to let Puppeteer use its default, or make it configurable if there's a specific reason to enforce landscape for desktop.

src/tools/emulation.ts
```diff
@@ -68,9 +98,124 @@ export const emulateCpu = defineTool({
...
+        });
+
+        await pageToEmulate.setUserAgent(
+          customUserAgent ||
```
starkshade, line 185: ## ⚠️ Issue

**Type:** Maintainability | **Severity:** Medium

**Line:** 185

The hardcoded Chrome 131.0.0.0 user agent will become outdated quickly, potentially causing issues with how websites perceive the 'desktop' browser.

**Fix:**
Dynamically retrieve the browser's user agent or use a more generic, less version-specific desktop user agent. Making it configurable could also be an option.

OrKoN:
Closing due to conflicts but the feature is in our backlog and we are considering implementing it.

PR #213, opened 2025-09-29
Fix: add dialog handling instruction in McpResponse.format

yasirusman85:
Added dialog handling (browser_handle_dialog) in McpResponse.format 
to satisfy test cases and improve consistency.


OrKoN:
Thanks for the PR but it does not appear like the change matches the title. Please re-open once the PR is consistent and fixes an issue.

PR #215, opened 2025-09-30
docs(getting-started): add `npx -y` to mcpServers args to auto-confirm install

wuchangming:
Avoids the "Need to install ... Ok to proceed?" interaction so the setup is non-interactive
for CI and first-time runs. (`-y` == `--yes`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ startup_timeout_ms = 20_000
 
 **Click the button to install:**
 
-[<img src="https://cursor.com/deeplink/mcp-install-dark.svg" alt="Install in Cursor">](https://cursor.com/en/install-mcp?name=chrome-devtools&config=eyJjb21tYW5kIjoibnB4IGNocm9tZS1kZXZ0b29scy1tY3BAbGF0ZXN0In0%3D)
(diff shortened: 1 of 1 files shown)


sebastianbenz:
Which client are you using? So far, all clients I've tried worked without the `-y` flag.

wuchangming:
> Which client are you using? So far, all clients I've tried worked without the `-y` flag.

I’m using Cursor. The issue happens when the version of chrome-devtools-mcp is higher than my local one — it fails to load. From what I can see, the installation gets interrupted at:
```
Need to install the following packages:
some-package@x.x.x
Ok to proceed? (y)
```
That’s why I had to add the -y flag.

README.md
sebastianbenz:
```suggestion
[<img src="https://cursor.com/deeplink/mcp-install-dark.svg" alt="Install in Cursor">](https://cursor.com/en/install-mcp?name=chrome-devtools&config=eyJjb21tYW5kIjoibnB4IC15IGNocm9tZS1kZXZ0b29scy1tY3BAbGF0ZXN0In0%3D)
```

sebastianbenz (approved):
Ah ok, seems to be an issue with Cursor then (just noticed that they also include the flag in their docs). I've added a suggestion to also update the add to cursor button. Thanks!

Issue #186, opened 2025-09-27
Snapshot does not capture Iframe content


aramvr:
**Describe the bug**
Currently, the snapshot tool does not capture content inside iframes.

I tested locally, and in `McpContext.ts`, adding `includeIframes: true` like this:
```
const rootNode = await page.accessibility.snapshot({
  includeIframes: true,
});
```
correctly includes iframe content, and all actions work as expected. There might be some additional context that I’m missing, but this seems to resolve the issue in my local tests. 

**To Reproduce**

1. Open a webpage that contains at least one <iframe> (e.g. an embedded form or external widget).
2. Run the snapshot tool to capture the accessibility tree.
3. Inspect the generated snapshot.
4. Notice that the iframe content is missing from the snapshot.

**Expected behavior**

**Chrome version:**
**Coding agent version:**
**Model version:**

**Screenshots**

**Chat log**
The full log of the chat with your coding agent.

aramvr:
👏

PR #217, opened 2025-09-30
fix: snapshot does not capture Iframe content (#186)

AlinaVarkki:


src/tools/snapshot.ts
```diff
@@ -36,11 +36,14 @@ export const waitFor = defineTool({
   },
   handler: async (request, response, context) => {
     const page = context.getSelectedPage();
+    const frames = page.frames();
 
-    await Locator.race([
-      page.locator(`aria/${request.params.text}`),
-      page.locator(`text/${request.params.text}`),
-    ]).wait();
+    const locators = frames.flatMap(frame => [
```
OrKoN, line 41: this change does not look right? is it intended

OrKoN (approved):

Issue #223, opened 2025-09-30
Configurable wait_for timeout


NT-Octaipipe:
**Is your feature request related to a problem? Please describe.**
I'm testing a reasonably fast website and even if I request a specific timeout in copilot chat or `copilot-instructions.md`, it always waits for 30s when an element is not present.

**Describe the solution you'd like**
Add a `wait_for` timeout to the configuration

Issue #116, opened 2025-09-24
Configurable timeout for page navigation


sitefinitysteve:
**Is your feature request related to a problem? Please describe.**
Sometimes depending on the page being checked, the CI login takes over 10 seconds and chrome-devtools throws a 10000 ms exceeded error

**Describe the solution you'd like**
Some way to configure the timeout to be longer

**Describe alternatives you've considered**
Would have to pre-load all pages before running

**Additional context**
The problem is it's asp.net dynamic compilation, so hitting a new page triggers each widget (controller) on that page to compile before the page returns results (if it's not already preloaded)

● I'll navigate to the /test-page using the CI login and check for console
   errors.

● chrome-devtools - navigate_page (MCP)(url: "https://dev.mysite.com/sso/cilogin?returnurl=/test-page")
  ⎿  Error: Navigation timeout of 10000 ms exceeded

asenrusev:
Same. Opening pages on dev can take longer if it needs to cold start a server function. I'd like to configure the timeout.

PR #228, opened 2025-09-30
feat: support timeouts in wait_for and navigations

OrKoN:
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/223
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/116

### DIFF
--- a/src/tools/pages.ts
+++ b/src/tools/pages.ts
@@ -9,7 +9,7 @@ import z from 'zod';
 import {logger} from '../logger.js';
 
 import {ToolCategories} from './categories.js';
-import {CLOSE_PAGE_ERROR, defineTool} from './ToolDefinition.js';
+import {CLOSE_PAGE_ERROR, defineTool, timeoutSchema} from './ToolDefinition.js';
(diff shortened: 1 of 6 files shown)


src/tools/pages.ts
```diff
@@ -129,15 +135,18 @@ export const navigatePageHistory = defineTool({
       .describe(
         'Whether to navigate back or navigate forward in the selected pages history',
       ),
+    ...timeoutSchema,
   },
   handler: async (request, response, context) => {
     const page = context.getSelectedPage();
-
+    const options = {
+      timeout: request.params.timeout ?? undefined,
```
Lightning00Blade:
```suggestion
      timeout: request.params.timeout,
```
?

OrKoN: Actually, meant to use `request.params.timeout ? request.params.timeout : undefined` as I do not think we should allow disabling the timeout.

Lightning00Blade (approved):

Issue #165, opened 2025-09-26
Feature Request: MCP-level configurable viewport dimensions


nwittwer:
**Is your feature request related to a problem? Please describe.**
The dimensions of the browser window/viewport by default is quite small, so complex web projects are more difficult to navigate due to limited space, and screenshots of the small window hide some parts of the UI due to scrollbars.

There _is_ the [`resize_page` tool](https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/docs/tool-reference.md#resize_page) which can be triggered via a prompt. 

The `resize_page` tool is interesting if the AI is implementing some responsive UI and and ensuring it works, but there's a lot of use cases where during normal development you test a few main browser window sizes consistently.

**Describe the solution you'd like**
Would be nice to be able to configure the viewport programmatically as a [MCP server configuration option](https://github.com/ChromeDevTools/chrome-devtools-mcp?tab=readme-ov-file#configuration).

```json5
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": [
        "chrome-devtools-mcp@latest",
        "--headless=true",
        "--viewport=1440x900", // <-- Set the initial viewport dimensions
      ]
    }
  }
}
```

Especially for teams using Cursor, where a `.cursor/mcp.json` file can be stored in the project for all team members to have consistent MCP server configuration. Because it would be programmatic, there's no risk that the Cursor rule about the custom viewport is ignored by the LLM when the tool session starts.

Future consideration could be given to help projects where they have 2-3 viewport sizes they often want to test. Could they comma-separate viewport `args`, and then multiple Chrome instances are used? Or it assumes always the first viewport in the list unless you prompt it to "test responsively" for example?

**Describe alternatives you've considered**
It's not super clear from the README, but would the current recommended approach be to use an always-applied Cursor rule (prompt), and suggest that the AI should use the `resize_page` tool whenever it uses the Chrome DevTools MCP? Risk that the LLM just ignores or forgets to do what you asked.

And is the vision that there would be more Puppeteer-level configuration options added as `tools` to this MCP server? There's 26 `tools` already, which at least with current models is quite high and often takes up a good chunk of input tokens. But context windows continue to increase, with prices generally decreasing.

**Additional context**
[...]

PR #229, opened 2025-09-30
feat: support initial viewport in the CLI

OrKoN:
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/165

This PR uses page_resize logic to set the initial viewport size. Not using the emulation because it looks buggy in headful which is our default mode.


src/cli.ts
```diff
@@ -54,6 +54,25 @@ export const cliOptions = {
...
+  viewport: {
+    type: 'string' as const,
+    describe:
+      'Initial viewport size for the Chromee instances started by the server. For example, `1280x720`',
+    coerce: (arg: string | undefined) => {
+      if (arg === undefined) {
+        return;
+      }
+      const [width, height] = arg.split('x').map(Number);
+      console.log(arg);
```
AlinaVarkki: `console.log('Viewport dimensions: ' + arg);` ?

PR #234, opened 2025-10-01
chore: miscellaneous fixes

Lightning00Blade:


src/logger.ts
```diff
@@ -18,12 +19,20 @@ export function saveLogsToFile(fileName: string): fs.WriteStream {
...
   debug.enable(namespacesToEnable.join(','));
 
-  const logFile = fs.createWriteStream(fileName, {flags: 'a'});
+  const logFile = fs.createWriteStream(fileName, {flags: 'a+'});
   debug.log = function (...chunks: any[]) {
     logFile.write(`${chunks.join(' ')}\n`);
   };
   logFile.on('error', function (error) {
-    console.log(`Error when opening/writing to log file: ${error.message}`);
+    console.log(
```
Lightning00Blade: If used inside `npx @modelcontextprotocol/inspector` this was erroring internally.
Thus the File error was unreadable. 
This should return a notification, though we may want to use a standard method type?

OrKoN: I think the fix is very specific to the stdio transport and would not work for HTTP transport. Can we emit it using the SDK and the server instance instead? we can pass the server instance to the logger.

OrKoN (commented):
cli.ts LGTM

src/logger.ts
OrKoN: also is not the inspector shows stderr logs in the left bottom side panel? I used to be able to see disclaimers there

Lightning00Blade: Logging with `console.error` work correctly to report it.

PR #237, opened 2025-10-01
Add an option value to the snapshot

### LINKED ISSUE #185: Cannot select <select> option by text
**Describe the bug**
When a <select> element has options like:
<option value="US">United States</option>
The snapshot records only the label (“United States”) instead of the value (“US”). The fill tool then tries to use “United States” as the value, which fails. Puppeteer seems to require the actual value attribute to select the option.

**To Reproduce**
Just try on a native <select> element where the value and text are different.

**Expected behavior**
The fill tool should select the option by value.

**Chrome version:** 140
**Coding agent version:** Cursor 1.4.5
**Model version:** Claude 4.0

**Screenshots**

**Chat log**
N/A

AlinaVarkki:
Fixes #185

The standard selector <option> value is not included in the AXNode, so add it separately.

### DIFF
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -285,15 +285,14 @@ so returned values have to JSON-serializable.
 
 - **args** (array) _(optional)_: An optional list of arguments to pass to the function.
 - **function** (string) **(required)**: A JavaScript function to run in the currently selected page.
-Example without arguments: `() => {
+  Example without arguments: `() => {
   return document.title
 }` or `async () => {
   return await fetch("example.com")
 }`.
-Example with arguments: `(el) => {
+  Example with arguments: `(el) => {
   return el.innerText;
 }`
-
 
 ---
 
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -285,14 +285,15 @@ so returned values have to JSON-serializable.
 
 - **args** (array) _(optional)_: An optional list of arguments to pass to the function.
 - **function** (string) **(required)**: A JavaScript function to run in the currently selected page.
-  Example without arguments: `() => {
+Example without arguments: `() => {
   return document.title
 }` or `async () => {
   return await fetch("example.com")
 }`.
-  Example with arguments: `(el) => {
+Example with arguments: `(el) => {
   return el.innerText;
 }`
+
 
 ---
 
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -30,7 +30,7 @@ import {WaitForHelper} from './WaitForHelper.js';
 export interface TextSnapshotNode extends SerializedAXNode {
   id: string;
   children: TextSnapshotNode[];
-  value?: string | number;
+  value?: string;
 }
 
 export interface TextSnapshot {
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -30,6 +30,7 @@ import {WaitForHelper} from './WaitForHelper.js';
 export interface TextSnapshotNode extends SerializedAXNode {
   id: string;
   children: TextSnapshotNode[];
+  value?: string | number;
 }
 
 export interface TextSnapshot {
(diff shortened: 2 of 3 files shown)


OrKoN (changes requested):
Thanks! Let us use conventional commits for the PR title. 

I think we should do the approach we discussed previously in which we resolve option label to a value when we will out an option. WDYT? I think doing evals during snapshot generation would be too expensive and bug-prone

AlinaVarkki:
> Thanks! Let us use conventional commits for the PR title.
> 
> I think we should do the approach we discussed previously in which we resolve option label to a value when we will out an option. WDYT? I think doing evals during snapshot generation would be too expensive and bug-prone

Do you mean check if it is an option within `fill_form` and get the value then? 
We can do that, but do you think there is no other case when we might need the value in the snapshot?
Maybe for some other tool to use?

src/tools/input.ts
```diff
@@ -97,7 +97,26 @@ export const fill = defineTool({
     const handle = await context.getElementByUid(request.params.uid);
     try {
       await context.waitForEventsAfterAction(async () => {
-        await handle.asLocator().fill(request.params.value);
+        // The AXNode for an option doesn't contain its `value`. We set text content of the option as value.
+        // To fill the form, get the correct option by its text value.
+        const isSelectElement = await handle.evaluate(
```
OrKoN: Could we lookup the a11y node by uid in the snapshot instead and check if it is an option?

src/tools/input.ts
```diff
@@ -97,7 +97,26 @@ export const fill = defineTool({
...
+          el => el.tagName === 'SELECT',
+        );
+        if (isSelectElement) {
```
OrKoN: we need to extract this method and re-use in fill_form as well

src/tools/input.ts
AlinaVarkki: the uid here is not for the option, but for a selector element.
We can look up that selector element by uid and check which of its childen value matches the one we are looking for. We still need to get its Element since we need to get the value that does not exist in the AXNode.

I changed it to do that

OrKoN: Ack, thanks!

src/tools/input.ts
```diff
@@ -78,6 +80,40 @@ export const hover = defineTool({
...
+  try {
+    // The AXNode for an option doesn't contain its `value`. We set text content of the option as value.
+    // If the form is a combobox, we need to find the correct option by its text value.
+    // To do that, loop through the children while checking which child's text matches the requested value (requested value is actually the text content).
+    // When the correct option is found, use the element handle to get the real value.
+    const aXNode = context.getAXNodeByUid(uid);
+    if (aXNode && aXNode.role === 'combobox' && aXNode.children) {
+      for (const child of aXNode.children) {
+        if (child.role === 'option' && child.name === value && child.value) {
+          const childHandle = await child.elementHandle();
```
OrKoN: we need to dispose the child handle after we use it

src/tools/input.ts
```diff
@@ -78,6 +80,40 @@ export const hover = defineTool({
...
+          if (childHandle) {
+            const childValueHandle = await childHandle.getProperty('value');
+            const childValue = await childValueHandle.jsonValue();
+            if (childValue) {
+              await handle.asLocator().fill(childValue.toString());
+            }
+            break;
+          }
+        }
```
OrKoN, line 111: we should throw an error if no option was found

src/tools/input.ts
```diff
@@ -78,6 +80,40 @@ export const hover = defineTool({
...
+  context: McpContext,
+) {
+  const handle = await context.getElementByUid(uid);
+  try {
+    // The AXNode for an option doesn't contain its `value`. We set text content of the option as value.
+    // If the form is a combobox, we need to find the correct option by its text value.
+    // To do that, loop through the children while checking which child's text matches the requested value (requested value is actually the text content).
+    // When the correct option is found, use the element handle to get the real value.
+    const aXNode = context.getAXNodeByUid(uid);
+    if (aXNode && aXNode.role === 'combobox' && aXNode.children) {
```
OrKoN: maybe we could extract the logic for finding an option in a combobox into a helper function?

PR #241, opened 2025-10-01
fix: PageCollector subscribing multiple times

Lightning00Blade:
We were setting the storage inside the first collect callback which mean that we double subscribed to the events.

This affected Network request and Console messages


src/PageCollector.ts
```diff
@@ -9,6 +9,11 @@ import type {Browser, HTTPRequest, Page} from 'puppeteer-core';
 export class PageCollector<T> {
```
OrKoN, line 9: could we add a test to catch this next time?

Lightning00Blade: Done.

OrKoN (approved):

PR #242, opened 2025-10-01
refactor: extract the pagination logic

Lightning00Blade:


src/McpResponse.ts
```diff
@@ -264,6 +246,32 @@ Call ${handleDialog.name} to handle it before continuing.`);
...
+      response.push('Invalid page number provided. Showing first page.');
+    }
+
+    const {startIndex, endIndex, currentPage, totalPages} = paginationResult;
+    response.push(
+      `Showing ${startIndex + 1}-${endIndex} of ${data.length} (Page ${currentPage + 1} of ${totalPages}).`,
+    );
+    if (pagination) {
+      if (paginationResult.hasNextPage) {
+        response.push(`Next page: ${currentPage + 1}`);
```
AlinaVarkki, line 262: should this not be `+2`?

src/McpResponse.ts
```diff
@@ -264,6 +246,32 @@ Call ${handleDialog.name} to handle it before continuing.`);
...
+      }
+      if (paginationResult.hasPreviousPage) {
+        response.push(`Previous page: ${currentPage - 1}`);
```
AlinaVarkki, line 265: and this just `currentPage `

src/McpResponse.ts
```diff
@@ -264,6 +246,32 @@ Call ${handleDialog.name} to handle it before continuing.`);
     return [text, ...images];
   }
 
+  #dataWithPagination<T>(data: T[], pagination?: PaginationOptions) {
```
AlinaVarkki, line 249: return type is missing

AlinaVarkki (commented):
something is failing :D

Lightning00Blade:
There was an GitHub Actions mac outage yesterday so those runners failed.

src/McpResponse.ts
Lightning00Blade: https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/250#discussion_r2397827804

src/McpResponse.ts
Lightning00Blade: I think the naming is wrong from the `paginate` helper I think this is suppose to be indexes or something.
As I have not change that impl I think we can merge as is, feel free to file an issue about it.

src/McpResponse.ts
Lightning00Blade: See the other thread.

PR #247, opened 2025-10-01
fix(emulation): Allow disabling CPU throttling

### LINKED ISSUE #245: Cannot Disable CPU Emulation in `emulate_cpu` Tool
*Description:*  
The `emulate_cpu` tool located in `src/tools/emulation.ts` does not currently support disabling CPU emulation once it has been enabled. The `throttlingRate` parameter enforces a minimum value of `1`, which applies the lowest level of throttling but does *not* fully disable the emulation. This behavior is inconsistent with the `emulate_network` tool, which allows network emulation to be completely turned off.

The underlying *Puppeteer API* (`page.emulateCPUThrottling`) expects a value of `null` or `0` to disable CPU throttling, which is not supported by the current implementation.

---

*To Reproduce:*

1. Enable CPU emulation using the `emulate_cpu` tool.
2. Try setting `throttlingRate = 0` or attempt to disable the feature.
3. Observe that throttling remains active at the minimum level instead of turning off.
4. Compare this behavior with `emulate_network`, which allows disabling.

---

*Expected Behavior:*  
Setting `throttlingRate` to `0` or `null` should *fully disable* CPU emulation, in line with how `emulate_network` behaves. The tool should allow disabling the feature entirely without requiring a full reload.

---

mithun50:
This PR Fixes #245 

The emulate_cpu tool did not provide a way to disable CPU emulation once it was enabled. This was because the throttlingRate parameter had a minimum value of 1, which only set the throttling to its lowest level but didn't turn off the emulation feature.

This change modifies the emulate_cpu tool to allow a throttlingRate of 0 to disable CPU emulation. The handler now passes null to Puppeteer's emulateCPUThrottling method when the rate is 0, which is the correct way to disable it.

A new test case has been added to verify that CPU throttling can be disabled by passing a throttlingRate of 0.

### DIFF
--- a/src/tools/emulation.ts
+++ b/src/tools/emulation.ts
@@ -60,17 +60,20 @@ export const emulateCpu = defineTool({
   schema: {
     throttlingRate: z
       .number()
-      .min(1)
+      .min(0)
       .max(20)
       .describe(
-        'The CPU throttling rate representing the slowdown factor 1-20x. Set the rate to 1 to disable throttling',
+        'The CPU throttling rate representing the slowdown factor 1-20x. Set the rate to 0 to disable throttling.',
       ),
   },
   handler: async (request, _response, context) => {
     const page = context.getSelectedPage();
     const {throttlingRate} = request.params;
 
-    await page.emulateCPUThrottling(throttlingRate);
+    await page.emulateCPUThrottling(
+      throttlingRate === 0 ? null : throttlingRate,
+    );
     context.setCpuThrottlingRate(throttlingRate);
   },
 });
+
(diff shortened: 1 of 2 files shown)


Lightning00Blade:
Thank you for the PR but this will have no affect on the underlying implementation as setting it to 1 effectively removes any set throttling. So I will close this PR and we can discuss on the issue (https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/245), if there are other issue.

Chromium source.
https://crsrc.org/c/third_party/blink/renderer/platform/scheduler/common/thread_cpu_throttler.cc;drc=746b6b8836d1f832ade517ba09d5149cbb58ec3f;l=203

PR #250, opened 2025-10-01
refactor: use context agnostic write to file

Lightning00Blade:
Similar to the write to temporary file we should hide the implementation behind the Context

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -58,9 +58,7 @@ function getNetworkMultiplierFromString(condition: string | null): number {
   return 1;
 }
 
-function getExtensionFromMimeType(
-  mimeType: 'image/png' | 'image/jpeg' | 'image/webp',
-) {
+function getExtensionFromMimeType(mimeType: string) {
   switch (mimeType) {
     case 'image/png':
       return 'png';
@@ -369,7 +367,7 @@ export class McpContext implements Context {
       return {filename};
     } catch (err) {
       this.logger(err);
-      throw new Error('Could not save a screenshot to a file');
+      throw new Error('Could not save a screenshot to a file', {cause: err});
     }
   }
   async saveFile(
@@ -382,7 +380,7 @@ export class McpContext implements Context {
       return {filename};
     } catch (err) {
       this.logger(err);
-      throw new Error('Could not save a screenshot to a file');
+      throw new Error('Could not save a screenshot to a file', {cause: err});
     }
   }
 


src/McpContext.ts
```diff
@@ -58,6 +58,20 @@ function getNetworkMultiplierFromString(condition: string | null): number {
   return 1;
 }
 
+function getExtensionFromMimeType(
```
AlinaVarkki: let's add a return type

src/tools/ToolDefinition.ts
```diff
@@ -74,6 +74,10 @@ export type Context = Readonly<{
     data: Uint8Array<ArrayBufferLike>,
     mimeType: 'image/png' | 'image/jpeg' | 'image/webp',
```
AlinaVarkki, line 75: the types are listen again here. We can make some `export type MimeType`

src/McpContext.ts
```diff
@@ -346,20 +360,31 @@ export class McpContext implements Context {
...
+    data: Uint8Array<ArrayBufferLike>,
+    filename: string,
+  ): Promise<{filename: string}> {
+    try {
+      const filePath = path.resolve(filename);
+      await fs.writeFile(filePath, data);
+      return {filename};
+    } catch (err) {
+      this.logger(err);
+      throw new Error('Could not save a screenshot to a file');
```
AlinaVarkki: nit: `throw new Error(`Could not save screenshot to path: ${filename}`);`

AlinaVarkki (approved):
Great change, lgtm!

src/McpContext.ts
Lightning00Blade: Not needed - https://google.github.io/styleguide/tsguide.html#type-inference

Usually you would want that for very big project to reduce type checking and building times.

src/tools/ToolDefinition.ts
Lightning00Blade: I think we should make this a `string` as this is not an exhaustive list of all MimeTypes

src/McpContext.ts
Lightning00Blade: I think that is fine but we can send the `{cause: err}`, and let the underlying implementation transport it.

src/tools/ToolDefinition.ts
AlinaVarkki: but we are only accepting those right now. Sounds like more of a reason to make a type for them so we don't list them in multiple places

`export type MimeType = 'image/png' | 'image/jpeg' | 'image/webp';`

I don't feel strongly about it, feel free to merge

PR #266, opened 2025-10-03
(do not merge): exposing additional performance functions

jackfranklin:
This PR is so we have a branch that has the bigger set of performance
functions available to it that are equivalent to the ones the DevTools
Performance Agent has.

If we decide we want to expose these functions, we will do so via
reusing the implementation from upstream, not via the copy/paste of this
PR, but this is the quickest way to get them available to test.


OrKoN:
Closing as not planning to land soon.

PR #267, opened 2025-10-03
feat: add request and response body

nroscino:
This PR adds the request and response bodies into the context if they are available and in a readable format.

### DIFF
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -16,7 +16,6 @@ import {
   getFormattedRequestBody,
   getShortDescriptionForRequest,
   getStatusFromRequest,
-  BODY_CONTEXT_SIZE_LIMIT,
 } from './formatters/networkFormatter.js';
 import {formatA11ySnapshot} from './formatters/snapshotFormatter.js';
 import type {McpContext} from './McpContext.js';
@@ -172,10 +171,7 @@ export class McpResponse implements Response {
   async processResponseBody(
     httpResponse: HTTPResponse,
   ): Promise<string | null> {
-    const formattedResponseData = await getFormattedResponseBody(
-      httpResponse,
-      BODY_CONTEXT_SIZE_LIMIT,
-    );
+    const formattedResponseData = await getFormattedResponseBody(httpResponse);
     if (formattedResponseData.length > 0) {
       return formattedResponseData;
     }
@@ -184,12 +180,9 @@ export class McpResponse implements Response {
   }
 
   async processRequestBody(httpRequest: HTTPRequest): Promise<string | null> {
-    const formattedResponseData = await getFormattedRequestBody(
-      httpRequest,
-      BODY_CONTEXT_SIZE_LIMIT,
-    );
-    if (formattedResponseData.length > 0) {
-      return formattedResponseData;
+    const formattedRequestData = await getFormattedRequestBody(httpRequest);
+    if (formattedRequestData.length > 0) {
+      return formattedRequestData;
     }
 
     return null;
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -7,7 +7,7 @@ import type {
   ImageContent,
   TextContent,
 } from '@modelcontextprotocol/sdk/types.js';
-import type {HTTPRequest, HTTPResponse, ResourceType} from 'puppeteer-core';
+import type {ResourceType} from 'puppeteer-core';
 
 import {formatConsoleEvent} from './formatters/consoleFormatter.js';
 import {
@@ -144,12 +144,12 @@ export class McpResponse implements Response {
       );
 
       this.#attachedNetworkRequestData.requestBody =
-        await this.processRequestBody(request);
+        await getFormattedRequestBody(request);
 
       const response = request.response();
       if (response) {
         this.#attachedNetworkRequestData.responseBody =
-          await this.processResponseBody(response);
+          await getFormattedResponseBody(response);
       }
     }
 
@@ -164,28 +164,6 @@ export class McpResponse implements Response {
     }
 
     return this.format(toolName, context);
-  }
-
-  async processResponseBody(
-    httpResponse: HTTPResponse,
-  ): Promise<string | undefined> {
-    const formattedResponseData = await getFormattedResponseBody(httpResponse);
-    if (formattedResponseData.length > 0) {
-      return formattedResponseData;
-    }
-
-    return undefined;
-  }
-
-  async processRequestBody(
-    httpRequest: HTTPRequest,
-  ): Promise<string | undefined> {
-    const formattedRequestData = await getFormattedRequestBody(httpRequest);
-    if (formattedRequestData.length > 0) {
-      return formattedRequestData;
-    }
-
-    return undefined;
   }
 
   format(
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -7,9 +7,9 @@ import type {
   ImageContent,
   TextContent,
 } from '@modelcontextprotocol/sdk/types.js';
-import type {ResourceType} from 'puppeteer-core';
+import type { HTTPRequest, HTTPResponse, ResourceType } from 'puppeteer-core';
 
-import {formatConsoleEvent} from './formatters/consoleFormatter.js';
+import { formatConsoleEvent } from './formatters/consoleFormatter.js';
 import {
   getFormattedHeaderValue,
   getFormattedResponseBody,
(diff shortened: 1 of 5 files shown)


src/formatters/networkFormatter.ts
```diff
@@ -4,7 +4,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import type {HTTPRequest} from 'puppeteer-core';
+import {isUtf8} from 'node:buffer';
+
+import type {HTTPRequest, HTTPResponse} from 'puppeteer-core';
+
+export const BODY_CONTEXT_SIZE_LIMIT = 10000;
```
Lightning00Blade: Let's have this a Default value and not export it.
Also small nit, it makes it easier to read.
```suggestion
export const BODY_CONTEXT_SIZE_LIMIT = 10_000;
```

src/McpResponse.ts
```diff
@@ -7,24 +7,33 @@ import type {
...
   getStatusFromRequest,
+  BODY_CONTEXT_SIZE_LIMIT,
 } from './formatters/networkFormatter.js';
 import {formatA11ySnapshot} from './formatters/snapshotFormatter.js';
 import type {McpContext} from './McpContext.js';
 import {handleDialog} from './tools/pages.js';
 import type {ImageContentData, Response} from './tools/ToolDefinition.js';
 import {paginate, type PaginationOptions} from './utils/pagination.js';
 
+export interface NetworkRequestData {
```
Lightning00Blade: Why do we need to export this interface?

src/McpResponse.ts
```diff
@@ -7,24 +7,33 @@ import type {
...
+  networkRequestUrl: string;
+  requestBody: string | null;
+  responseBody: string | null;
```
Lightning00Blade:
```suggestion
  requestBody?: string;
  responseBody?: string;
```

src/McpResponse.ts
```diff
@@ -74,7 +83,11 @@ export class McpResponse implements Response {
   }
 
   attachNetworkRequest(url: string): void {
-    this.#attachedNetworkRequestUrl = url;
+    this.#attachedNetworkRequestData = {
+      networkRequestUrl: url,
+      responseBody: null,
+      requestBody: null,
```
Lightning00Blade: We can remove this with the change above

src/McpResponse.ts
```diff
@@ -140,6 +169,32 @@ export class McpResponse implements Response {
...
+    );
+    if (formattedResponseData.length > 0) {
+      return formattedResponseData;
+    }
+
+    return null;
+  }
+
+  async processRequestBody(httpRequest: HTTPRequest): Promise<string | null> {
+    const formattedResponseData = await getFormattedRequestBody(
```
Lightning00Blade: Can we move this login inside the formatted and return string|null from there directly?
And remove this wrapper.

src/McpResponse.ts
```diff
@@ -140,6 +169,32 @@ export class McpResponse implements Response {
     return this.format(toolName, context);
   }
 
+  async processResponseBody(
+    httpResponse: HTTPResponse,
+  ): Promise<string | null> {
+    const formattedResponseData = await getFormattedResponseBody(
```
Lightning00Blade: Can we move this login inside the formatted and return string|null from there directly?
And remove this wrapper.

src/formatters/networkFormatter.ts
```diff
@@ -37,3 +41,61 @@ export function getFormattedHeaderValue(
...
+      return '';
+    }
+  }
+
+  return '';
+}
+
+function getSizeLimitedString(text: string, sizeLimit: number) {
+  if (text.length > sizeLimit) {
+    return `${text.substring(0, sizeLimit) + '... <truncated>'}`;
```
Lightning00Blade, line 97: Technically this goes over the limit as you are not accounting for the added text.

src/McpResponse.ts
nroscino: You are right, we don't need to!

src/formatters/networkFormatter.ts
nroscino: Yeah this was wanted, but I can include this text in the limit if we want to. What do you think?

src/formatters/networkFormatter.ts
```diff
@@ -37,3 +41,61 @@ export function getFormattedHeaderValue(
   }
   return response;
 }
+
+export async function getFormattedResponseBody(
+  httpResponse: HTTPResponse,
+  sizeLimit: number = BODY_CONTEXT_SIZE_LIMIT,
```
Lightning00Blade: Nit:
```suggestion
  sizeLimit = BODY_CONTEXT_SIZE_LIMIT,
```

src/formatters/networkFormatter.ts
Lightning00Blade: I see, I think we can keep as is.

src/formatters/networkFormatter.ts
```diff
@@ -37,3 +41,61 @@ export function getFormattedHeaderValue(
...
+        return `<empty response>`;
+      }
+
+      return `${getSizeLimitedString(responseAsTest, sizeLimit)}`;
+    }
+
+    return `<binary data>`;
+  } catch {
+    // buffer() call might fail with CDP exception, in this case we don't print anything in the context
+    return '';
```
Lightning00Blade: Let's just make this return type `string|undefined`  and we can remove the wrappers in the MCPResponse.
I think in the case the Body is empty we still want to show that it was empty as long as the above succeeded.

src/formatters/networkFormatter.ts
```diff
@@ -37,3 +41,61 @@ export function getFormattedHeaderValue(
...
+
+    try {
+      const fetchData = await httpRequest.fetchPostData();
+
+      if (fetchData) {
+        return `${getSizeLimitedString(fetchData, sizeLimit)}`;
+      }
+    } catch {
+      // fetchPostData() call might fail with CDP exception, in this case we don't print anything in the context
+      return '';
```
Lightning00Blade: Ditto.

src/McpResponse.ts
nroscino: Instead of null I returned `undefined` otherwise I will have a type error when I try to assign it to a `string | undefined`.

src/formatters/networkFormatter.ts
```diff
@@ -37,3 +41,61 @@ export function getFormattedHeaderValue(
...
+        return `<empty response>`;
+      }
+
+      return `${getSizeLimitedString(responseAsTest, sizeLimit)}`;
+    }
+
+    return `<binary data>`;
+  } catch {
+    // buffer() call might fail with CDP exception, in this case we don't print anything in the context
+    return undefined;
```
Lightning00Blade: Nit:
```suggestion
    return;
```

src/formatters/networkFormatter.ts
```diff
@@ -37,3 +41,61 @@ export function getFormattedHeaderValue(
...
+
+    try {
+      const fetchData = await httpRequest.fetchPostData();
+
+      if (fetchData) {
+        return `${getSizeLimitedString(fetchData, sizeLimit)}`;
+      }
+    } catch {
+      // fetchPostData() call might fail with CDP exception, in this case we don't print anything in the context
+      return undefined;
```
Lightning00Blade: Ditto

src/formatters/networkFormatter.ts
```diff
@@ -37,3 +41,61 @@ export function getFormattedHeaderValue(
...
+    }
+  }
+
+  return undefined;
```
Lightning00Blade: Ditto

tests/McpResponse.test.ts
```diff
@@ -202,6 +202,51 @@ http://example.com GET [pending]`,
...
   });
 
+  it('add network request when attached with POST data', async () => {
+    await withBrowser(async (response, context) => {
+      response.setIncludeNetworkRequests(true);
+      const httpResponse = getMockResponse();
+      httpResponse.buffer = () => {
+        return Promise.resolve(Buffer.from(JSON.stringify({response: 'body'})));
+      };
+      httpResponse.headers = () => {
```
Lightning00Blade, line 212: Nit: remove this as we don't need to test this part.

nroscino: Here I can only return a `{}` but cannot avoid also adding the headers mock since if an `httpResponse` exists then it checks for the headers (McpResponse.#getIncludeNetworkRequestsData:322-327). I also noticed that there's no test covering those lines, should I add it separately or since we cannot create a response without the headers we can test the response detail as a whole?

Lightning00Blade: I see. Both options sound OK - leaving as is or changing to `{}`. It would be good to have a test for that as well but lets do it in a separate PR.

PR #270, opened 2025-10-04
fix(cli): handle empty --browserUrl to prevent Invalid URL crash

yasirusman85:
This PR fixes a startup bug in `chrome-devtools-mcp` where passing `--browserUrl ""` (as happens in **VS Code Insiders with GitHub Copilot MCP integration**) caused the process to exit with:

```
Invalid URL
```

### Before fix

```bash
node build/src/index.js --browserUrl ""
# → Invalid URL, process exited with code 1
```

### After fix

```bash
node build/src/index.js --browserUrl ""
# → Falls back to channel=stable, server starts successfully
chrome-devtools-mcp exposes content of the browser instance...
```

### Root cause

* `cli.ts` coerced even empty strings via `new URL(url)`, which throws.
* VS Code Insiders MCP integration sometimes passes `--browserUrl ""` instead of omitting the flag.

### Fix

* Updated `coerce` in `cli.ts` to safely ignore empty/blank strings.
* Fallback logic now correctly assigns `channel: "stable"`.

### Impact

* Fixes MCP server crash in **VS Code Insiders**.
* Preserves existing behavior for valid URLs and other CLI options.


Lightning00Blade:
@yasirusman85
Please remove all unrelated changes to the console, and after that I will be happy to review the PR.

OrKoN:
Closed via https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/298

PR #271, opened 2025-10-04
feat: support protocolTimeout

dungnguyent8:
This PR introduces a new `--protocolTimeout` CLI option, allowing users to configure the timeout for Chrome DevTools Protocol operations.

**Key Changes:**
*   [`README.md`](README.md): Updated to document the new `--protocolTimeout` CLI option.
*   [`src/cli.ts`](src/cli.ts): Added the `protocolTimeout` option to the CLI argument parser.
*   [`src/browser.ts`](src/browser.ts): Modified browser connection and launch functions to accept and utilize the `protocolTimeout` value, defaulting to 10 seconds if not specified. A clarifying comment was added to the `McpLaunchOptions` interface.
*   [`src/main.ts`](src/main.ts): Integrated the `protocolTimeout` argument when connecting to or launching the browser.

These changes enhance flexibility by enabling users to adjust the protocol timeout based on their specific needs, improving stability for operations that might require longer execution times.

Related issue:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,10 @@ The Chrome DevTools MCP server supports the following configuration option:
   If enabled, ignores errors relative to self-signed and expired certificates. Use with caution.
   - **Type:** boolean
 
+- **`--protocolTimeout`**
+  Timeout for the Chrome DevTools Protocol operations in milliseconds.
+  - **Type:** number
+
 <!-- END AUTO GENERATED OPTIONS -->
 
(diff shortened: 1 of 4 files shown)


OrKoN:
@Lightning00Blade I wonder if we should just bump the protocol timeout value or use Puppeteer's default instead of making it configurable? WDYT?

Lightning00Blade:
I think using the default should be OK. We do set some timeouts high ourselves when emulation is enabled.

PR #272, opened 2025-10-04
feat: add emulate_device_profile tool

hiowenluke:
- add predefined mobile device profiles and CDP emulation logic
- register emulate_device_profile tool, document usage and Copilot prompt example
- cover the new tool with automation tests and update the tool listing


albertehc:
Any updates on this feature?
It would be interesting to be able to emulate a mobile device from the MCP

PR #279, opened 2025-10-04
add a tail argument to getConsoleData

myers:
This adds a default limit to how many console messages are returned to avoid filling LLM's context.


Lightning00Blade:
The PR https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/387 add pagination to the console messages, so this one is no longer needed. It will be released as part of v0.9.0

PR #295, opened 2025-10-07
fix: guard performance_stop_trace when tracing inactive

leslieo2:
PR Description

  - Fix the stop-trace guard to call context.isRunningPerformanceTrace() so we skip page.tracing.stop() when nothing is recording, avoiding the Puppeteer error that bubbled up
    to users.
  - Extend the “does nothing” test to stub tracing.stop() and assert it stays untouched, guaranteeing the guard path is covered.

  Testing

  - npm run build
  - node --require ./build/tests/setup.js --no-warnings=ExperimentalWarning --test-force-exit --test-reporter spec --test build/tests/tools/performance.test.js

### DIFF
--- a/src/tools/performance.ts
+++ b/src/tools/performance.ts
@@ -111,7 +111,7 @@ export const stopTrace = defineTool({
   },
   schema: {},
   handler: async (_request, response, context) => {
-    if (!context.isRunningPerformanceTrace) {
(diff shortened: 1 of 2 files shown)


leslieo2:
Blocked by “workflows awaiting approval”, needs maintainer approval to run CI.

PR #301, opened 2025-10-07
chore: provide a structured bug template

Lightning00Blade:
I think this will help people structure what is needed more easily.
The current version does not restrict anything so it's as close as a no-op as possible

TODO: I need to look in how to add the screenshot part back.


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,73 @@
+name: Bug report
+description: File a bug report for chrome-devtools-mcp
+title: '[Bug] '
```
sebastianbenz: Should we prefix bugs with a title? This is redundant to labels

sebastianbenz: Speaking of, can we add the `bug` label here?

Lightning00Blade: So recently GitHub added a type for issues, but I can't find how to set it in the docs.
But I see we heavily use the label, so I will add for now.
As for the Prefix I think we can omit it for now, to keep it as simple transition as possible.

Lightning00Blade: Done

.github/ISSUE_TEMPLATE/01-bug.yml
```diff
@@ -0,0 +1,73 @@
+name: Bug report
```
Lightning00Blade, line 1: The files are always show in an alphanumerical order so we prefix it with 01 to make sure its the first one.

sebastianbenz (approved):

PR #318, opened 2025-10-08
refactor: dispose listeners on page destroyed

Lightning00Blade:


src/McpContext.ts
```diff
@@ -94,26 +95,24 @@ export class McpContext implements Context {
...
-        page.on('request', request => {
+    this.#networkCollector = new NetworkCollector(this.browser, collect => {
+      return {
+        request: request => {
           collect(request);
-        });
-      },
-    );
+        },
+      } as ListenerMap;
```
Lightning00Blade, line 103: Somehow the types are right but incompatible

src/PageCollector.ts
```diff
@@ -4,11 +4,26 @@
...
+  type Frame,
+  type Handler,
+  type HTTPRequest,
+  type Page,
+  type PageEvents,
+} from 'puppeteer-core';
+
+export type ListenerMap<EventMap extends PageEvents = PageEvents> = {
+  [K in keyof EventMap]?: (event: EventMap[K]) => void;
+  // request: (event: PageEvents['request']) => void;
```
OrKoN: remove?

src/PageCollector.ts
```diff
@@ -37,6 +52,13 @@ export class PageCollector<T> {
       }
       this.#initializePage(page);
     });
+    this.#browser.on('targetdestroyed', async target => {
+      const page = await target.page();
```
OrKoN, line 55: would it resolve into a page for destroyed targets?

OrKoN: let's add a test for this scenario

Lightning00Blade: Done

Issue #259, opened 2025-10-02
Improve error handling in navigate_page_history tool


mithun50:
The navigate_page_history tool in src/tools/pages.ts currently suppresses specific errors when a navigation action (back or forward) fails. Instead of providing context about why the navigation failed (e.g., no page in history), it returns a generic message: Unable to navigate {back|forward} in currently selected page..

This should be improved to check the navigation history before attempting the action. If the navigation is not possible, a more informative error message should be returned to the user, such as "Cannot go back, no previous page in history." This will improve the tool's usability and provide better feedback.

mithun50:
If you guys assign this issue I can work on this issue

natorion:
In what scenarios, with what prompts would you use this?

PR #321, opened 2025-10-09
fix: Improve navigate_page_history error messages

mithun50:
## Problem
  The `navigate_page_history` tool previously suppressed all
  navigation errors with a generic message: "Unable to
  navigate {back|forward} in currently selected page." This
  provided no context about why the navigation failed, making
  it difficult for users to understand whether:
  - No page exists in the history
  - A timeout occurred
  - Some other error happened

  ## Solution
  Enhanced error handling to provide specific, actionable
  feedback:

  1. **Check navigation result**: Capture the return value
  from `goBack()`/`goForward()` to detect when navigation is
  impossible
  2. **No history detection**: When result is `null`, return:
  "Cannot navigate back, no previous page in history."
  3. **Timeout errors**: Show duration: "Navigation back timed
   out after 30000ms."
  4. **Other errors**: Include actual error message instead of
   generic text

  ## Changes
  - Modified `navigate_page_history` handler in
  `src/tools/pages.ts`
  - Added result checking for null (no history available)
  - Enhanced error messages with specific context
  - Improved debugging experience for users

  ## Testing
  - Verify "no previous page" message when navigating back on
  first page
  - Verify "no next page" message when navigating forward at
  end of history
  - Confirm timeout errors show duration
  - Ensure other errors display actual error messages

### DIFF
--- a/tests/tools/pages.test.ts
+++ b/tests/tools/pages.test.ts
@@ -166,7 +166,9 @@ describe('pages', () => {
         assert.ok(
           response.responseLines
             .at(0)
-            ?.startsWith('Unable to navigate forward in currently selected page.'),
+            ?.startsWith(
+              'Unable to navigate forward in currently selected page.',
+            ),
         );
         assert.ok(response.includePages);
       });


mithun50:
Fixes: #259

src/tools/pages.ts
```diff
@@ -142,16 +142,34 @@ export const navigatePageHistory = defineTool({
...
     try {
+      let result;
       if (request.params.navigate === 'back') {
-        await page.goBack(options);
+        result = await page.goBack(options);
+      } else {
+        result = await page.goForward(options);
+      }
+
+      // If result is null, navigation wasn't possible (no history)
```
OrKoN: I think this is not correct. Successful navigations could result in a null result if the navigation has no network request. We fixed Puppeteer in https://github.com/puppeteer/puppeteer/pull/14160 to throw an error when there is no history. So the null result, should be treated as a successful navigation.

src/tools/pages.ts
```diff
@@ -142,16 +142,34 @@ export const navigatePageHistory = defineTool({
...
+        );
       } else {
-        await page.goForward(options);
+        response.appendResponseLine(
+          `Unable to navigate ${request.params.navigate}: ${error.message || 'Unknown error occurred'}`,
+        );
       }
-    } catch {
-      response.appendResponseLine(
-        `Unable to navigate ${request.params.navigate} in currently selected page.`,
```
OrKoN, line 153: I think it would be sufficient to add the error.message to this message and revert other changes

src/tools/pages.ts
mithun50: @OrKoN can you please review the code

Issue #240, opened 2025-10-01
list_network_requests: it's not clear to the LLM that these are not reset after each call


sebastianbenz:
**Describe the bug**
The LLM used list_network_requests to verify caching behavior and clicked a button twice and listed the network requests after each click. It was confused that the network requests stay the same. Might be worth explaining more in the description when the requests are reset.

**Expected behavior**
The LLM should understand when network requests are reset.

**Model version:** Gemini 2.5

**Chat log**

I've applied the fix to cache the articles. Now I'll verify that the delay is gone by clicking the search button multiple times and checking the network requests. First, I'll navigate to the page.

check
navigate_page (chrome-devtools MCP Server).../localhost:4321/
stat_minus_1
Thinking
Okay, I'm on the page. I'll take a snapshot to find the search button and then click it.

check
take_snapshot (chrome-devtools MCP Server)
stat_minus_1
Thinking
Okay, I have the snapshot. Clicking the search button now.

check
click (chrome-devtools MCP Server)7_1
stat_1
Thinking
Finding the Target
I've identified the "Search" button and the search textbox. Now I need to locate and target the close button within the search snapshot. The next step is to extract the close button's identifier, building on the previous search command's output.

Isolating the Element
I've confirmed the search button has been pressed, generating a new UI snapshot. Analyzing the output, the close button, identified as uid=8_5, is clear. My next move is to execute a click on this specific element, following the established pattern.

Okay, I've clicked the search button. Now I'll click the close button, which is uid=8_5.

check
click (chrome-devtools MCP Server)8_5
stat_minus_1
Thinking
Okay, flyout closed. Clicking the search button again.

check
click (chrome-devtools MCP Server)9_1
stat_minus_1
Thinking
Okay, I've clicked the search button again. Now, I'll check the network requests to confirm the fix.

check
list_network_requests (chrome-devtools MCP Server)
Arguments
stat_minus_1
Thinking
My apologies, it seems my previous fix didn't work as intended. The articles are still being fetched multiple times.

I'm going to implement a more robust caching mechanism to resolve this. This will ensure that the articles are only fetched once, which should eliminate the delay you're experiencing. I'll apply this new fix now.

help

PR #335, opened 2025-10-10
fix: document that console and requests are since the last nav

OrKoN:
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/240

### DIFF
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -264,7 +264,7 @@
 
 ### `list_network_requests`
 
-**Description:** List all requests for the currently selected page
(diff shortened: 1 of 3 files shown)


OrKoN:
> Thanks! Have you been able to verify that this improves the situation?

Unfortunately, the bug did not contain the steps to reproduce. When asked directly gemini responds: `No, the network requests are cleared after each navigation. The list_network_requests tool only shows the requests that have occurred since the last navigation on the currently active page.` so I think it should help. If you could re-test it would be great.

PR #349, opened 2025-10-11
refactor: use release-please to bump version in files

Lightning00Blade:


src/main.ts
```diff
@@ -33,33 +29,21 @@ import * as scriptTools from './tools/script.js';
...
-    return json;
-  } catch {
-    return {};
-  }
-}
-
-const version = readPackageJson().version ?? 'unknown';
+// If moved update release-please config
+// x-release-please-start-version
+const VERSION = '0.6.0';
```
Lightning00Blade:
```suggestion
const VERSION = '0.8.0';
```

Lightning00Blade:
Yeah we can, after we verify it works here.

PR #362, opened 2025-10-13
fix: add an option value to the snapshot

### LINKED ISSUE #185: Cannot select <select> option by text

AlinaVarkki:
Fixes #185 

The standard selector value is not included in the AXNode, so add it separately.

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -352,7 +352,7 @@ export class McpContext implements Context {
       return nodeWithId;
     };
 
-    const rootNodeWithId = await assignIds(rootNode);
+    const rootNodeWithId = assignIds(rootNode);
     this.#textSnapshot = {
       root: rootNodeWithId,
       snapshotId: String(snapshotId),


src/McpContext.ts
```diff
@@ -326,19 +330,37 @@ export class McpContext implements Context {
...
         id: `${snapshotId}_${idCounter++}`,
-        children: node.children
-          ? node.children.map(child => assignIds(child))
-          : [],
+        children: [],
       };
+
+      // The AXNode for an option doesn't contain its `value`.
+      // Therefore, set text content of the option as value.
+      if (node.role === 'option') {
```
OrKoN, line 344: is there a reason why we cannot use the node.name? I really do not want use to have evals when building snapshots. If we cannot do it without evals, we need to fix it in the chromium code.

src/McpContext.ts
```diff
@@ -326,19 +330,33 @@ export class McpContext implements Context {
     // will be used for the tree serialization and mapping ids back to nodes.
     let idCounter = 0;
     const idToNode = new Map<string, TextSnapshotNode>();
-    const assignIds = (node: SerializedAXNode): TextSnapshotNode => {
+    const assignIds = async (
```
OrKoN: this does not need to be async now, let's change it back to sync and revert other changes related to making it async.

src/McpContext.ts
```diff
@@ -334,11 +338,21 @@ export class McpContext implements Context {
...
+          nodeWithId.value = optionText.toString();
+        }
+      }
+
       idToNode.set(nodeWithId.id, nodeWithId);
       return nodeWithId;
     };
 
-    const rootNodeWithId = assignIds(rootNode);
+    const rootNodeWithId = await assignIds(rootNode);
```
OrKoN:
```suggestion
    const rootNodeWithId = assignIds(rootNode);
```

PR #365, opened 2025-10-13
chore: use template for feature requests

Lightning00Blade:


.github/ISSUE_TEMPLATE/01-bug.yml
```diff
@@ -70,3 +70,11 @@ body:
         - Windows
         - macOS
         - Linux
+
+  - type: checkboxes
```
OrKoN, line 74: is this change intended?

.github/ISSUE_TEMPLATE/02-feature.yml
```diff
@@ -0,0 +1,44 @@
+name: Bug report
```
OrKoN: Feature?

.github/ISSUE_TEMPLATE/01-bug.yml
Lightning00Blade: I think we want it as people keep sending a separate message regarding something like this

.github/ISSUE_TEMPLATE/02-feature.yml
Lightning00Blade:
```suggestion
name: Feature
```

.github/ISSUE_TEMPLATE/01-bug.yml
OrKoN: Why does it say "What kinds of cats do you like?"?

Lightning00Blade: Sorry I just copy pasted the default example 😸

PR #364, opened 2025-10-13
fix: correct dialog default value display (#361)

mithun50:
## Description

  Fixes the dialog information display to show the actual
  default value instead of repeating the dialog message.

  Fixes #361

  ## Changes

  - Updated `McpResponse.ts` line 196 to use
  `dialog.defaultValue()` instead of `dialog.message()` for
  displaying the default value

  ## Before

  prompt: Enter your name (default value: Enter your name)

  ## After

  prompt: Enter your name (default value: John Doe)

  ## Testing

  - ✅ Existing dialog tests pass (handleDialog
  acceptance/dismissal)
  - ✅ Code formatting verified with prettier
  - ✅ No breaking changes

  ## Impact

  - **Type**: Bug fix
  - **Severity**: Minor
  - **User Impact**: Improves UX by showing correct dialog
  default values instead of confusing duplicated messages

Issue #361, opened 2025-10-13
bug: Dialog default value displays message instead of defaultValue


mithun50:
## Bug Description

  In `McpResponse.ts`, the dialog information display shows
  `dialog.message()` twice instead of showing the actual
  default value.

  ## Location

  **File**: `src/McpResponse.ts`
  **Line**: 196

  ## Current Behavior

  ```typescript
  ${dialog.type()}: ${dialog.message()} (default value:
  ${dialog.message()})

  This displays the dialog message as both the message AND the
   default value, which is incorrect.

  Expected Behavior

  ${dialog.type()}: ${dialog.message()} (default value:
  ${dialog.defaultValue()})
```
  Should use dialog.defaultValue() to display the actual
  default value of the dialog.

  Example

  Current output:
  prompt: Enter your name (default value: Enter your name)

  Expected output:
  prompt: Enter your name (default value: John Doe)

  Impact

  - Severity: Minor
  - User Impact: Confusing UX - users see incorrect
  information about dialog default values
  - Functionality: Does not break functionality, but displays
  misleading information

mithun50:
If you guys assign this issue to me I can work on it

PR #366, opened 2025-10-13
fix: return the default dialog value correctly

OrKoN:
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/364 
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/361

### DIFF
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -192,8 +192,12 @@ export class McpResponse implements Response {
 
     const dialog = context.getDialog();
     if (dialog) {
+      const defaultValueIfNeeded =
+        dialog.type() === 'prompt'
+          ? ` (default value: "${dialog.defaultValue()}")`
+          : '';
       response.push(`# Open dialog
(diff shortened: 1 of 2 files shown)


src/McpResponse.ts
```diff
@@ -193,7 +193,7 @@ export class McpResponse implements Response {
     const dialog = context.getDialog();
     if (dialog) {
       response.push(`# Open dialog
-${dialog.type()}: ${dialog.message()} (default value: ${dialog.message()}).
+${dialog.type()}: ${dialog.message()} (default value: ${dialog.defaultValue()}).
```
Lightning00Blade: Maybe we can add "" around the value to signify if it was empty?

```suggestion
${dialog.type()}: ${dialog.message()} (default value: "${dialog.defaultValue()}").
```

Lightning00Blade: Also we should only add this if there is value I think not unconditionally as it only applies to `promp` atm,

OrKoN: makes sense, done

Issue #363, opened 2025-10-13
Accessibility tree/element(s) snapshot - exposing semantics, roles, states, ARIA,...


BogdanCerovac:
**Is your feature request related to a problem? Please describe.**
I miss more functionality to extract accessibility information from DevTools with MCP. One basic thing that would be a great start is perhaps an snapshot of the accessibility tree (probably as an extension to the "text-only" snapshot or similar.

**Describe the solution you'd like**
Add a feature for accessibility tree snapshot of the whole page or of a single selector (CSS, XPATH) or a group of selectors (CSS, XPATH) - so that I can use MCP to extract accessibility information (semantics, ARIA and everything that can be done by manual inspection).

**Describe alternatives you've considered**
Not sure if there are any alternatives possible. If they are - we need to know about them - perhaps documentation with accessibility in mind would be an alternative (especially if feature is already in place).

**Additional context**
MCP for DevTools with accessibility in mind would help developers, designers and content providers. Perhaps also assistive technologies that can use MCP.

OrKoN:
Note that the text snapshot is based on the a11y tree snapshot. We currently do not support selectors but you should already get the a11y tree information for the entire page. Could you clarify which specific a11y information do you miss from the snapshot?

BogdanCerovac:
Right, thank you very much, I haven't checked the code, before, just went through the docs. Now I've checked the code and see that, yes, thanks.

In https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/src/formatters/snapshotFormatter.ts#L32 I think we can see a narrowed down list of possibilities (limited list). That probably means that with using snapshot today don't get "the whole picture" (I didn't run the code to see what happens, just guessing).

This can lead to issues when we want to improve accessibility - limiting the context (also for AI).

Perhaps we can have a "non filtered" snapshot - to get absolutely all accessibility information? Without any filtering.

Second suggestion is to be able to ask for specific elements / parts of the tree, removing the need to drill down manually (and making it possible to save on tokens if needed or if we analyze gigantic trees). But I guess that one is not so vital.

Having an unfiltered snapshot of the whole "accessibility tree" would be a great start... Supporting also elements beyond text (landmarks etc.).

OrKoN:
@BogdanCerovac could you perhaps provide details about the use case you are after? because an alternative could be is that we offer something like lighthouse a11y audits instead of given all data to the LLM.

BogdanCerovac:
[...]

PR #369, opened 2025-10-13
refactor: simplify snapshot generation

OrKoN:
the only difference is that checked/pressed are only included if the element was checked/pressed.

Refs: https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/363

### DIFF
--- a/src/formatters/snapshotFormatter.ts
+++ b/src/formatters/snapshotFormatter.ts
@@ -28,69 +28,32 @@ function getAttributes(serializedAXNodeRoot: TextSnapshotNode): string[] {
     `"${serializedAXNodeRoot.name || ''}"`, // Corrected: Added quotes around name
   ];
 
-  // Value properties
-  const valueProperties = [
-    'value',
-    'valuetext',
-    'valuemin',
-    'valuemax',
-    'level',
-    'autocomplete',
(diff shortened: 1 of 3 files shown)


src/formatters/snapshotFormatter.ts
```diff
@@ -28,69 +28,32 @@ function getAttributes(serializedAXNodeRoot: TextSnapshotNode): string[] {
...
-  ] as const;
-  for (const property of valueProperties) {
-    if (
-      property in serializedAXNodeRoot &&
-      serializedAXNodeRoot[property] !== undefined
-    ) {
-      attributes.push(`${property}="${serializedAXNodeRoot[property]}"`);
-    }
-  }
+  const excluded = new Set(['id', 'role', 'name', 'elementHandle', 'children']);
```
AlinaVarkki, line 31: do we want to keep 'orientation'?

OrKoN: yeah, I believe we are keeping orientation as everything not on this list will be returned as an attribute.

AlinaVarkki: yeah, I meant to ask if we want to add it to the list

PR #375, opened 2025-10-14
feat: support stable id for network requests

Lightning00Blade:
This is part 1, we just expose the uid here, in a follow up PR we will be used for finding the request.

### DIFF
--- a/src/formatters/networkFormatter.ts
+++ b/src/formatters/networkFormatter.ts
@@ -15,7 +15,7 @@ export function getShortDescriptionForRequest(
   id: number,
 ): string {
   // TODO truncate the URL
-  return `uid ${id} - ${request.url()} ${request.method()} ${getStatusFromRequest(request)}`;
+  return `reqid ${id} - ${request.url()} ${request.method()} ${getStatusFromRequest(request)}`;
 }
 
 export function getStatusFromRequest(request: HTTPRequest): string {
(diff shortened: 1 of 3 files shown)


src/formatters/networkFormatter.ts
```diff
@@ -10,8 +10,12 @@ import type {HTTPRequest, HTTPResponse} from 'puppeteer-core';
...
 const BODY_CONTEXT_SIZE_LIMIT = 10000;
 
-export function getShortDescriptionForRequest(request: HTTPRequest): string {
-  return `${request.url()} ${request.method()} ${getStatusFromRequest(request)}`;
+export function getShortDescriptionForRequest(
+  request: HTTPRequest,
+  id: number,
+): string {
+  // TODO truncate the URL
+  return `uid ${id} - ${request.url()} ${request.method()} ${getStatusFromRequest(request)}`;
```
OrKoN: let's not call it `uid` to avoid mix up with the element ids. I would propose `reqid` for this?

tests/McpResponse.test.ts
```diff
@@ -423,11 +424,11 @@ No requests found.`,
         `# test response
 ## Network requests
 Showing 1-5 of 5 (Page 1 of 1).
-http://example.com GET [pending]
-http://example.com GET [pending]
-http://example.com GET [pending]
-http://example.com GET [pending]
-http://example.com GET [pending]`,
+reqid 1 - http://example.com GET [pending]
```
OrKoN, line 427: I think perhaps it makes sense to change the format to`reqid=$id <method> <url> <status>`?

Lightning00Blade: Sounds good let me send a follow up

PR #380, opened 2025-10-14
refactor: use different format for reqid

Lightning00Blade:


src/formatters/networkFormatter.ts
```diff
@@ -15,7 +15,7 @@ export function getShortDescriptionForRequest(
   id: number,
 ): string {
   // TODO truncate the URL
-  return `reqid ${id} - ${request.url()} ${request.method()} ${getStatusFromRequest(request)}`;
+  return `reqid=${id} ${request.url()} ${request.method()} ${getStatusFromRequest(request)}`;
```
OrKoN: lets put method first?

Lightning00Blade: So - ID Method URL, 
will update

PR #386, opened 2025-10-15
docs: add instructions for Qoder

hacke2:
Adds instructions on how to add Chrome DevTools MCP to [Qoder](https://qoder.com/).

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,15 @@ Or, from the IDE **Activity Bar** > `Kiro` > `MCP Servers` > `Click Open MCP Con
 </details>
 
 <details>
+  <summary>Qoder</summary>
+
+In **Qoder Settings**, go to `MCP Server` > `+ Add` > Use the configuration snippet provided above.
+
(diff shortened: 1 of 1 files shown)


hacke2:
> Thanks for the PR. Could you please add the configuration in the alphabetical order by the tool name?

Okay, we will iterate in subsequent versions~

hacke2:
> Thanks for the PR. Could you please add the configuration in the alphabetical order by the tool name?

I checked and it turned out that the sorting of tools was not working, which has been resolved in [this PR](https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/389)

Lightning00Blade:
I think what is meant here is that the things inside the `README.md` and `MCP Client configuration` should be sorted.
So `Qoder` should be between `Kiro` and `Visual Studio`.

hacke2:
> I think what is meant here is that the things inside the `README.md` and `MCP Client configuration` should be sorted. So `Qoder` should be between `Kiro` and `Visual Studio`.

Thank you for reminding me. I misunderstood. I have already submitted the code

OrKoN (approved):

PR #389, opened 2025-10-15
refactor: streamline tool registration by combining declaration and sorting

hacke2:
The sort of tools is not currently working.


Lightning00Blade:
This is not needed. In JavaScript the sort does the sorting in-place - [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort).
You can verify by running something like in Chrome DevTools.

image

PR #390, opened 2025-10-15
perf: improve evaluate_script tool prompt

ArcherGu:
I've noticed that AI sometimes defines a function and then immediately executes it, like this:
```js
function foo() { return "foo" }; foo()
```


Lightning00Blade:
Thanks for flagging this and the PR, I opened https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/392
Which should address the issue and provides a little bit better wording.

PR #409, opened 2025-10-17
refactor: bundle modelcontextprotocol-sdk

zyzyzyryxy:
Improve boot-up time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,6 @@
   "homepage": "https://github.com/ChromeDevTools/chrome-devtools-mcp#readme",
   "mcpName": "io.github.ChromeDevTools/chrome-devtools-mcp",
   "dependencies": {
-    "@modelcontextprotocol/sdk": "1.20.0",
     "core-js": "3.46.0",
     "debug": "4.4.3",
     "puppeteer-core": "^24.24.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -46,6 +46,9 @@
   },
   "devDependencies": {
     "@eslint/js": "^9.35.0",
```
OrKoN, line 47: let's move the sdk dependency to devDependencies

rollup.config.mjs
```diff
@@ -0,0 +1,88 @@
...
+        output: {
+          file: path.join(
+            'build',
+            'src',
+            'third_party',
+            'modelcontextprotocol-sdk',
+            'THIRD_PARTY_NOTICES',
+          ),
+          template(dependencies) {
+            const stringified_dependencies = dependencies.map(dependency => {
```
Lightning00Blade, line 59: I think there is something wrong with this code 
As currently the `build/src/third_party/mcp/THIRD_PARTY_NOTICES` is empty and I am expecting this to be filled with the MCP data.

OrKoN: Make sure you clean up your build folder

Lightning00Blade: Strange should work without this.

OrKoN: we do not clean the build right now in any way, I think some tools have the issues of skipping work if files are there.

OrKoN:
@Lightning00Blade @zyzyzyryxy updated the description to clarify that it intends to make boot-up time a bit faster and not the installation time.

Issue #88, opened 2025-09-23
Feature Request: Support Capturing Network Requests Across Page Navigation


thoughtentity:
Chrome DevTools MCP should support capturing network requests across page navigations. This would allow the MCP to capture network requests that occur during a redirect. This would enable additional debugging scenarios, such as allowing the LLM to observe and debug logout flows where the authentication cookies are cleared in a transient redirect response.

The current network tools only allow listing network requests for the currently loaded page, which means that transient network requests, such as redirects across pages, get missed.

neopal:
Your point is valid nevertheless classic workaround is to use evaluate script or build an extension and use console message

natorion:
Thank your for the feature request.

This could work similar as https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/docs/tool-reference.md#performance_start_trace, but for the network.

Edit: What would be the prompt you would use in your MCP client/AI agent to start such journey? Do you have example workload?

Sayvai:
Nice issue. I think it would also be great if we could also persist `console.log()` messages across page navigation / redirections, as well as persisting `network logs`.

sfc-gh-mochen:
my use case is to develop a script to complete oauth flow. i want AI to observe what requests are made to complete oauth

PR #411, opened 2025-10-17
refactor: store the last 3 navigations in PageCollector

Lightning00Blade:
Refactors the code to support https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/88

We keep each navigation in a separate array, that we can then expose to the user.

### DIFF
--- a/src/PageCollector.ts
+++ b/src/PageCollector.ts
@@ -191,9 +191,7 @@ export class NetworkCollector extends PageCollector<HTTPRequest> {
     // navigation request itself.
     // Keep the reference
     if (lastRequestIdx) {
-      const fromCurrentNavigation = requests.splice(
-        Math.min(lastRequestIdx, 0),
-      );
+      const fromCurrentNavigation = requests.splice(lastRequestIdx);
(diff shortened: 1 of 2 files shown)


src/PageCollector.ts
```diff
@@ -127,22 +134,28 @@ export class PageCollector<T> {
...
+    if (!navigations) {
       throw new Error('No requests found for selected page');
     }
 
-    for (const collected of data) {
-      if (collected[stableIdSymbol] === stableId) {
-        return collected;
+    for (const navigation of navigations) {
+      for (const collected of navigation) {
+        if (collected[stableIdSymbol] === stableId) {
```
OrKoN, line 156: would this already expose data from previous navigation by id? should it be marked as feature then?

Lightning00Blade: Yes it does, but I am planning on submitting the code in one release anyway.

Lightning00Blade:
Currently we cover that we can get the data properly formatted for the current navigation with an e2e test, so that should be sufficient to show we did not regress.
I will add e2e test with navigation once we fully expose the data to the MCP.

PR #414, opened 2025-10-17
refactor: bundle zod together with modelcontextprotocol/sdk

zyzyzyryxy:
Continue bundling dependencies for faster startup time.

### DIFF
--- a/src/third_party/modelcontextprotocol-sdk/index.ts
+++ b/src/third_party/modelcontextprotocol-sdk/index.ts
@@ -12,4 +12,4 @@ export {
   type ImageContent,
   type TextContent,
 } from '@modelcontextprotocol/sdk/types.js';
-export {default as z} from 'zod';
+export {z as zod} from 'zod';
--- a/src/tools/ToolDefinition.ts
+++ b/src/tools/ToolDefinition.ts
@@ -7,13 +7,15 @@
 import type {Dialog, ElementHandle, Page} from 'puppeteer-core';
 
 import type {TextSnapshotNode} from '../McpContext.js';
-import {z} from '../third_party/modelcontextprotocol-sdk/index.js';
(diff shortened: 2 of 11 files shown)


src/third_party/modelcontextprotocol-sdk/index.ts
```diff
@@ -12,3 +12,4 @@ export {
   type ImageContent,
   type TextContent,
 } from '@modelcontextprotocol/sdk/types.js';
+export {default as z} from 'zod';
```
Lightning00Blade: Let's bundle it separably or rename the folder that export this,
as it is it's own module.

package.json
```diff
@@ -40,8 +40,7 @@
     "core-js": "3.46.0",
     "debug": "4.4.3",
     "puppeteer-core": "^24.24.1",
-    "yargs": "18.0.0",
-    "zod": "^3.25.76"
```
Lightning00Blade, line 44: We should also explicitly add it to devDependecies and not rely on anything pulling it in for us

OrKoN: Zod is already bundled into the SDK so we do not need to bundle it separately again

Lightning00Blade: If you think this is better. I general  we should have just bundled and minify the whole project rather then just pieces.

src/third_party/modelcontextprotocol-sdk/index.ts
Lightning00Blade: Let's rename is zod for clarity

OrKoN:
```suggestion
export {z} from 'zod';
```

zyzyzyryxy: Done

PR #416, opened 2025-10-17
chore: clean up rollup config

Lightning00Blade:
Source maps were not provided for local builds.

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -1,7 +1,22 @@
 /**
- * @license
- * Copyright 2025 Google Inc.
- * SPDX-License-Identifier: Apache-2.0
+ * Copyright 2021 Google LLC.
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


rollup.config.mjs
```diff
@@ -37,7 +30,6 @@ const sdk = {
       comments: [/Copyright/i],
     }),
     license({
-      debug: true,
```
Lightning00Blade, line 40: Polluted the output.

rollup.config.mjs
```diff
@@ -1,19 +1,9 @@
 /**
- * Copyright 2021 Google LLC.
```
OrKoN, line 2: this is taken from chromium-bidi, please keep it

Lightning00Blade: I see

.github/workflows/publish-to-npm-on-tag.yml
```diff
@@ -76,6 +78,8 @@ jobs:
 
       - name: Build
         run: npm run build
+        env:
+          NODE_ENV: 'production'
```
OrKoN, line 82: why is it needed? would it cause minification?

Lightning00Blade: No I use it to control source maps what I use. See rollup.config.js

OrKoN: I see in that case it is fine but AFAIK we have no source map support at all in the project right now

OrKoN (changes requested):
Let's revert the license change, otherwise LGTM

OrKoN:
cc @zyzyzyryxy

PR #417, opened 2025-10-17
refactor: bundle puppeteer-core

zyzyzyryxy:
Continue speeding up boot time.

### DIFF
--- a/src/tools/script.ts
+++ b/src/tools/script.ts
@@ -5,7 +5,11 @@
  */
 
 import {zod} from '../third_party/modelcontextprotocol-sdk/index.js';
-import type {Frame, JSHandle, Page} from '../third_party/puppeteer-core/index.js';
+import type {
+  Frame,
+  JSHandle,
+  Page,
+} from '../third_party/puppeteer-core/index.js';
 
 import {ToolCategories} from './categories.js';
 import {defineTool} from './ToolDefinition.js';


src/third_party/puppeteer-core/index.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export type {
```
Lightning00Blade: This can be a * as it is type

zyzyzyryxy:
> Do we need to change the type imports? They should be interoperable and easier to manage with auto importing, we just need to ban non type imports from Puppeteer

Good point, type imports should work with devDependency too, same as with imports in tests. I'll update.

rollup.config.mjs
```diff
@@ -90,6 +100,11 @@ const sdk = {
...
     nodeResolve(),
   ],
-};
+});
 
-export default [sdk];
+export default [
+  bundleDependency('./build/src/third_party/modelcontextprotocol-sdk/index.js'),
+  bundleDependency('./build/src/third_party/puppeteer-core/index.js', {
+    inlineDynamicImports: true,
```
OrKoN: what imports would be inlined? I think we might need them for some node deps but in general it is mostly used for WebDriver BiDi so we can mark such dependencies as external.

rollup.config.mjs
```diff
@@ -90,6 +100,11 @@ const sdk = {
     json(),
     nodeResolve(),
   ],
-};
+});
 
-export default [sdk];
+export default [
+  bundleDependency('./build/src/third_party/modelcontextprotocol-sdk/index.js'),
+  bundleDependency('./build/src/third_party/puppeteer-core/index.js', {
```
OrKoN: let's remove puppeteer-core from dependencies (devDependencies already have puppeteer)

rollup.config.mjs
```diff
@@ -29,11 +29,22 @@ import license from 'rollup-plugin-license';
 
 const isProduction = process.env.NODE_ENV === 'production';
 
-/** @type {import('rollup').RollupOptions} */
-const sdk = {
-  input: './build/src/third_party/modelcontextprotocol-sdk/index.js',
+const allowed_licenses = [
```
OrKoN:
```suggestion
const allowedLicenses = [
```

src/McpContext.ts
```diff
@@ -91,9 +92,16 @@ export class McpContext implements Context {
   #nextSnapshotId = 1;
   #traceResults: TraceResult[] = [];
 
-  private constructor(browser: Browser, logger: Debugger) {
+  #locatorClass: typeof Locator;
+
+  private constructor(
+    browser: Browser,
+    logger: Debugger,
+    locatorClass: typeof Locator,
```
OrKoN, line 100: let's leave a comment why we are doing it

src/browser.ts
```diff
@@ -14,7 +14,8 @@ import type {
   LaunchOptions,
   Target,
 } from 'puppeteer-core';
```
OrKoN: let's also move the type imports

src/third_party/puppeteer-core/index.ts
zyzyzyryxy: Done

rollup.config.mjs
zyzyzyryxy: Done

rollup.config.mjs
zyzyzyryxy: Fixed

src/McpContext.ts
zyzyzyryxy: Done

src/browser.ts
zyzyzyryxy: Done except for the tests, similarly to mcp-sdk.

rollup.config.mjs
zyzyzyryxy: Dynamically imported:
   * ../node/NodeWebSocketTransport.js
   * ../common/BrowserWebSocketTransport.js
   * debug
   * ./bidi.js
   * ../bidi/bidi.js
I externalized bidi for now.

OrKoN (approved):

OrKoN:
@zyzyzyryxy  please mark conversations on the PR as resolved if they are resolved.

PR #418, opened 2025-10-17
test: stabilize test locally

Lightning00Blade:
Locally the test `with full page resulting in a large screenshot` would fail semi consistently.
With what I think is just things not getting loaded fully, so scroll to the bottom to fix that.

### DIFF
--- a/tests/tools/screenshot.test.ts
+++ b/tests/tools/screenshot.test.ts
@@ -74,7 +74,7 @@ describe('screenshot', () => {
       });
     });
 
-    it.only('with full page resulting in a large screenshot', async () => {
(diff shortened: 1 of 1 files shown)


tests/tools/screenshot.test.ts
```diff
@@ -74,12 +74,23 @@ describe('screenshot', () => {
       });
     });
 
-    it('with full page resulting in a large screenshot', async () => {
+    it.only('with full page resulting in a large screenshot', async () => {
```
OrKoN:
```suggestion
    it('with full page resulting in a large screenshot', async () => {
```

OrKoN (approved):

PR #419, opened 2025-10-17
feat: expose previous navigations to the MCP

Lightning00Blade:
We should consider changing the order of element shown in the from the collectors to be show newest first rather then the current oldest first approach.

Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/88

### DIFF
--- a/tests/tools/network.test.js.snapshot
+++ b/tests/tools/network.test.js.snapshot
@@ -15,6 +15,8 @@ Status:  [success - 200]
 - content-type:text/html; charset=utf-8
 - date:<long date>
 - keep-alive:timeout=5
+### Response Body
+<not available anymore>
 `;
 
 exports[`network > network_list_requests > list requests form current navigations only 1`] = `


src/tools/network.ts
```diff
@@ -63,12 +63,22 @@ export const listNetworkRequests = defineTool({
       .describe(
         'Filter requests to only return requests of the specified resource types. When omitted or empty, returns all requests.',
       ),
+    includePreviousNavigations: zod
```
OrKoN, line 65: I wonder if we should just make it a boolean? it is easier to consume for the LLM.

tests/PageCollector.test.ts
```diff
@@ -250,4 +250,40 @@ describe('NetworkCollector', () => {
     assert.equal(collector.getData(page)[0], navRequest);
     assert.equal(collector.getData(page)[1], request2);
   });
+
+  it('correctly picks up after multiple back to back navigations', async () => {
```
OrKoN, line 254: let's add a e2e test that involves actually formatting requests from multiple navigations

src/tools/network.ts
Lightning00Blade: I think that makes sense.

tests/PageCollector.test.ts
Lightning00Blade: I am not sure what the requirement here is? This test does actually format the request from multiple navigations
Can you please point me to what the test needs to do differently?

src/tools/network.ts
```diff
@@ -63,12 +63,20 @@ export const listNetworkRequests = defineTool({
...
+      .boolean()
+      .default(false)
+      .optional()
+      .describe(
+        'Number of previous navigations to include. Current navigation is always included.',
```
OrKoN: need to update the description as well

tests/tools/network.test.ts
```diff
@@ -21,6 +21,44 @@ describe('network', () => {
         assert.strictEqual(response.networkRequestsPageIdx, undefined);
       });
     });
+
+    it('list requests form current navigations only', async t => {
+      await withBrowser(async (response, context) => {
+        const page = await context.getSelectedPage();
+        await page.goto('data:text/html,<div>Hello 1</div>');
```
OrKoN: could we test with something that actually causes the network traffic and let's add a test verifying that the details can be also retrieved

tests/PageCollector.test.ts
OrKoN: It's not about this specific test: a general comment for better test coverage. This test covers the PageCollector but does not actually verify that we can format requests from earlier navigations successfully. E.g., bodies could be not available anymore and we need to make sure we handle that.

Lightning00Blade: I see you are talking about asking for a single request from a previous navigation.

src/tools/network.ts
```diff
@@ -62,12 +62,18 @@ export const listNetworkRequests = defineTool({
       .describe(
         'Filter requests to only return requests of the specified resource types. When omitted or empty, returns all requests.',
       ),
+    includePreviousNavigations: zod
+      .boolean()
+      .default(false)
+      .optional()
+      .describe('Whether to include request from previous navigations.'),
```
```suggestion
      .describe('Whether to include requests from previous navigations.'),
```

PR #425, opened 2025-10-18
fix: console messages filter missing

ArcherGu:
It seems we forgot it


AlinaVarkki:
Thank you for the change! We have, however, fixed it in:
https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/435

Therefore, I'll close this PR

PR #426, opened 2025-10-18
perf: improve options for sharing content

ArcherGu:
We should probably extract the option parameters for content that can be shared by multiple tools

### DIFF
--- a/src/tools/ToolDefinition.ts
+++ b/src/tools/ToolDefinition.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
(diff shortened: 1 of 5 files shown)


OrKoN (changes requested):
Thanks for the PR. Refactoring makes sense but we do not plan to make the snapshot configurable on each tool call right now. We might re-consider this later but for now could you please separate refactoring from adding new tool definitions.

ArcherGu:
> Thanks for the PR. Refactoring makes sense but we do not plan to make the snapshot configurable on each tool call right now. We might re-consider this later but for now could you please separate refactoring from adding new tool definitions.

I'm not quite sure what you mean? Should I close this PR?

PR #440, opened 2025-10-20
chore: improve console message format

OrKoN:
I checked Puppeteer implementation and it computes the text based on all args so we should treat the first arg in a special way. Also, changes the format to print count of args in the list and full args in the details only.

### DIFF
--- a/tests/formatters/consoleFormatter.test.ts
+++ b/tests/formatters/consoleFormatter.test.ts
@@ -12,7 +12,7 @@ import {
 } from '../../src/formatters/consoleFormatter.js';
 import type {ConsoleMessageData} from '../../src/McpResponse.js';
 
-describe.only('consoleFormatter', () => {
+describe('consoleFormatter', () => {
   describe('formatConsoleEventShort', () => {
     it('formats a console.log message', t => {
       const message: ConsoleMessageData = {


tests/formatters/consoleFormatter.test.ts
```diff
@@ -13,131 +12,84 @@ import {
 } from '../../src/formatters/consoleFormatter.js';
 import type {ConsoleMessageData} from '../../src/McpResponse.js';
 
-describe('consoleFormatter', () => {
+describe.only('consoleFormatter', () => {
```
OrKoN:
```suggestion
describe('consoleFormatter', () => {
```

tests/formatters/consoleFormatter.test.js.snapshot
```diff
@@ -0,0 +1,38 @@
...
+`;
+
+exports[`consoleFormatter > formatConsoleEventShort > formats a console.log message with one argument 1`] = `
+msgid=2 [log] Processing file: (1 args)
+`;
+
+exports[`consoleFormatter > formatConsoleEventVerbose > formats a console.error message 1`] = `
+ID: 4
+Message: error> Something went wrong
+
```
Lightning00Blade, line 16: I think somewhere we sneak 1 extra new line by accident with the new line.

Lightning00Blade: Count it as a `supernit`

OrKoN: let me look into it in a follow-up, will stacking some refactoring on top

PR #444, opened 2025-10-20
build: use correct Node types

Lightning00Blade:


Lightning00Blade:
Aaa because of the tester we can't do this...

PR #445, opened 2025-10-21
build: clean third_party build only

OrKoN:


package.json
```diff
@@ -6,8 +6,9 @@
   "bin": "./build/src/index.js",
   "main": "index.js",
   "scripts": {
-    "clean": "node -e \"require('fs').rmSync('build', {recursive: true, force: true})\"",
-    "build": "npm run clean && tsc && node --experimental-strip-types --no-warnings=ExperimentalWarning scripts/post-build.ts && rollup -c rollup.config.mjs",
+    "clean": "node -e \"require('fs').rmSync('build/src/third_party', {recursive: true, force: true})\"",
+    "clean:3p": "node -e \"require('fs').rmSync('build', {recursive: true, force: true})\"",
```
Lightning00Blade, line 10: I think this is flipped.

OrKoN: right, that's why it "worked". Actually it does not work... tsc does not seem to check the state of the output and only relies on the tsbuildinfo so this will not work.

PR #446, opened 2025-10-21
fix: indicate when request and response bodies are not available

OrKoN:


src/formatters/networkFormatter.ts
```diff
@@ -90,9 +89,8 @@ export async function getFormattedRequestBody(
       if (fetchData) {
         return `${getSizeLimitedString(fetchData, sizeLimit)}`;
       }
-    } catch {
-      // fetchPostData() call might fail with CDP exception, in this case we don't print anything in the context
-      return;
+    } catch (err) {
```
Lightning00Blade:
```suggestion
    } catch {
```

src/formatters/networkFormatter.ts
```diff
@@ -67,9 +67,8 @@ export async function getFormattedResponseBody(
     }
 
     return `<binary data>`;
-  } catch {
-    // buffer() call might fail with CDP exception, in this case we don't print anything in the context
-    return;
+  } catch (err) {
```
Lightning00Blade:
```suggestion
  } catch {
```

PR #447, opened 2025-10-21
chore(deps): bump puppeteer

OrKoN:


package.json
```diff
@@ -62,7 +62,7 @@
     "eslint-plugin-import": "^2.32.0",
     "globals": "^16.4.0",
     "prettier": "^3.6.2",
-    "puppeteer": "24.25.0",
+    "puppeteer": "24.26.0",
```
Lightning00Blade, line 65: Super Nit: given that we don't import `puppeteer-core` directly can you remove the changes to the `.dependabot` that split puppeteer of in a sperate bump?

PR #448, opened 2025-10-21
ci: create separate devDependency PRs for puppeteer and chrome-devtools-frontend

OrKoN:


.github/dependabot.yml
```diff
@@ -10,15 +10,13 @@ updates:
...
         dependency-type: production
-        exclude-patterns:
-          - 'puppeteer*'
-          - 'chrome-devtools-frontend'
         patterns:
           - '*'
       dev-dependencies:
         dependency-type: development
         exclude-patterns:
           - 'puppeteer*'
```
Lightning00Blade, line 18: Do we need to exclude Puppeteer,
We did it only because we need the same version of Pupeteer and Pupeteer-core.

OrKoN: would it include puppeteer into the default devDependencies PR if removed from exclude-patterns? or does it figure out that it is covered by another group?

Lightning00Blade: No its not smart enough.
My question is why do we need to have a separate group for puppeteer only.

OrKoN: It makes easier to review and make changes. Same as the devtools-frontend. It's also more critical to land these compared to other dev dependencies.

Lightning00Blade: OK

PR #450, opened 2025-10-21
refactor: bundle all dependencies together

zyzyzyryxy:
Final bundling step for improved startup times.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,6 @@
   },
   "homepage": "https://github.com/ChromeDevTools/chrome-devtools-mcp#readme",
   "mcpName": "io.github.ChromeDevTools/chrome-devtools-mcp",
-  "dependencies": {},
   "devDependencies": {
     "@eslint/js": "^9.35.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -37,11 +37,7 @@
   },
   "homepage": "https://github.com/ChromeDevTools/chrome-devtools-mcp#readme",
   "mcpName": "io.github.ChromeDevTools/chrome-devtools-mcp",
-  "dependencies": {
-    "core-js": "3.46.0",
-    "debug": "4.4.3",
-    "yargs": "18.0.0"
-  },
+  "dependencies": {},
```
Lightning00Blade: Remove all together

zyzyzyryxy: Done

OrKoN:
> LGTM with 1 comment.
> 
> @OrKoN Do you think it make sense to add a rule in ESLint (`no-restricted-imports`) that stops people from importing non-node packages?

probably would be nice to have it.

PR #458, opened 2025-10-23
feat: add a press_key tool

OrKoN:
Press key can be used to test shortcuts and individual key presses.

Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/246

### DIFF
--- a/src/utils/keyboard.ts
+++ b/src/utils/keyboard.ts
@@ -5,6 +5,7 @@
  */
 import type {KeyInput} from '../third_party';
 
+// See the KeyInput type for the list of supported keys.
 const validKeys = new Set([
   '0',
   '1',
@@ -273,7 +274,7 @@ function throwIfInvalidKey(key: string): KeyInput {
 }
 
 /**
- * Returns the key, followed by modifiers.
+ * Returns the primary key, followed by modifiers in original order.
  */
 export function parseKey(keyInput: string): [KeyInput, ...KeyInput[]] {
   let key = '';
@@ -287,7 +288,9 @@ export function parseKey(keyInput: string): [KeyInput, ...KeyInput[]] {
       key += ch;
     }
   }
-  if (key) result.push(throwIfInvalidKey(key));
+  if (key) {
+    result.push(throwIfInvalidKey(key));
+  }
 
   if (result.length === 0) {
     throw new Error(`Key ${keyInput} could not be parsed.`);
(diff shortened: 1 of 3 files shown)


src/utils/keyboard.ts
```diff
@@ -0,0 +1,301 @@
...
+
+  if (result.length === 0) {
+    throw new Error(`Key ${keyInput} could not be parsed.`);
+  }
+
+  if (new Set(result).size !== result.length) {
+    throw new Error(`Key ${keyInput} contains duplicate keys.`);
+  }
+
+  return result.reverse() as [KeyInput, ...KeyInput[]];
```
Lightning00Blade: Why do we reverse it here for then only to loop back in the first for loop in the tool?
I would prefer to have two `for of` loops with the one using `toReversed()` (Node v20.11 supported).
I makes the logic easier to follow as we get each key split
so `Ctrl+Shift+C` becomes ['Ctrl', 'Shift','C']

src/utils/keyboard.ts
```diff
@@ -0,0 +1,301 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import type {KeyInput} from '../third_party';
+
+const validKeys = new Set([
```
Lightning00Blade, line 9: Nit: is there a link or something we can point to where we gather this?

src/utils/keyboard.ts
OrKoN: It’s easier to consume (via de structuring) if the key is first and optional modifiers are following it. Modifiers need to be unpressed in reverse order for consistency. So in the result modifiers are in the order to key up and in the reverse order for key down.

OrKoN: Added toReversed and kept the input order.

src/utils/keyboard.ts
```diff
@@ -0,0 +1,302 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import type {KeyInput} from '../third_party';
```
Lightning00Blade:
```suggestion
import type {KeyInput} from '../third_party/index.js';
```

src/utils/keyboard.ts
```diff
@@ -0,0 +1,302 @@
...
+  for (const ch of keyInput) {
+    // Handle cases like Shift++.
+    if (ch === '+' && key) {
+      result.push(throwIfInvalidKey(key));
+      key = '';
+    } else {
+      key += ch;
+    }
+  }
+  if (key) result.push(throwIfInvalidKey(key));
```
Lightning00Blade: Let's use brackets here.

Lightning00Blade (approved):

src/utils/keyboard.ts
Lightning00Blade: This change was no applied but comment was resolved.

OrKoN: I think it was applied but I forgot to pull to the local branch.

PR #462, opened 2025-10-23
feat: support page reload

OrKoN:
- merges navigation tools into a single tool.
- adds reload functionality.
- in a follow-up we will add options to reload.


src/tools/pages.ts
```diff
@@ -142,18 +120,69 @@ export const navigatePageHistory = defineTool({
...
+    if (!request.params.type && !request.params.url) {
+      throw new Error('Either URL or a type is required.');
     }
 
+    if (!request.params.type) {
+      request.params.type = 'url';
+    }
+
+    await context.waitForEventsAfterAction(async () => {
+      try {
```
Lightning00Blade: I think it may make it easier to read to have a single switch and individual try catches as that way.
Logic related to each different type is close together.

Lightning00Blade (approved):
LGTM with 1 comment

src/tools/pages.ts
Ikee2463: Dwight Johnson,what property or line am I responsible fot

PR #466, opened 2025-10-24
refactor: connect to DevTools targets by default

OrKoN:
in preparation for integration with DevTools, this PR moves the pages filtering to the MCP context.

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -325,10 +325,10 @@ export class McpContext implements Context {
    */
   async createPagesSnapshot(): Promise<Page[]> {
     this.#pages = (await this.browser.pages()).filter(page => {
-      if (this.#options.devtools) {
-        return true;
+      if (page.url().startsWith('devtools://')) {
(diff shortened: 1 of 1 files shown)


src/McpContext.ts
```diff
@@ -315,7 +324,12 @@ export class McpContext implements Context {
    * Creates a snapshot of the pages.
    */
   async createPagesSnapshot(): Promise<Page[]> {
-    this.#pages = await this.browser.pages();
+    this.#pages = (await this.browser.pages()).filter(page => {
+      if (this.#options.devtools) {
+        return true;
+      }
+      return !page.url().startsWith('devtools://');
```
Lightning00Blade: I think this is a tiny bit clearer to what we want to do
```suggestion
      if (page.url().startsWith('devtools://')) {
        return this.#options.devtools;
      }
      return true;
```

OrKoN: I think the suggestion was actually incorrect (will send a follow-up).

PR #467, opened 2025-10-24
refactor: detect DevTools page for each page

OrKoN:


src/McpResponse.ts
```diff
@@ -347,8 +347,9 @@ Call ${handleDialog.name} to handle it before continuing.`);
       const parts = [`## Pages`];
       let idx = 0;
       for (const page of context.getPages()) {
+        const devToolsPage = context.getDevToolsPage(page);
         parts.push(
-          `${idx}: ${page.url()}${idx === context.getSelectedPageIdx() ? ' [selected]' : ''}`,
+          `${idx}: ${page.url()}${idx === context.getSelectedPageIdx() ? ' [selected]' : ''}${devToolsPage ? ' [has DevTools opened]' : ''}`,
```
OrKoN: actually let me think if we should land this or add a useful feature right away

src/McpContext.ts
```diff
@@ -82,6 +83,7 @@ export class McpContext implements Context {
 
   // The most recent page state.
   #pages: Page[] = [];
+  #pageToDevToolsPage = new Map<Page, Page>();
```
Lightning00Blade, line 86: We can turn this to a WeakMap

src/McpContext.ts
```diff
@@ -324,19 +326,50 @@ export class McpContext implements Context {
...
+    const allPages = await this.browser.pages();
+
+    this.#pages = allPages.filter(page => {
+      // If we allow debugging DevTools windows, return all pages.
+      if (this.#options.experimentalDevToolsDebugging) {
+        return true;
       }
-      return true;
+      // If we are in regular mode, the user should only see non-DevTools page.
+      return !page.url().startsWith('devtools://');
```
Lightning00Blade:
```suggestion
      if (page.url().startsWith('devtools://'){
        // If we are in regular mode, the user should only see non-DevTools page.
        return this.#options.experimentalDevToolsDebugging;
      }

      return true;
```

src/McpContext.ts
```diff
@@ -324,19 +326,50 @@ export class McpContext implements Context {
...
     });
+
+    await this.#detectOpenDevToolsWindows(allPages);
+
     return this.#pages;
   }
 
+  async #detectOpenDevToolsWindows(pages: Page[]) {
+    this.#pageToDevToolsPage = new Map<Page, Page>();
```
Lightning00Blade, line 346: No need to reset if we use a WeakMap.

src/McpContext.ts
OrKoN: no, the suggestion is wrong as I explained in the previous PR.

src/McpContext.ts
OrKoN: it does not matter, because DevTools pages are retained by this map as values and regular pages are retained by #pages

src/McpContext.ts
OrKoN: GC-lifetime of the Page does not match the lifetime of the snapshot so I prefer to keep this updated explicitly.

Lightning00Blade (approved):

src/McpContext.ts
OrKoN: refactored after an offline discussion

PR #477, opened 2025-10-27
feat: detect network requests inspected in DevTools UI

OrKoN:
If you manually inspect a network request in the DevTools UI, the Chrome DevTools MCP server is now able to detect it and allow you to refer to the selected request. The approach was tested in a couple of models.

In the list of the network requests we have a short indication about which request is selected in DevTools UI. We refer to the DevTools as "DevTools UI" to disambiguate from the DevTools MCP server.


src/PageCollector.ts
```diff
@@ -173,6 +173,22 @@ export class PageCollector<T> {
...
   }
+
+  find(page: Page, filter: (item: T) => boolean): T | undefined {
+    const navigations = this.storage.get(page);
+    if (!navigations) {
+      return;
+    }
+
+    for (const navigation of navigations) {
+      for (const collected of navigation) {
```
Lightning00Blade: Nit: How about we we do 
```suggestion
      for (const collected of navigation.filter(predicate)) {
```

src/PageCollector.ts
```diff
@@ -163,16 +163,32 @@ export class PageCollector<T> {
...
+    const navigations = this.storage.get(page);
+    if (!navigations) {
+      return;
+    }
+
+    for (const navigation of navigations) {
+      const item = navigation.find(filter);
+      if (item) {
+        return item;
+      }
```
Lightning00Blade, line 188:
```suggestion
      return item;
```

Lightning00Blade: Actually we can return the navigation.find directly.

OrKoN: I don't think we can?

Lightning00Blade: Aaa yes we are in a loop.

PR #487, opened 2025-10-28
feat: add tool to query Chrome UX Report

paulirish:
b/446630695


paulirish:
Not pursing this direction.

PR #494, opened 2025-10-29
refactor: merge emulate tools into one

OrKoN:
Merges emulateCpu and emulateNetwork into a single tool to reduce token consumption. Changes descriptions to be more concise.


src/tools/emulation.ts
```diff
@@ -15,73 +15,65 @@ const throttlingOptions: [string, ...string[]] = [
...
+      if (networkConditions === 'No emulation') {
+        await page.emulateNetworkConditions(null);
+        context.setNetworkConditions(null);
+        return;
+      }
+
+      if (networkConditions === 'Offline') {
+        await page.emulateNetworkConditions({
+          offline: true,
+          download: 0,
```
Lightning00Blade, line 56: Q: I guess this is CDP quirk? Should we try to handle this in the Puppeteer side?

OrKoN: yeah, I think we could. We need to migrate away from the current CDP implementation anyway. There were also fixes in https://chromium-review.googlesource.com/c/chromium/src/+/7055463 So maybe with M143 we can clean it up.

OrKoN: do you want to file an issue for Puppeteer linking https://chromium-review.googlesource.com/c/chromium/src/+/7055463 and https://chromium-review.googlesource.com/c/chromium/src/+/6827307? maybe we can add to the v25 as we could improve types here as well

OrKoN: filed https://github.com/puppeteer/puppeteer/issues/14387

PR #496, opened 2025-10-29
chore: new bug report template not showing up

sebastianbenz:
Removed JSON rendering for MCP configuration field.


Lightning00Blade:
Yeah this needs to be in the attributes part. Thanks!

Issue #497, opened 2025-10-29
get_network_request: improve get selected request description


sebastianbenz:
### Is your feature request related to a problem? Please describe.

Gemini CLI doesn't seem to understand our tool description for getting the selected element.

This doesn't work in the latest gemini-cli:

```
 Explain the selected network request

✦ I need a reqid to explain a specific network request. Please provide the reqid of the request you'd like me to explain. If you want to see all requests, you can use list_network_requests().
```

But this works:

```
> Explain the network request selected in the DevTools UI

──────────────────────────────────────────────────╮
│ ✓  get_network_request (chrome-devtools MCP Server) {}                                                                                                                                                                                                                                                                                         │
│```

### Describe the solution you'd like

n/a

### Describe alternatives you've considered

n/a

### Additional context

_No response_

PR #500, opened 2025-10-30
fix: improve get_network_request description

OrKoN:
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/497

### DIFF
--- a/src/tools/network.ts
+++ b/src/tools/network.ts
@@ -94,7 +94,7 @@ export const getNetworkRequest = defineTool({
       .number()
       .optional()
       .describe(
-        'The reqid of the network request. If omitted, looks up the current request selected in DevTools UI.',
+        'The reqid of the network request. If omitted returns the currently selected request in the DevTools Network panel.',
       ),
   },
   handler: async (request, response, context) => {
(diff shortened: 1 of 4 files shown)


OrKoN:
Screenshot 2025-10-30 at 08 10 58

it seems to perform worse if the model is downgraded but it seems to work in general.

src/tools/network.ts
```diff
@@ -84,8 +84,7 @@ export const listNetworkRequests = defineTool({
 
 export const getNetworkRequest = defineTool({
   name: 'get_network_request',
-  description: `Gets a network request by reqid. You can get all requests by calling ${listNetworkRequests.name}.
-Get the request currently selected in the DevTools UI by ommitting reqid`,
+  description: `Gets a network request by an optional reqid, if omitted returns the request selected in DevTools UI.`,
```
sebastianbenz:
```suggestion
  description: `Gets a network request by an optional reqid, if omitted returns the currently selected request in the DevTools network panel.`,
```

sebastianbenz: Would it make sense to make DevTools UI more clear?

OrKoN: Have you tried it? I think specifying the panel would be indeed nice. I think we should capitalize the word network?

```suggestion
  description: `Gets a network request by an optional reqid, if omitted returns the currently selected request in the DevTools Network panel.`,
```

OrKoN: PTAL again and please give it a try if you can.

sebastianbenz (approved):

PR #502, opened 2025-10-30
docs: Update installation link for Cursor MCP

alsotang:
move args from `command` prop to `args`, which would be more standardized.

from `{"command":"npx -y chrome-devtools-mcp@latest"}` to  `{"command": "npx","args": ["-y", "chrome-devtools-mcp@latest"]}`

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ Configure the following fields and press `CTRL+S` to save the configuration:
 
 **Click the button to install:**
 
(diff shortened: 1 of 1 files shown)


OrKoN:
@sebastianbenz could you please review?

sebastianbenz:
Thanks for creating this! I used Cursor's tool to generate the config and i'd prefer to stick with what their recommend way. I'll close the PR for now  (nevertheless I agree with your PR in that this is the more standard way).

PR #505, opened 2025-10-31
feat: report console issues in list_console_messages

nattallius:
BEGIN_COMMIT_OVERRIDE
chore: report console issues in list_console_messages
END_COMMIT_OVERRIDE

### DIFF
--- a/src/PageCollector.ts
+++ b/src/PageCollector.ts
@@ -76,7 +76,7 @@ export class PageCollector<T> {
   async init() {
     const pages = await this.#browser.pages(this.#includeAllPages);
     for (const page of pages) {
-      await this.addPage(page);
+      this.addPage(page);
     }
 
     this.#browser.on('targetcreated', async target => {
@@ -84,7 +84,7 @@ export class PageCollector<T> {
       if (!page) {
         return;
       }
-      await this.addPage(page);
+      this.addPage(page);
     });
     this.#browser.on('targetdestroyed', async target => {
       const page = await target.page();
@@ -245,7 +245,10 @@ export class ConsoleCollector extends PageCollector<
           data.issue satisfies Protocol.Audits.InspectorIssue;
         // @ts-expect-error Types of protocol from Puppeteer and CDP are incomparable for InspectorIssueCode, one is union, other is enum
         const issue = createIssuesFromProtocolIssue(null, inspectorIssue)[0];
-        if (!issue) return;
+        if (!issue) {
+          logger('No issue mapping for for the issue: ', inspectorIssue.code);
+          return;
+        }
 
         const seenKeys = this.#seenIssueKeys.get(page)!;
         const primaryKey = issue.primaryKey();
--- a/tests/PageCollector.test.ts
+++ b/tests/PageCollector.test.ts
@@ -429,7 +429,9 @@ describe('ConsoleCollector', () => {
 
     cdpSession.emit('Audits.issueAdded', {issue});
     cdpSession.emit('Audits.issueAdded', {issue});
-    const collectedIssue = collector.getData(page)[0];
+    const data = collector.getData(page);
+    assert.equal(data.length, 1);
+    const collectedIssue = data[0];
     assert(collectedIssue instanceof AggregatedIssue);
     assert.equal(collectedIssue.code(), 'MixedContentIssue');
     assert.equal(collectedIssue.getAggregatedIssuesCount(), 1);


test.txt
OrKoN, line 1: Please delete the file

scripts/post-build.ts
```diff
@@ -169,6 +169,19 @@ export const hostConfig = {};
...
+function copyDevToolsDescriptionFiles() {
+  const sourceDir = path.join(
+   process.cwd(),
+   'node_modules/chrome-devtools-frontend/front_end/models/issues_manager/descriptions',
+    );
+   const destDir = path.join(
+    BUILD_DIR,
+    'node_modules/chrome-devtools-frontend/front_end/models/issues_manager/descriptions',
+    );
+    fs.cpSync(sourceDir, destDir, {recursive: true});
```
OrKoN: is the destDir the same as sourceDir?

src/issue-descriptions.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+
+const DESCRIPTIONS_PATH = path.join(
+  process.cwd(),
```
OrKoN: we should not use the process.cwd() because the server could be start from any current working directory. Instead let's use `import.meta.dirname`

src/PageCollector.ts
```diff
@@ -113,12 +138,54 @@ export class PageCollector<T> {
     this.#listeners.set(page, listeners);
   }
 
+  protected async subscribeForIssues(page: Page) {
+    if (this instanceof NetworkCollector) {
```
OrKoN: Let's create a subclass for ConsoleCollector and invoke this by extending the addPage call. Let's also make sure we set up listeners first (by calling this.#initializePage) and then call `subscribeForIssues`

scripts/post-build.ts
nattallius: Yes, why not?

OrKoN: Overlooked that is cwd to build_dir

OrKoN:
There are conflicts with the main branch. Could you please rebase?

src/PageCollector.ts
```diff
@@ -40,11 +54,7 @@ export class PageCollector<T> {
   #listeners = new WeakMap<Page, ListenerMap>();
   #maxNavigationSaved = 3;
 
-  /**
-   * This maps a Page to a list of navigations with a sub-list
-   * of all collected resources.
-   * The newer navigations come first.
-   */
+
```
OrKoN: is there a reason for removing the comment?

src/PageCollector.ts
```diff
@@ -58,45 +68,58 @@ export class PageCollector<T> {
...
+      const currentNavigation = navigations[0];
+
+      // The aggregator sends the same object instance for updates, so we just
+      // need to ensure it's in the list.
+      if (!currentNavigation.includes(withId)) {
+        currentNavigation.push(withId);
+      }
+    };
+
+    // await this.subscribeForIssues(page);
```
OrKoN: remove the commented out code?

src/PageCollector.ts
```diff
@@ -61,45 +75,55 @@ export class PageCollector<T> {
...
-
-  #initializePage(page: Page) {
+  public async addPage(page: Page) {
     if (this.storage.has(page)) {
       return;
     }
+    await this.#initializePage(page);
+  }
 
+  async #initializePage(page: Page) {
```
OrKoN: let's try to keep these methods sync and handle subscription asynchronously in the ConsoleCollector sub-class (e.g., `void this.#subscribeToIssues()`).

src/PageCollector.ts
```diff
@@ -61,45 +75,55 @@ export class PageCollector<T> {
...
     const idGenerator = createIdGenerator();
     const storedLists: Array<Array<WithSymbolId<T>>> = [[]];
     this.storage.set(page, storedLists);
 
-    const listeners = this.#listenersInitializer(value => {
+    const collector = (value: T) => {
```
OrKoN: do we need changes here? can we change the emitting code to only emit issues once instead?

src/PageCollector.ts
```diff
@@ -194,6 +206,74 @@ export class PageCollector<T> {
...
+      const mockManager = this.#mockIssuesManagers.get(page);
+      if (mockManager) {
+        mockManager.dispatchEventToListeners(IssuesManagerEvents.ISSUE_ADDED, {
+          issue,
+          // @ts-expect-error We don't care that issues model is null
+          issuesModel: null,
+        });
+      }
+    });
+    await session.send('Audits.enable');
```
OrKoN: we need to catch possible errors when we enable the audits.

src/McpResponse.ts
```diff
@@ -295,6 +305,27 @@ export class McpResponse implements Response {
               ),
             };
           }
+          if (item instanceof AggregatedIssue) {
+            const count = item.getAggregatedIssuesCount();
+            const filename = item.getDescription()?.file;
+            const rawMarkdown = filename ? getIssueDescription(filename) : null;
+            if (!rawMarkdown) {
```
OrKoN: if there is not markdown title, let's skip the item because it is not very useful for the clients (and log a debug error)

src/McpResponse.ts
```diff
@@ -295,6 +305,27 @@ export class McpResponse implements Response {
...
+                message: `${item.code()} (count: ${count})`,
+                args: [],
+              };
+            }
+            const markdownAst = Marked.Marked.lexer(rawMarkdown);
+            const title = findTitleFromMarkdownAst(markdownAst);
+            return {
+              consoleMessageStableId,
+              type: 'issue',
+              message: `${title} (count: ${count})`,
```
OrKoN: let's add a new attribute called `count` instead of adding count inline in the message. Instead we should modify the console formatter to use the count attribute. We should also handle args for the issue type differently in the formatter.

src/formatters/consoleFormatter.ts
```diff
@@ -8,11 +8,15 @@ export interface ConsoleMessageData {
...
   type?: string;
   message?: string;
+  count?: number;
   args?: string[];
 }
 
 // The short format for a console message, based on a previous format.
 export function formatConsoleEventShort(msg: ConsoleMessageData): string {
+  if (msg.type === 'issue') {
+    return `msgid=${msg.consoleMessageStableId} [${msg.type}] ${msg.message} (count: ${msg.count}) (${msg.args?.length ?? 0} args)`;
```
OrKoN:
```suggestion
    return `msgid=${msg.consoleMessageStableId} [${msg.type}] ${msg.message} (count: ${msg.count})`;
```

src/McpResponse.ts
nattallius: Let's handle issues arguments in separate pr

src/PageCollector.ts
```diff
@@ -242,28 +243,36 @@ export class ConsoleCollector extends PageCollector<
...
-      if (mockManager) {
-        mockManager.dispatchEventToListeners(IssuesManagerEvents.ISSUE_ADDED, {
-          issue,
-          // @ts-expect-error We don't care that issues model is null
-          issuesModel: null,
-        });
-      }
-    });
-    await session.send('Audits.enable');
+    try {
```
OrKoN, line 241: try / catch is only needed around `await session.send('Audits.enable');`

src/PageCollector.ts
```diff
@@ -242,28 +243,36 @@ export class ConsoleCollector extends PageCollector<
...
+              // @ts-expect-error We don't care that issues model is null
+              issuesModel: null,
+            },
+          );
+        }
+      });
+      await session.send('Audits.enable');
+    } catch (e) {
+      const errorText = e instanceof Error ? e.message : JSON.stringify(e);
+      logger(`Error subscribing to issues: ${errorText}`);
```
OrKoN:
```suggestion
      logger(`Error subscribing to issues`, e);
```

would work as well?

src/PageCollector.ts
```diff
@@ -242,28 +243,36 @@ export class ConsoleCollector extends PageCollector<
...
+        if (!issue) {
+          return;
+        }
+        const seenKeys = this.#seenIssueKeys.get(page)!;
+        const primaryKey = issue.primaryKey();
+        if (seenKeys.has(primaryKey)) return;
+        seenKeys.add(primaryKey);
+
+        const mockManager = this.#mockIssuesManagers.get(page);
+        if (mockManager) {
```
OrKoN: let's inverse the if:

```
if (!mockManager) {return;}
```

src/PageCollector.ts
OrKoN: actually, never mind `createCDPSession` might need it too

src/McpResponse.ts
```diff
@@ -283,57 +284,58 @@ export class McpResponse implements Response {
...
+                logger(`no markdown ${filename} found for issue:` + item.code);
+                return null;
+              }
+              const markdownAst = Marked.Marked.lexer(rawMarkdown);
+              const title = findTitleFromMarkdownAst(markdownAst);
               return {
                 consoleMessageStableId,
                 type: 'issue',
-                message: `${item.code()} (count: ${count})`,
+                message: `${title}`,
```
```suggestion
                message: title,
```

src/McpResponse.ts
```diff
@@ -283,57 +284,58 @@ export class McpResponse implements Response {
...
+                : null;
+              if (!rawMarkdown) {
+                logger(`no markdown ${filename} found for issue:` + item.code);
+                return null;
+              }
+              const markdownAst = Marked.Marked.lexer(rawMarkdown);
+              const title = findTitleFromMarkdownAst(markdownAst);
               return {
                 consoleMessageStableId,
                 type: 'issue',
```
OrKoN, line 328: to simplify future work, can we also add the `item` to the console data item object?

OrKoN:
> Thanks! let's also add tests to make sure everything is working end-to-end and please address the lint and formatting issues.

I think we still need a e2e test similar to tests in tests/McpContext.test.ts in which we use Puppeteer to actually navigate to a page that has an issue. We can pick an issue that is easy to trigger via page.setContent().

src/PageCollector.ts
```diff
@@ -61,34 +76,33 @@ export class PageCollector<T> {
   async init() {
     const pages = await this.#browser.pages(this.#includeAllPages);
     for (const page of pages) {
-      this.#initializePage(page);
+      await this.addPage(page);
```
OrKoN:
```suggestion
      this.addPage(page);
```

src/PageCollector.ts
```diff
@@ -61,34 +76,33 @@ export class PageCollector<T> {
...
     }
 
     this.#browser.on('targetcreated', async target => {
       const page = await target.page();
       if (!page) {
         return;
       }
-      this.#initializePage(page);
+      await this.addPage(page);
```
OrKoN:
```suggestion
      this.addPage(page);
```

src/PageCollector.ts
```diff
@@ -61,34 +76,33 @@ export class PageCollector<T> {
...
+      this.cleanupPageDestroyed(page);
     });
   }
 
   public addPage(page: Page) {
-    this.#initializePage(page);
-  }
-
-  #initializePage(page: Page) {
     if (this.storage.has(page)) {
```
OrKoN: why did we move this check out of `#initializePage`?

src/PageCollector.ts
```diff
@@ -194,6 +207,82 @@ export class PageCollector<T> {
...
+  #seenIssueKeys = new WeakMap<Page, Set<string>>();
+  #issuesAggregators = new WeakMap<Page, IssueAggregator>();
+  #mockIssuesManagers = new WeakMap<Page, FakeIssuesManager>();
+
+  override addPage(page: Page): void {
+    if (this.storage.has(page)) {
+      return;
+    }
+    super.addPage(page);
+    void this.subscribeForIssues(page);
```
OrKoN:
```suggestion
    super.addPage(page);
    void this.subscribeForIssues(page);
```

src/PageCollector.ts
```diff
@@ -194,6 +207,82 @@ export class PageCollector<T> {
...
+  }
+  async subscribeForIssues(page: Page) {
+    if (!this.#seenIssueKeys.has(page)) {
+      this.#seenIssueKeys.set(page, new Set());
+    }
```
OrKoN:
```suggestion
    if (this.#seenIssueKeys.has(page)) {
      return;
    }
    this.#seenIssueKeys.set(page, new Set());
```

tests/PageCollector.test.ts
```diff
@@ -319,3 +341,110 @@ describe('NetworkCollector', () => {
...
+    await collector.init();
+
+    const issue = {
+      code: 'MixedContentIssue' as const,
+      details: {
+        mixedContentIssueDetails: {
+          insecureURL: 'test.url',
+        },
+      },
+    };
```
OrKoN:
```suggestion
    const issue = {
      code: 'MixedContentIssue' as const,
      details: {
        mixedContentIssueDetails: {
          insecureURL: 'test.url',
          resolutionStatus: 'MixedContentBlocked',
          mainResourceURL: ''
        },
      },
    } satisfies Protocol.Audits.InspectorIssue;
```

OrKoN: let's do the same for other tests

src/McpResponse.ts
OrKoN: I think you already handled them. Issues have no arguments.

tests/PageCollector.test.ts
```diff
@@ -319,3 +341,110 @@ describe('NetworkCollector', () => {
...
+
+    // @ts-expect-error Types of protocol from Puppeteer and CDP are incopatible for Issues but it's the same type
+    cdpSession.emit('Audits.issueAdded', {issue});
+    // @ts-expect-error Types of protocol from Puppeteer and CDP are incopatible for Issues but it's the same type
+    cdpSession.emit('Audits.issueAdded', {issue});
+    const collectedIssue = collector.getData(page)[0] as AggregatedIssue;
```
OrKoN:
```suggestion
    const collectedIssue = collector.getData(page)[0]!;
```

I think this would work too?

nattallius: I refactored test a little bit, so it doesn't use `as` anymore

OrKoN:
LGTM thanks!

Issue #492, opened 2025-10-29
Lots of "No page selected" errors


ivailo-uxify:
### Is your feature request related to a problem? Please describe.

Asking to browse multiple pages of a website and run analysis across all of them doesn't seem to work. Multiple tabs get opened in the browser but all of them are stuck at `about:blank` and there are a lot of `No page selected` errors.

Here is a link to the session: https://opencode.ai/s/7OZLfOWw

### Describe the solution you'd like

Ideally it should be able to correctly open all pages and run the requested analysis on each of them.

### Describe alternatives you've considered

Asking the same question one page at a time works.

OrKoN:
Interesting, we do serialize all tool calls on the server side so multiple pages should open fine (as well as other tools). I was not able to reproduce with gemini-cli that is known to issue multiple tool calls in paralellel. Is it specific to opencode? Could you please provide steps how to reproduce the issue? What's the version of chrome-devtools-mcp you are using?

sebastianbenz:
Just tried with opencode + claude and it worked flawlessly - can you maybe try with a different model?

ivailo-uxify:
Hi, tested with opencode + claude and it worked. Then I went back to the original setup of opencode + qwen3-235b-a22b-07-25 and now it also works that way. Weird, but I can no longer reproduce the issue 🤷‍♂️ 
Thanks!

rusakovic:
Confirm the issue with Claude Code CLI 2.0.31.

OrKoN:
We have found reproduction steps for this issue.

michaeldong:
the issue with  cursor IDE 2.5.25

PR #537, opened 2025-11-07
fix: avoid no page selected errors

OrKoN:
Uses object identity to track the selected page. This helps to be in sync with the browser state better.

Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/492

### DIFF
--- a/tests/tools/emulation.test.ts
+++ b/tests/tools/emulation.test.ts
@@ -74,7 +74,7 @@ describe('emulation', () => {
       });
     });
 
-    it.only('report correctly for the currently selected page', async () => {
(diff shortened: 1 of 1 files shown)


tests/tools/emulation.test.ts
```diff
@@ -74,9 +74,8 @@ describe('emulation', () => {
       });
     });
 
-    it('report correctly for the currently selected page', async () => {
+    it.only('report correctly for the currently selected page', async () => {
```
Lightning00Blade:
```suggestion
    it('report correctly for the currently selected page', async () => {
```

Issue #567, opened 2025-11-17
Add ability to inject script to run on page load


gernberg:
### Is your feature request related to a problem? Please describe.

I need the ability to inject scripts early on a page's lifecycle.

This means the `evaluate_script` tool is not useful for me as it executes too late.

### Describe the solution you'd like

Puppeteer has a method `page.evaluateOnNewDocument` that seems to do exactly what I want.

I did a small POC adding it as a param to navigate_page:

```ts
...
    initScript: zod.string().optional().describe(
      `(optional) A JavaScript function declaration to be executed by the tool on new document for every page load.
      `),
...
```

```ts
...
    if(request.params.initScript){
      page.evaluateOnNewDocument(request.params.initScript);
    }
...
```

### Describe alternatives you've considered

One approach would be to have a new "tool" but I think having it as an option on navigate makes more sense here as it's just enhancing that workflow.

OrKoN:
> I need the ability to inject scripts early on a page's lifecycle.

Could you please provide more details about the use case? How do you prompt your MCP client to use this feature?

gernberg:
Hey @OrKoN, 
Yeah - I understand it might not be a super standard thing I'm doing :)

My use case is based on that I'm often debugging pages where I don't have code access to the page itself.
In a manual workflow I would use the "override sources" feature in devtools, breakpoints and/or rely on some home built extensions that enhance DevTools + console. 

Prompt wise what I'm looking for is to be able to prompt "Before other scripts on the page execute, evaluate the following script(s) ..."

**One concrete example (simplified for readability):**
Let's say I want to figure out which CSS rule is causing issues with ads visibilty status at the time when the ad request happens.

In order to capture this I need to somehow have a breakpoint / something that captures state at a specific time during the page load. In a manual workflow I would typically use a breakpoint or have some additional logging added.

For chrome mcp / gemini based workflow breakpoints seems a bit complicated (at the moment at least).
So I added a feature that injects logging via an event listener, which give me something that works for most of my partners and gives additional context for gemini cli to work with.

Example script I might want to inject:
```javascript
  window.googletag = window.googletag || {cmd: []};
  window.googletag.pubads().addEventListener('slotRequested', e => {
    const slotId = e.slot.getSlotElementId();
    console.log("Ad slot requested: ", slotId, `checkVisibility=${document.getElementById(slotId).checkVisibility()}`);
  })
```

[...]

PR #568, opened 2025-11-17
feat: Add ability to inject script to run on page load

gernberg:
Adds a parameter "initScript" to be used to inject a script into the page load before other scripts execute.
[Page.evaluateOnNewDocument](https://pptr.dev/api/puppeteer.page.evaluateonnewdocument)

See #567

### DIFF
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -157,7 +157,7 @@
 
 - **handleBeforeUnload** (enum: "accept", "decline") _(optional)_: Whether to auto accept or beforeunload dialogs triggered by this navigation. Default is accept.
 - **ignoreCache** (boolean) _(optional)_: Whether to ignore cache on reload.
-- **initScript** (string) _(optional)_: (optional) A JavaScript function to be executed by the tool on new document for every page load before any other scripts.
+- **initScript** (string) _(optional)_: A JavaScript script to be executed on each new document before any other scripts for the next navigation.
 - **timeout** (integer) _(optional)_: Maximum wait time in milliseconds. If set to 0, the default timeout will be used.
 - **type** (enum: "url", "back", "forward", "reload") _(optional)_: Navigate the page by URL, back or forward in history, or reload.
 - **url** (string) _(optional)_: Target URL (only type=url)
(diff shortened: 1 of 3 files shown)


docs/tool-reference.md
```diff
@@ -149,6 +149,7 @@
 **Parameters:**
 
 - **ignoreCache** (boolean) _(optional)_: Whether to ignore cache on reload.
+- **initScript** (string) _(optional)_: (optional) A JavaScript function to be executed by the tool on new document for every page load before any other scripts.
```
OrKoN: thanks for the PR! I have asked a few details on the issue. Do you think the new_page tool could be a better place for the init script? Any thoughts how to remove a previously specified init script and if removing it is needed? For your use case, could it be better to pass an init script for all pages via a CLI argument or do you want the LLM to generate the init script?

gernberg: Haven't thought that much about the different approaches to be honest. 
I would be fine to just close this PR for now as it's very minimal (it's just a snapshot basically of what I'm playing with locally at the moment) and keep the discussions in the issues for now to land on a good  / better general solution for the problem I'm trying to solve :)

src/tools/pages.ts
```diff
@@ -132,6 +138,10 @@ export const navigatePage = defineTool({
       request.params.type = 'url';
     }
 
+    if (request.params.initScript) {
+      await page.evaluateOnNewDocument(request.params.initScript);
```
OrKoN: let's remove the script once the navigation is over (see https://pptr.dev/api/puppeteer.page.removescripttoevaluateonnewdocument)

src/tools/pages.ts
```diff
@@ -116,6 +116,12 @@ export const navigatePage = defineTool({
       .boolean()
       .optional()
       .describe('Whether to ignore cache on reload.'),
+    initScript: zod
+      .string()
+      .optional()
+      .describe(
+        '(optional) A JavaScript function to be executed by the tool on new document for every page load before any other scripts.',
```
```suggestion
        'A JavaScript function to be executed by the tool on the new document before any other scripts for the next page load.',
```

PR #594, opened 2025-11-21
chore: report console issues in get_console_messages 

nattallius:


src/formatters/consoleFormatter.ts
```diff
@@ -34,12 +40,19 @@ function getArgs(msg: ConsoleMessageData) {
 
 // The verbose format for a console message, including all details.
 export function formatConsoleEventVerbose(msg: ConsoleMessageData): string {
+  if (msg.item instanceof AggregatedIssue) {
```
OrKoN: let's add this condition to line 53 inline as only the message is different

src/McpResponse.ts
```diff
@@ -584,3 +572,37 @@ Call ${handleDialog.name} to handle it before continuing.`);
     this.#textResponseLines = [];
   }
 }
+
+function mapIssuesMessage(
+  message: AggregatedIssue,
```
```suggestion
  issue: AggregatedIssue,
```

src/McpResponse.ts
```diff
@@ -584,3 +572,37 @@ Call ${handleDialog.name} to handle it before continuing.`);
...
+  if (!markdownDescription) {
+    logger(`no description found for issue:` + message.code);
+    return null;
+  }
+  const rawMarkdown = filename ? getIssueDescription(filename) : null;
+  if (!rawMarkdown) {
+    logger(`no markdown ${filename} found for issue:` + message.code);
+    return null;
+  }
+  const processedMarkdown = MarkdownIssueDescription.substitutePlaceholders(
```
OrKoN: let's save processedMarkdown on the returned data, I guess we can call the field "details" so that we do not need to create markdown again during formatting.

tests/tools/console.test.ts
```diff
@@ -128,6 +128,12 @@ describe('console', () => {
   });
 
   describe('get_console_message', () => {
+    beforeEach(() => {
```
OrKoN: please create a new describe block for issues within this block because the `'gets a specific console message'` test does not depend on the flags.

src/formatters/consoleFormatter.ts
nattallius: I think it's better to leave it as it is, cause moving it there needs typescript handling of the _description_ and _item_

tests/tools/console.test.ts
```diff
@@ -149,5 +155,40 @@ describe('console', () => {
...
+
+    describe('issues type', () => {
+      beforeEach(() => {
+        setIssuesEnabled(true);
+      });
+      afterEach(() => {
+        setIssuesEnabled(false);
+      });
+
+      it('lists issues', async () => {
```
OrKoN:
```suggestion
      it('gets issue details', async () => {
```

?

tests/tools/console.test.ts
```diff
@@ -149,5 +155,40 @@ describe('console', () => {
...
+        await withBrowser(async (response, context) => {
+          const page = await context.newPage();
+          const issuePromise = new Promise<void>(resolve => {
+            page.once('issue', () => {
+              resolve();
+            });
+          });
+          await page.setContent('<input type="text" name="username" />');
+          await issuePromise;
+          await listConsoleMessages.handler({params: {}}, response, context);
```
OrKoN, line 173: we should not be re-using the same response object for multiple calls but I see that this already happened in existing tests. let's create a new response instance for the second call.

tests/tools/console.test.ts
```diff
@@ -149,5 +155,40 @@ describe('console', () => {
...
+          await getConsoleMessage.handler(
+            {params: {msgid: 1}},
+            response,
+            context,
+          );
+          const formattedResponse = await response.handle('test', context);
+          const textContent = formattedResponse[0] as {text: string};
+          assert.ok(
+            textContent.text.includes(
+              "Message: issue> An element doesn't have an autocomplete attribute",
```
OrKoN: let's assert that the title, description and other details are present

OrKoN: (this assertion looks like the list formatting)

tests/tools/console.test.ts
OrKoN: you can remove these beforeEach/afterEach after moving them to a separate describe.

OrKoN:
@nattallius could you please rebase and fix CI errors?

src/formatters/consoleFormatter.ts
OrKoN: I am not sure I got why it does not work. Does the following not work?

```
const isIssue = msg.item instanceof AggregatedIssue;

const result = [
    `ID: ${msg.consoleMessageStableId}`,
    `Message: ${msg.type}> ${isIssue ? formatIssue(msg.item, msg.description)} : msg.message}`,
    isIssue ? undefined: formatArgs(msg),
  ].filter(line => !!line);
```

nattallius: No, it doesn't satisfy typecheck and it still throws 
`Argument of type 'unknown' is not assignable to parameter of type 'AggregatedIssue'.`

src/formatters/consoleFormatter.ts
```diff
@@ -4,12 +4,15 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import {AggregatedIssue} from '../../node_modules/chrome-devtools-frontend/mcp/mcp.js';
+
 export interface ConsoleMessageData {
   consoleMessageStableId: number;
   type?: string;
   item?: unknown;
```
OrKoN:
```suggestion
  item?: AggregatedIssue;
```

src/formatters/consoleFormatter.ts
OrKoN:
```
  const aggregatedIssue = msg.item;
  const result = [
    `ID: ${msg.consoleMessageStableId}`,
    `Message: ${msg.type}> ${aggregatedIssue ? formatIssue(aggregatedIssue, msg.description) : msg.message}`,
    aggregatedIssue ? undefined: formatArgs(msg),
  ].filter(line => !!line);
```

this works I think (if the item type is defined).

OrKoN:
@nattallius since issues are not available by default yet, let's change the title of this PR to start with `chore:`.

PR #611, opened 2025-11-26
chore: report issues in details in get_console_mesasge 

nattallius:


tests/tools/console.test.ts
```diff
@@ -178,16 +178,39 @@ describe('console', () => {
...
+        await withBrowser(async (response, context) => {
+          const page = await context.newPage();
+          const issuePromise = new Promise<void>(resolve => {
+            page.once('issue', () => {
+              resolve();
+            });
+          });
+          await page.setContent(`
+            <script>
+              fetch('https://example.com/data.json', {
```
OrKoN: let's not user use real sites in our tests. You can use the test server instead.

OrKoN (changes requested):
Mostly looks good, left a comment about tests 👍

PR #613, opened 2025-11-26
docs: clarify browser-url parameter in README

### LINKED ISSUE #609: Document browserURL
### Description of the bug

The docs don't really say what `browserURL` **_is_**.

It's not a browserWSEndpoint (which I do know..) but what is it?  https://pptr.dev/api/puppeteer.connectoptions also doesn't describe it.

Is it just `http://localhost:9222` ?  

I'm really just asking for an example in the docs. :)

sebastianbenz:
Fixes #609

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,7 +322,7 @@ The Chrome DevTools MCP server supports the following configuration option:
 <!-- BEGIN AUTO GENERATED OPTIONS -->
 
 - **`--browserUrl`, `-u`**
-  Connect to a running Chrome instance using port forwarding. The value of this option should be the URL of the running Chrome instance together with the configured debugging port. `http://127.0.0.1:9222` is a common default. For more details see: https://github.com/ChromeDevTools/chrome-devtools-mcp?tab=readme-ov-file#connecting-to-a-running-chrome-instance.
+  Connect to a running, debuggable Chrome instance (e.g. `http://127.0.0.1:9222`). For more details see: https://github.com/ChromeDevTools/chrome-devtools-mcp#connecting-to-a-running-chrome-instance.
   - **Type:** string
 
 - **`--wsEndpoint`, `-w`**
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -11,7 +11,7 @@ export const cliOptions = {
   browserUrl: {
     type: 'string',
     description:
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -322,7 +322,7 @@ The Chrome DevTools MCP server supports the following configuration option:
 <!-- BEGIN AUTO GENERATED OPTIONS -->
 
 - **`--browserUrl`, `-u`**
-  Connect to a running Chrome instance using port forwarding. For more details see: https://developer.chrome.com/docs/devtools/remote-debugging/local-server.
```
paulirish, line 325: Aside: Does anyone know why  this Case 1 https://developer.chrome.com/docs/devtools/remote-debugging/local-server#port-forwarding-instance is documented?  The android port forwarding bit makes sense but  I don't know why you'd use one Chrome to debug another that's also running on the same machine.  (And I can't come up with good usecase for cross-LAN debugging but.. if this is for it.. it's not explained appropriately).  Asking because if nobody has a defense, I'll put up a CL to drop it.

README.md
```diff
@@ -322,7 +322,7 @@ The Chrome DevTools MCP server supports the following configuration option:
...
+  Connect to a running Chrome instance using port forwarding. The value of this option should be the URL of the running Chrome instance together with the configured debugging port. `http://127.0.0.1:9222` is a common default. For more details see: https://github.com/ChromeDevTools/chrome-devtools-mcp?tab=readme-ov-file#connecting-to-a-running-chrome-instance.
```
paulirish: Thanks Sebastian for this!  And yes while this flag is discussed lower down in the readme, I didn't see that. I only looked here and in --help. So thanks for that. :)


A few thoughts

1) I don't know where the port forwarding bit came from but that feels like a distraction. My usecase aligns with what's described in #connecting-to-a-running-chrome-instance .. and that's just a local Chrome.  
2) It might be clearer to call this an "origin" as `http://localhost:port` is exactly that.  But honestly with the example inlined, it doesnt matter too much.

Maybe this?

```suggestion
  Connect to a running, debuggable Chrome instance (eg `http://127.0.0.1:9222`). For more details see: https://github.com/ChromeDevTools/chrome-devtools-mcp#connecting-to-a-running-chrome-instance.
```

README.md
sebastianbenz: I don't have any context on this. Maybe @OrKoN has.

OrKoN: Maybe for Docker/VM setups? not sure either. Maybe just to showcase that you can debug desktop Chrome remotely and not only Android? I think the docs could be made more useful if they cover Docker/VM setup.

sebastianbenz: Filed content bug for this b/465403797

OrKoN (approved):

Issue #617, opened 2025-11-27
If puppeteer connections fails, the server returns [object Object]


OrKoN:
### Description of the bug

Have a browserUrl/wsEndpoint that does not exist and try to connect.

### Chrome DevTools MCP version

0.10.1

### Operating system

None

### Extra checklist

- [ ] I want to provide a PR to fix this bug

PR #618, opened 2025-11-27
fix: handle error messages that are not instanceof Error

OrKoN:
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/617

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -142,29 +142,22 @@ function registerTool(tool: ToolDefinition): void {
           response,
           context,
         );
-        try {
-          const content = await response.handle(tool.name, context);
-          return {
-            content,
-          };
-        } catch (error) {
-          logger(`${tool.name} response handling error:`, error, error.stack);
-          const errorText =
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -149,22 +149,20 @@ function registerTool(tool: ToolDefinition): void {
           };
         } catch (error) {
```
Lightning00Blade: I think we can remove this try/catch block
It only adds a extra log, with a little bit more context, but we should have that via stack?

src/main.ts
```diff
@@ -149,22 +149,20 @@ function registerTool(tool: ToolDefinition): void {
...
-                type: 'text',
-                text: errorText,
-              },
-            ],
-            isError: true,
-          };
+          throw error;
         }
       } catch (err) {
         logger(`${tool.name} error:`, err, err.stack);
```
Lightning00Blade: Because `throw undefined` is a thing...
```suggestion
        logger(`${tool.name} error:`, err, err?.stack);
```

PR #621, opened 2025-11-28
chore: add UniverseManager to create DevTools Universes for each Page

szuend:
A follow-up PR will add a single instance of the `UniverseManager` to `McpContext` similarly to the *Collectors. We don't do that in this PR yet, as we need to wait for some chrome-devtools-frontend changes to roll.

### DIFF
--- a/src/DevtoolsUtils.ts
+++ b/src/DevtoolsUtils.ts
@@ -8,6 +8,10 @@ import {
   type Issue,
   type AggregatedIssue,
   type IssuesManagerEventTypes,
+  type Target,
+  DebuggerModel,
+  Foundation,
+  TargetManager,
   MarkdownIssueDescription,
   Marked,
   ProtocolClient,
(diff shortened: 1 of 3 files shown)


src/DevtoolsUtils.ts
```diff
@@ -134,3 +145,116 @@ I18n.DevToolsLocale.DevToolsLocale.instance({
...
+  readonly #createUniverseFor: TargetUniverseFactoryFn;
+  readonly #includeAllPages: boolean;
+  readonly #universes = new WeakMap<Page, TargetUniverse>();
+
+  /** Guard access to #universes so we don't create unnecessary universes */
+  readonly #mutex = new Mutex();
+
+  constructor(
+    browser: Browser,
+    includeAllPages?: boolean,
```
OrKoN: can we pass initial pages instead of making it fetch on its own?

src/DevtoolsUtils.ts
```diff
@@ -134,3 +145,116 @@ I18n.DevToolsLocale.DevToolsLocale.instance({
...
+      this.#universes.set(page, await this.#createUniverseFor(page));
+    } finally {
+      this.#mutex.release();
+    }
+  };
+
+  #onTargetDestroyed = async (target: PuppeteerTarget) => {
+    const page = await target.page();
+    try {
+      await this.#mutex.acquire();
```
OrKoN, line 223: does this need to be mutexed? it looks sync

OrKoN: ah I see, you want to guard against the setup step running slow

src/DevtoolsUtils.ts
szuend: Should we pass them to `init` instead or via ctor?

OrKoN: This CL does not include where it is instantiated but I think if it is where the MCPContext is initialized, then yes

szuend: Done

PR #631, opened 2025-11-28
chore: skip all pauses in all DevTools universes

szuend:


paulirish:
@szuend 

how soon are you expecting to add this in src/McpContext.ts  ?

```ts
    await this.#universeManager.init(pages);
```

I ask because the crux-data feature would certainly benefit from reusing devtools frontend code, but that requires proper targets etc.  In my branch all this does seem to work but I was only testing my stuff. :)

szuend:
@paulirish Should work, just not done yet because we haven't added stack traces to console messages yet so there is no user for it. But feel free to already enable it if you need DevTools things. I'm not aware of any blockers but there could be.

Issue #296, opened 2025-10-07
Emulate Geolocation


sebastianbenz:
**Is your feature request related to a problem? Please describe.**
It'd be great to have the ability to emulate the geo location, e.g.: 

```
Why does the map not show our store in Munich when the current location is set to : 48°8'14.75"N, 11°34'31.76"E?
```

**Describe the solution you'd like**
Puppeteer already supports it via: https://pptr.dev/api/puppeteer.page.setgeolocation

natorion:
what about the other stuff in Sensors?

waterbang:
that's a really good idea

OrKoN:
We already have this.

PR #634, opened 2025-11-28
feat(emulation): add geolocation emulation tool

JustasMonkev:
Adding emulate_geolocation tool to emulate device location for testing location-based features. Supports setting latitude/longitude or clearing the override. 

Closing https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/296

### DIFF
--- a/src/tools/emulation.ts
+++ b/src/tools/emulation.ts
@@ -37,36 +37,28 @@ export const emulate = defineTool({
       .describe(
         'Represents the CPU slowdown factor. Set the rate to 1 to disable throttling. If omitted, throttling remains unchanged.',
       ),
-    latitude: zod
-      .number()
-      .min(-90)
-      .max(90)
+    geolocation: zod
+      .object({
+        latitude: zod
+          .number()
+          .min(-90)
+          .max(90)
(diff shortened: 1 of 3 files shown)


OrKoN (changes requested):
Thanks for the PR. Please integrate the geolocation functionality into the emulate tool instead of a new tool to keep the number of tools managable.

JustasMonkev:
please check again @OrKoN

src/tools/emulation.ts
```diff
@@ -37,31 +37,50 @@ export const emulate = defineTool({
       .describe(
         'Represents the CPU slowdown factor. Set the rate to 1 to disable throttling. If omitted, throttling remains unchanged.',
       ),
+    latitude: zod
```
OrKoN: let's create an object called `geolocation` to hold geolocation related properties. Can we make it so that setting it to null resets the emulation? I do not want to have a separate flag to just clear the emulation. "Must be provided together" would not be needed as well, as we can make the lat/long non-optional in the new object.

JustasMonkev: please recheck again

PR #639, opened 2025-12-02
feat: add performance_analyze_file tool

miyanaga:
This adds a new tool `performance_analyze_file` that lets you analyze trace files from the local filesystem.

## Why
Currently, you can only analyze traces recorded live with `performance_start_trace`. But sometimes you already have trace files from Lighthouse, Chrome DevTools export, or other sources that you want to analyze.

## What
- Added `performance_analyze_file` tool that takes a file path and parses the trace
- The analyzed trace is stored so you can use `performance_analyze_insight` to get details
- Added tests for success, file not found, and invalid JSON cases

## Testing
```
npm run test
```
All tests pass.


miyanaga:
I have signed the CLA.

OrKoN:
cc @natorion

natorion:
I think this is useful as part of an automation. 

I would suggest naming it performance_analyze_trace or something. Its a tracefile after all. I am not sure if we need it enabled/available per default though.

src/tools/performance.ts
```diff
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import fs from 'node:fs/promises';
```
destyriyan511-pixel, line 7: [](url)

OrKoN:
Closing the PR as stale.

PR #650, opened 2025-12-08
fix: use page.setViewport instead of page.resize

### LINKED ISSUE #645: resize_page fails with Protocol error: Browser.setContentsSize wasn't found
## Description
When calling `resize_page` to test mobile viewports, the command fails with a protocol error.

## Error Message
```
Protocol error (Browser.setContentsSize): 'Browser.setContentsSize' wasn't found
```

## Environment
- Chrome DevTools MCP version: Latest (installed via Claude Code)
- Chrome version: Unknown (remote debugging)
- Platform: Linux (WSL2)
- Claude Code: CLI

## Steps to Reproduce
1. Navigate to any page using `navigate_page`
2. Call `resize_page` with mobile dimensions:
```
resize_page(width=414, height=896)
```
3. Observe protocol error

## Expected Behavior
Page viewport should resize to specified dimensions for mobile testing.

## Actual Behavior
Command fails with `Browser.setContentsSize` not found error.

## Workaround
Currently using `emulate` for CPU/network throttling, but cannot test mobile viewport sizes.

## Additional Context
This appears to be a Chrome DevTools Protocol compatibility issue - `Browser.setContentsSize` may not be available in all Chrome versions or may have been deprecated.

---
Filed from Claude Code session

yasirusman85:
## Description
Fixes #645.

The `resize_page` tool was failing with a `Protocol error (Browser.setContentsSize): 'Browser.setContentsSize' wasn't found` when running against certain Chrome environments (e.g., remote debugging). This was because it relied on an internal `page.resize` method that uses `Browser.setContentsSize`, which is not universally supported.

This PR replaces the internal `page.resize` call with the standard Puppeteer `page.setViewport` method. This uses `Emulation.setDeviceMetricsOverride` under the hood, which is the standard and supported way to resize the viewport for testing and emulation.

## Changes
- Replaced `page.resize` with `page.setViewport` in [src/tools/pages.ts](cci:7://file:///c:/Users/Home/Desktop/code/Chrom-dev-tools/src/tools/pages.ts:0:0-0:0).

## Verification
- Verified that `resize_page` now correctly calls `page.setViewport` with the provided dimensions.

### DIFF
--- a/src/tools/pages.ts
+++ b/src/tools/pages.ts
@@ -207,10 +207,9 @@ export const resizePage = defineTool({
   handler: async (request, response, context) => {
     const page = context.getSelectedPage();
 
-    // @ts-expect-error internal API for now.
-    await page.resize({
-      contentWidth: request.params.width,
-      contentHeight: request.params.height,
+    await page.setViewport({
+      width: request.params.width,
+      height: request.params.height,
     });
 
     response.setIncludePages(true);


OrKoN:
Thanks for the PR. We are not using set viewport because of the issues with visual rendering that can arise when the window size does not match the viewport. We will keep using the resize for now.

PR #656, opened 2025-12-09
build: bundle devtools frontend

OrKoN:


rollup.config.mjs
```diff
@@ -95,6 +96,57 @@ const bundleDependency = (
...
+
+            // Manual license handling for chrome-devtools-frontend third_party
+            const tsConfig = JSON.parse(
+              fs.readFileSync(
+                path.join(process.cwd(), 'tsconfig.json'),
+                'utf-8',
+              ),
+            );
+            const thirdPartyDirectories = tsConfig.include.filter(location => {
+              return location.includes(
```
zyzyzyryxy: Super nit: could shorten to just `location => location.includes(.)`

rollup.config.mjs
```diff
@@ -95,6 +96,57 @@ const bundleDependency = (
...
+
+            for (const thirdPartyDir of thirdPartyDirectories) {
+              const fullPath = path.join(process.cwd(), thirdPartyDir);
+              const licenseFile = path.join(fullPath, 'LICENSE');
+              if (fs.existsSync(licenseFile)) {
+                const name = path.basename(thirdPartyDir);
+                manualLicenses.push(
+                  [
+                    `Name: ${name}`,
+                    `License: Custom`,
```
zyzyzyryxy: I wonder if we should attempt detecting type of the license? third-party-web seems to be MIT, legacy-javascript - Apache 2.0 etc.

package.json
```diff
@@ -16,11 +16,12 @@
...
-    "test:only": "npm run build && node --require ./build/tests/setup.js --no-warnings=ExperimentalWarning --test-reporter spec --test-force-exit --test --test-only \"build/tests/**/*.test.js\"",
-    "test:only:no-build": "node --require ./build/tests/setup.js --no-warnings=ExperimentalWarning --test-reporter spec --test-force-exit --test --test-only \"build/tests/**/*.test.js\"",
-    "test:update-snapshots": "npm run build && node --require ./build/tests/setup.js --no-warnings=ExperimentalWarning --test-force-exit --test --test-update-snapshots \"build/tests/**/*.test.js\"",
+    "test:node20": "node --import ./build/tests/setup.js --test-reporter spec --test-force-exit --test build/tests",
+    "test:no-build": "node --import ./build/tests/setup.js --no-warnings=ExperimentalWarning --experimental-print-required-tla --test-reporter spec --test-force-exit --test \"build/tests/**/*.test.js\"",
+    "test": "npm run build && node --import ./build/tests/setup.js --no-warnings=ExperimentalWarning --experimental-print-required-tla --test-reporter spec --test-force-exit --test \"build/tests/**/*.test.js\"",
```
zyzyzyryxy: Nit: Wouldn't it be better to have "test" run build then test:no-build to reduce duplication? Same for test:only

src/third_party/index.ts
```diff
@@ -35,3 +35,31 @@ export {
...
+
+export {
+  AgentFocus,
+  TraceEngine,
+  PerformanceTraceFormatter,
+  PerformanceInsightFormatter,
+  // Issue is a type in some contexts, but let's export it as is if it's a class or if we can use export type?
+  // Use export type for pure types if possible, or just export { ... }
+  // The users act as if they are values (classes) mostly, except explicit type imports.
+  // Re-exporting everything as value allows both (if they are values).
```
zyzyzyryxy: This comment confuses me a bit. "Issue is a type in some contexts, but let's export it as is" yet "type Issue" below?
Is it about AggregatedIssue instead?

rollup.config.mjs
```diff
@@ -95,6 +96,57 @@ const bundleDependency = (
...
+                    '',
+                    fs.readFileSync(licenseFile, 'utf-8').replaceAll('\r', ''),
+                  ].join('\n'),
+                );
+              }
+            }
+
+            if (manualLicenses.length > 0) {
+              stringified_dependencies.push(...manualLicenses);
```
Lightning00Blade: Super nit: Style guide dictate this needs to be cameCase

rollup.config.mjs
Lightning00Blade: I think we must do that.

package.json
OrKoN: test:only can runs a slightly different command. We could probably spend time to figure how to forward extra args so that also `--` forwarding keeps working. But I think it's fine for now. Changed test to run test:no-build.

rollup.config.mjs
OrKoN: Tbh we have no obligation to classify licenses, we should just include the content. If it does not cause any violations in tools, I would just drop `Custom`.

package.json
zyzyzyryxy: I see test and test:only are different, I meant maybe test:only should call test:only:no-build. The commands are identical now, sans the build.

OrKoN: I think I have not changed test:only and test:only:no-build in this CL.

PR #685, opened 2025-12-16
feat(screenshot): add full-page screenshot support for iframes

absarasw:
Add ability to capture full-page screenshots of iframe content by
temporarily expanding the iframe to reveal all scrollable content.

Changes:
- Add `iframeUid` parameter to explicitly target an iframe for capture
- Auto-detect main content iframe when `fullPage=true` is used
- Temporarily expand iframe dimensions to capture full scrollable content
- Restore original iframe styles after screenshot capture
- Add proper element handle disposal for better memory management

This enables capturing complete iframe content (like embedded editors or
content frames) that would otherwise be clipped to the visible viewport.


OrKoN:
cc @natorion wdyt?

src/tools/screenshot.ts
```diff
@@ -41,7 +224,13 @@ export const screenshot = defineTool({
       .boolean()
       .optional()
       .describe(
-        'If set to true takes a screenshot of the full page instead of the currently visible viewport. Incompatible with uid.',
+        'If set to true takes a screenshot of the full page instead of the currently visible viewport. Incompatible with uid unless iframeUid is also provided.',
+      ),
+    iframeUid: zod
```
OrKoN: can this be done without introducing a new parameter? e.g., detect if uid points to an iframe element?

src/tools/screenshot.ts
```diff
@@ -51,41 +240,93 @@ export const screenshot = defineTool({
...
+      screenshot = await page.screenshot({
+        type: format,
+        fullPage: false,
+        quality,
+        optimizeForSpeed: true,
+      });
+      responseMessage = "Took a screenshot of the current page's viewport.";
     }
 
+    response.appendResponseLine(responseMessage);
```
OrKoN, line 557: would it be possible to add test coverage including local and out-of-process iframes?

natorion:
> cc @natorion wdyt?

I think this is fine, but the warning should be also highlight potential other issues, no?

absarasw:
@OrKoN , @natorion , can you please review changes again?

tests/tools/screenshot.test.ts
```diff
@@ -260,5 +260,141 @@ describe('screenshot', () => {
...
+          const page = context.getSelectedPage();
+          await page.setContent(fixture.html);
+
+          await screenshot.handler(
+            {params: {format: 'png', fullPage: true}},
+            response,
+            context,
+          );
+
+          assert.equal(response.images.length, 1);
```
OrKoN, line 277: is there a way to verify that the image is actually correct?

src/tools/screenshot.ts
```diff
@@ -10,6 +10,403 @@ import type {ElementHandle, Page} from '../third_party/index.js';
 import {ToolCategory} from './categories.js';
 import {defineTool} from './ToolDefinition.js';
 
+/**
+ * Takes a full-page screenshot of an iframe's content by temporarily
+ * expanding the iframe to show all scrollable content.
+ */
+async function takeIframeFullPageScreenshot(
```
OrKoN, line 17: This seems to change the iframe content to take the screenshot. I think we should find a way to do it with CDP to minimize the need for checking the content. Changing iframe styles might affect the content that is being screenshoted.

OrKoN:
Closing the PR as stale. The FR affects the web content by changing its DOM using script execution. This is potentially very fragile and we won't accept it now. Please file a feature request if this is still relevant for your workflows so that we can re-prioritize and perhaps implement native CDP support for this.

PR #692, opened 2025-12-18
feat: support adding extra headers to mcp for each request

clChenLiang:
In test scenarios, 
the test environment of the production domain is typically used, which is usually distinguished by headers. 
This merge request supports adding headers parameters when accessing the test environment of a project.


clChenLiang:
can @OrKoN @Lightning00Blade to see this pr ?

src/McpContext.ts
```diff
@@ -127,8 +132,11 @@ export class McpContext implements Context {
     this.logger = logger;
     this.#locatorClass = locatorClass;
     this.#options = options;
+    this.#headers = options.headers;
 
-    this.#networkCollector = new NetworkCollector(this.browser);
+    this.#networkCollector = new NetworkCollector(this.browser, undefined, {
```
Lightning00Blade, line 137: This should not happen in the NetworkCollector. Collectors are there to get "collect" data from the pages as things happen. The code should live inside the McpContext class.

tests/McpContext.test.ts
```diff
@@ -102,3 +102,22 @@ describe('McpContext', () => {
...
+      // Verify context was created successfully
+      assert.ok(context);
+      
+      // Test that we can make a request (headers should be applied if any)
+      const navigationPromise = page.goto('data:text/html,<html><body>Test</body></html>');
+      await navigationPromise;
+      
+      // If we reach here without errors, headers functionality is working
+      assert.ok(true);
+    }, { debug: false });
```
Lightning00Blade, line 121:
```suggestion
    });
```

tests/McpContext.test.ts
```diff
@@ -102,3 +102,22 @@ describe('McpContext', () => {
...
+    await withMcpContext(async (_response, context) => {
+      const page = context.getSelectedPage();
+      await page.setContent('<html><body>Test page</body></html>');
+      
+      // Verify context was created successfully
+      assert.ok(context);
+      
+      // Test that we can make a request (headers should be applied if any)
+      const navigationPromise = page.goto('data:text/html,<html><body>Test</body></html>');
+      await navigationPromise;
```
Lightning00Blade, line 117:
```suggestion
      await page.goto('data:text/html,<html><body>Test</body></html>');
```

tests/McpContext.test.ts
```diff
@@ -102,3 +102,22 @@ describe('McpContext', () => {
     );
   });
 });
+
+describe('McpContext headers functionality', () => {
+  it('works with headers in context options', async () => {
```
Lightning00Blade, line 107: This test seems to test nothing. 
We don't set the headers and we don't check that the headers are properly send on the response attached to the page navigation.

Lightning00Blade (commented):
Thank you for your PR, we will discuss if this is something we want to include.
It would be nice to open an issue so other people can find it easier if they want the same feature.

In the mean while I left some comments for the PR itself.

OrKoN:
Closing as stale.

Issue #709, opened 2025-12-24
Add Gemini CLI extension context file for improved tool usage guidance for CLI


LyalinDotCom:
### Is your feature request related to a problem? Please describe.

When using Chrome DevTools MCP with Gemini CLI, the AI doesn't have built-in knowledge of how to effectively use the available tools. Users interacting with Gemini CLI could get mixed results especially if similar MCP tools exists from other installed servers in the context.

### Describe the solution you'd like

Add a gemini-extension.md context file that provides system-level instructions to Gemini CLI when this extension is installed.  We'd also need to update the gemini-extension.json to reference the new file via the contextFileName property.

### Describe alternatives you've considered

This is how Gemini CLI Extensions work best so there is not an alternative to suggest

### Additional context

I'm from the Gemini CLI team and figured it would not hurt to add this feature, so will be submitting a PR as a starting point for this. But expert users of this MCP server might find additional instructions to include in the markdown that I'm not considering.

OrKoN:
cc @natorion

natorion:
This would be great to have - is this the same mechanism as Skills? https://geminicli.com/docs/cli/skills/

art-shen:
A universal skill compatible with different AI agents (OpenCode, Gemini CLI, Claude Code) could be really helpful!

Thebigboss1978:
Antigravity vs Cli Gemini ?

natorion:
> Antigravity vs Cli Gemini ?

?

PR #710, opened 2025-12-24
feat: add Gemini CLI extension context for better tool usage

LyalinDotCom:
Add gemini-extension.md as a system instruction file for Gemini CLI, as suggested in #709. 

**Context**

When users install this MCP server as a Gemini CLI extension extension, Gemini CLI will use this additional context in the markdown file provided to understand how to better use the Chrome DevTools MCP tools.

Updated gemini-extension.json with contextFileName property to reference the new context file.

**Testing**
This was tested with Gemini CLI as its not a direct edit to any of the functionality of the MCP server itself.

### DIFF
--- a/gemini-extension.md
+++ b/gemini-extension.md
@@ -1,0 +1,64 @@
+# Chrome DevTools MCP Extension
+
+This extension enables you to control a real Chrome browser instance to inspect, debug, and automate web interactions via the Model Context Protocol (MCP).
+
+## Core Capabilities & Tools
+
+You have access to a suite of tools categorized by function. Use these tools to perform actions in the browser.
+
+### 1. Navigation & Page Management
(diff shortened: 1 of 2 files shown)


gemini-extension.md
```diff
@@ -0,0 +1,64 @@
+# Chrome DevTools MCP Extension
+
+This extension enables you to control a real Chrome browser instance to inspect, debug, and automate web interactions via the Model Context Protocol (MCP).
+
+## Core Capabilities & Tools
```
OrKoN, line 5: Thanks! I think that the client already has the list of tools and their descriptions from the MCP server. Do you see the duplication of tool definitions as necessary or perhaps the `## Core Capabilities & Tools` section can be removed?

PR #712, opened 2025-12-27
feat(tool-page): support visit pages with header

clChenLiang:
support to visit some page with special headers

### DIFF
--- a/src/tools/pages.ts
+++ b/src/tools/pages.ts
@@ -129,7 +129,7 @@ export const navigatePage = defineTool({
     headers: zod
       .record(zod.string())
       .optional()
-      .describe('Optional headers to send with the navigation request.'),
+      .describe('Optional headers to send with the all request.'),
     ignoreCache: zod
       .boolean()
       .optional()


src/tools/pages.ts
```diff
@@ -88,11 +88,19 @@ export const newPage = defineTool({
   },
   schema: {
     url: zod.string().describe('URL to load in a new page.'),
+    headers: zod
+      .record(zod.string())
+      .optional()
+      .describe('Optional headers to send with the navigation request.'),
```
OrKoN, line 94: the headers would be sent for all requests on the page until the headers are removed so this description is not correct. Could you clarify what use case you have? Would https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/692 work better for the use case? cc @natorion @Lightning00Blade

clChenLiang: we always have an application, which will be test in a prod env. 
this test condition, need to divide different with really prod env which user use.
the best way, is to use a special headers

clChenLiang: the #692 send with header in the whole mcp lives. 
may be in this, can be more  flexible

OrKoN:
Closing as stale. The current implementation in this PR leaks headers beyond individual navigations. The ideal solution would be to either have a separate parameter in emulate tool or remove extra headers after the navigation is done.

PR #715, opened 2025-12-29
feat: Add destructiveHint and openWorldHint tool annotations

bryankthompson:
## Summary

This PR adds the missing MCP tool annotations (`destructiveHint` and `openWorldHint`) to all 26 tools in chrome-devtools-mcp.

### Changes
- **Updated `ToolDefinition.ts`**: Added optional `destructiveHint` and `openWorldHint` fields to the annotations interface
- **Added annotations to all 26 tools**:
  - `openWorldHint: true` for all tools (browser automation inherently interacts with external websites)
  - `destructiveHint: true` for `closePage`, `navigatePage`, and `evaluateScript` (operations that can cause irreversible changes)
  - `destructiveHint: false` for other write operations (reversible modifications like click, fill, hover)

### Why These Annotations Matter

Per the [MCP specification](https://spec.modelcontextprotocol.io/specification/2025-03-26/server/tools/#annotations):

- **`openWorldHint`**: Indicates tools that interact with an "open world" of external entities. All browser automation tools interact with external websites, so this is true for all tools.
- **`destructiveHint`**: Indicates tools that may perform irreversible updates. This helps AI assistants determine when to request user confirmation.

### Tools Summary

| Category | Tools | destructiveHint |
|----------|-------|-----------------|
| Navigation | closePage, navigatePage | true (irreversible) |
| Script | evaluateScript | true (can modify page state) |
| Input | click, hover, fill, drag, fillForm, uploadFile, pressKey, handleDialog | false |
| Debugging | screenshot, takeSnapshot, listConsoleMessages, getConsoleMessage | false |
| Network | listNetworkRequests, getNetworkRequest | N/A (readOnly) |
| Performance | startTrace, stopTrace, analyzeInsight | N/A (readOnly) |
| Other | newPage, selectPage, listPages, resizePage, emulate, waitFor | false or N/A |

### DIFF
--- a/src/tools/emulation.ts
+++ b/src/tools/emulation.ts
@@ -21,7 +21,6 @@ export const emulate = defineTool({
   annotations: {
     category: ToolCategory.EMULATION,
     readOnlyHint: false,
-    destructiveHint: false,
     openWorldHint: true,
   },
   schema: {
--- a/src/tools/input.ts
+++ b/src/tools/input.ts
@@ -18,7 +18,6 @@ export const click = defineTool({
   annotations: {
     category: ToolCategory.INPUT,
     readOnlyHint: false,
-    destructiveHint: false,
     openWorldHint: true,
   },
   schema: {
@@ -59,7 +58,6 @@ export const hover = defineTool({
   annotations: {
     category: ToolCategory.INPUT,
     readOnlyHint: false,
-    destructiveHint: false,
     openWorldHint: true,
   },
   schema: {
@@ -145,7 +143,6 @@ export const fill = defineTool({
   annotations: {
     category: ToolCategory.INPUT,
     readOnlyHint: false,
-    destructiveHint: false,
     openWorldHint: true,
   },
   schema: {
@@ -175,7 +172,6 @@ export const drag = defineTool({
   annotations: {
     category: ToolCategory.INPUT,
     readOnlyHint: false,
-    destructiveHint: false,
     openWorldHint: true,
   },
   schema: {
@@ -206,7 +202,6 @@ export const fillForm = defineTool({
   annotations: {
     category: ToolCategory.INPUT,
     readOnlyHint: false,
-    destructiveHint: false,
     openWorldHint: true,
   },
   schema: {
@@ -240,7 +235,6 @@ export const uploadFile = defineTool({
   annotations: {
     category: ToolCategory.INPUT,
     readOnlyHint: false,
-    destructiveHint: false,
     openWorldHint: true,
   },
   schema: {
@@ -290,7 +284,6 @@ export const pressKey = defineTool({
   annotations: {
     category: ToolCategory.INPUT,
     readOnlyHint: false,
-    destructiveHint: false,
     openWorldHint: true,
   },
   schema: {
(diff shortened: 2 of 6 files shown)


Thebigboss1978:
Thank you, own thank you this is if it was an under human hand I will not succeed any step of my life ha ha ha ha ha ha ha ha thank you

OrKoN:
I am not sure I agree with the destructiveHint classification. `click, hover, fill, drag, fillForm, uploadFile, pressKey` can also trigger navigations and modify the page state. Also, closePage and navigatePage are reversible as they pages can be re-opened and navigated. I think in all cases the destructiveness depends on the payload and cannot be inferred on the tool level. WDYT @sebastianbenz @natorion ?  I think openWorldHint can be added.

sebastianbenz:
+1 to Alex's comments about descructive hint. These steps are also not something where I would want to have a user confirmation.

bryankthompson:
Thanks @OrKoN and @sebastianbenz for the thoughtful review!

You're right - the destructiveness of browser automation tools is entirely payload-dependent. A `click` on a "Delete Account" button is destructive, but a `click` on a "Read More" link is not. This can't be reliably inferred at the tool level.

I've updated the PR to:
1. **Remove all `destructiveHint` annotations** - since destructiveness is context-dependent
2. **Keep `openWorldHint: true`** on all tools - they all interact with external web content
3. **Keep `readOnlyHint: true`** only for pure query tools (`list_pages`, `select_page`, `list_console_messages`, etc.)

This approach accurately reflects that these tools interact with unbounded external web resources without incorrectly triggering confirmation prompts.

sebastianbenz:
Looking at the [spec](https://modelcontextprotocol.io/specification/2025-11-25/schema#toolexecution:~:text=tool%20is%20not.-,default%3A%20true,-ToolExecution), it seems like openWorldHint is already `true` by default. Hence, I don't see much value in explicitly adding it here.

natorion:
> Looking at the [spec](https://modelcontextprotocol.io/specification/2025-11-25/schema#toolexecution:~:text=tool%20is%20not.-,default%3A%20true,-ToolExecution), it seems like openWorldHint is already `true` by default. Hence, I don't see much value in explicitly adding it here.

I agree.

bryankthompson:
Thanks for the feedback @sebastianbenz @natorion - you're absolutely right that `openWorldHint: true` is redundant since that's the [spec default](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.json).

However, I believe the PR still provides concrete value through **`readOnlyHint: true`** on genuinely read-only tools. Unlike `openWorldHint`, `readOnlyHint` defaults to `false`, so explicitly setting it enables MCP clients to auto-approve safe operations without user confirmation.

**Proposed changes:**
- ❌ Remove all `openWorldHint: true` annotations (redundant)
- ✅ Keep `readOnlyHint: true` on query-only tools (`getPageInfo`, `getScreenshot`, `findElement`, `getConsoleMessages`, etc.)
- ✅ Keep `title` annotations for human-readable display

This focused approach addresses your concern about redundancy while still providing value for client implementations.

Would this revised scope be acceptable?

sebastianbenz:
The read-only hints are already in the code base. Closing this PR for now. Thanks for flagging this nevertheless.

PR #724, opened 2026-01-03
fix: use relative path for plugin source in marketplace

liby:
The previous GitHub self-reference format was not recognized by Claude Code when adding this repository as an external marketplace. Using a relative path `./` follows the pattern used by other official plugins like [Sentry](https://github.com/getsentry/sentry-for-claude/blob/main/.claude-plugin/marketplace.json) and ensures proper plugin discovery.

image


liby:
The failing test is a known flaky test, unrelated to this change. Could you please re-run CI? Thanks!

PR #740, opened 2026-01-08
feat: include stack trace in 'get_console_message' tool

szuend:
This is the stack trace of the console message itself. If the argument is an Error object or an "Error.stack" like string we don't do anything special (yet).

The stack trace is source mapped if source maps are available.

The function matches the target ID of the console message to the `SDK.Target` of DevTools. The only oddity is that DevTools works in a lazy manor: It creates stack traces upfront and as source maps come in, stack traces get updated. This does not work for the MCP server. Instead, we wait for script parsed events and source maps to attach before creating the stack trace. Since this could take potentially a while, we guard it with a time out.

### DIFF
--- a/src/DevtoolsUtils.ts
+++ b/src/DevtoolsUtils.ts
@@ -9,7 +9,9 @@ import {Mutex} from './Mutex.js';
 import {DevTools} from './third_party/index.js';
 import type {
   Browser,
+  ConsoleMessage,
   Page,
+  Protocol,
   Target as PuppeteerTarget,
 } from './third_party/index.js';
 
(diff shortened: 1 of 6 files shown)


src/DevtoolsUtils.ts
```diff
@@ -270,3 +272,77 @@ const SKIP_ALL_PAUSES = {
...
+  // loads before creating the stack trace. We might also have to wait for Debugger.ScriptParsed events if
+  // the stack trace is created particularly early.
+  const scriptIds = new Set<Protocol.Runtime.ScriptId>();
+  rawStackTrace.callFrames.forEach(frame => scriptIds.add(frame.scriptId));
+  for (
+    let asyncStack = rawStackTrace.parent;
+    asyncStack;
+    asyncStack = asyncStack.parent
+  ) {
+    asyncStack.callFrames.forEach(frame => scriptIds.add(frame.scriptId));
```
OrKoN: nit: we prefer for/of

src/DevtoolsUtils.ts
```diff
@@ -270,3 +272,77 @@ const SKIP_ALL_PAUSES = {
...
+  scriptId: Protocol.Runtime.ScriptId,
+) {
+  while (true) {
+    const script = model.scriptForId(scriptId);
+    if (script) {
+      return script;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    await model.once('ParsedScriptSource' as any);
```
OrKoN: can we type cast it to the type we expect instead of any.

src/DevtoolsUtils.ts
```diff
@@ -270,3 +272,77 @@ const SKIP_ALL_PAUSES = {
...
+    rawStackTrace as Parameters<
+      DevTools.DebuggerWorkspaceBinding['createStackTraceFromProtocolRuntime']
+    >[0],
+    target,
+  );
+}
+
+// Waits indefinitely for the script so pair it with Promise.race.
+async function waitForScript(
+  model: DevTools.DebuggerModel,
```
OrKoN, line 294: let's give it an abort signal to avoid spinning it infinitely

src/DevtoolsUtils.ts
szuend: Done

src/DevtoolsUtils.ts
szuend: We don't have the type exported but used `Parameters` to work around that.

src/DevtoolsUtils.ts
szuend: Done

Issue #689, opened 2025-12-17
Option to return structured content


OrKoN:
### Is your feature request related to a problem? Please describe.

Currently, the MCP server only returns text/markdown responses. To enable interoperability with out tools, we should have an option to enable structured responses.

### Describe the solution you'd like

We could leverage structured content in the MCP server to provide JSON data alongside the text representation https://modelcontextprotocol.io/specification/2025-11-25/server/tools#structured-content

### Describe alternatives you've considered

- returning JSON in the text content chunks of the MCP
- providing a library interface to get JSON data

PR #744, opened 2026-01-08
chore: support structured content

OrKoN:
The idea is to turn formatters into instances and support both text and JSON formatting.  The structured content is output if the experimental structured content flag is passed. For now, only the snapshots are returned in a structured way. 

Refs https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/689

### DIFF
--- a/src/formatters/SnapshotFormatter.ts
+++ b/src/formatters/SnapshotFormatter.ts
@@ -6,20 +6,21 @@
 import type {TextSnapshot, TextSnapshotNode} from '../McpContext.js';
 
 export class SnapshotFormatter {
-  constructor(public snapshot?: TextSnapshot) {}
+  #snapshot: TextSnapshot;
+
+  constructor(snapshot: TextSnapshot) {
+    this.#snapshot = snapshot;
+  }
 
   toString(): string {
-    if (!this.snapshot) {
-      return '';
-    }
     const chunks: string[] = [];
-    const root = this.snapshot.root;
+    const root = this.#snapshot.root;
 
     // Top-level content of the snapshot.
     if (
-      this.snapshot.verbose &&
-      this.snapshot.hasSelectedElement &&
-      !this.snapshot.selectedElementUid
+      this.#snapshot.verbose &&
+      this.#snapshot.hasSelectedElement &&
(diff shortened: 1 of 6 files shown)


OrKoN:
Screenshot 2026-01-12 at 14 31 44

src/formatters/SnapshotFormatter.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    chunks.push(this.#formatNode(root, 0));
+    return chunks.join('');
+  }
+
+  toJSON(): object {
+    return this.#nodeToJSON(this.#snapshot.root);
+  }
+
+  #formatNode(node: TextSnapshotNode, depth: number): string {
```
Lightning00Blade: Super nit, depth here makes sense to be optional = 0, as its the start, else I thought it would be the depth to stop at.

src/formatters/SnapshotFormatter.ts
```diff
@@ -0,0 +1,162 @@
...
+    for (const child of node.children) {
+      chunks.push(this.#formatNode(child, depth + 1));
+    }
+    return chunks.join('');
+  }
+
+  #nodeToJSON(node: TextSnapshotNode): object {
+    const rawAttrs = this.#getAttributesMap(node);
+    const children = node.children.map(child => this.#nodeToJSON(child));
+    const result: Record<string, unknown> = {
```
Lightning00Blade: I think we can safely use structuredClone.

src/formatters/SnapshotFormatter.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  #getAttributesMap(
+    node: TextSnapshotNode,
+    excludeSpecial = false,
+  ): Record<string, unknown> {
+    const result: Record<string, unknown> = {};
+    if (!excludeSpecial) {
+      result.id = node.id;
+      if (node.role) result.role = node.role;
```
Lightning00Blade: Style here seems wrong.

OrKoN: the linter allows 🤷‍♂️

Issue #465, opened 2025-10-24
Support resize_page when browser window is maximized/fullscreen


ZYK1236:
### Is your feature request related to a problem? Please describe.

**Problem:**
Currently, `resize_page` fails when the browser window is maximized or in full-screen mode with the error:
```
Protocol error (Browser.setContentsSize): Restore window to normal state before setting content size
```

**Expected Behavior:**
`resize_page` should work regardless of the browser window state, similar to how Chrome DevTools "Toggle device toolbar" feature allows dimension changes even when the browser is maximized.

### Describe the solution you'd like

Could the tool automatically handle window state restoration before resizing, or provide an option to simulate viewport resizing (like DevTools device emulation) instead of actual window resizing?

### Describe alternatives you've considered

provide an option to simulate viewport resizing (like DevTools device emulation) instead of actual window resizing

OrKoN:
Thanks, yeah, I think we should automatically unmaximize the window to resize it. The emulation for devices is tracked in separate feature requests.

PeterDraex:
Is there any workaround available?

OrKoN:
Fixed by https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/748

PR #748, opened 2026-01-09
fix: support resize_page when browser window is maximized/fullscreenwindow state

0xlakshan:
Related Issue - [465](https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/465)

### Description

This change make sure that the browser window is restored to a `normal` state before resizing the page. When the window is in `fullscreen` mode we have to set the state twice to match how Chrome CDP behaves. 

Tests cover window resizing for every available window states `type WindowState = 'normal' | 'minimized' | 'maximized' | 'fullscreen';`

### DIFF
--- a/tests/tools/pages.test.ts
+++ b/tests/tools/pages.test.ts
@@ -228,6 +228,9 @@ describe('pages', () => {
           context,
         );
         await resizePromise;
+        await page.waitForFunction(
+          () => window.innerWidth === 700 && window.innerHeight === 500,
+        );
         const dimensions = await page.evaluate(() => {
           return [window.innerWidth, window.innerHeight];
         });
@@ -256,6 +259,9 @@ describe('pages', () => {
           context,
         );
         await resizePromise;
+        await page.waitForFunction(
+          () => window.innerWidth === 650 && window.innerHeight === 450,
+        );
         const dimensions = await page.evaluate(() => {
           return [window.innerWidth, window.innerHeight];
         });
@@ -284,6 +290,9 @@ describe('pages', () => {
           context,
         );
         await resizePromise;
+        await page.waitForFunction(
+          () => window.innerWidth === 750 && window.innerHeight === 550,
+        );
         const dimensions = await page.evaluate(() => {
           return [window.innerWidth, window.innerHeight];
         });
@@ -312,6 +321,9 @@ describe('pages', () => {
           context,
         );
         await resizePromise;
+        await page.waitForFunction(
+          () => window.innerWidth === 725 && window.innerHeight === 525,
+        );
         const dimensions = await page.evaluate(() => {
           return [window.innerWidth, window.innerHeight];
         });
@@ -340,6 +352,9 @@ describe('pages', () => {
           context,
         );
         await resizePromise;
+        await page.waitForFunction(
+          () => window.innerWidth === 850 && window.innerHeight === 650,
+        );
         const dimensions = await page.evaluate(() => {
           return [window.innerWidth, window.innerHeight];
         });


src/tools/pages.ts
```diff
@@ -211,6 +211,18 @@ export const resizePage = defineTool({
...
 
+    const browser = page.browser();
+    const windowId = await page.windowId();
+
+    const bounds = await browser.getWindowBounds(windowId);
+
+    if (bounds.windowState === 'fullscreen') {
+      // have to call this twice when window is in fullscreen mode
+      await browser.setWindowBounds(windowId, {windowState: 'normal'});
+      await browser.setWindowBounds(windowId, {windowState: 'normal'});
```
OrKoN: is doing it two times really required? At least testing on Mac, one time is sufficient. Which OS have you tested with?

0xlakshan: I’m on `Ubuntu 24.04.3 LTS (x86_64)` and testing with the MCP Inspector. When the window is in `fullscreen` mode and I try to resize it, the inspector throws the following error, so I have to call it twice

`Protocol error (Browser.setContentsSize): Restore window to normal state before setting content size`

OrKoN: Right, but do we need to restore to normal twice? is there a race condition perhaps, i.e., the first `browser.setWindowBounds(windowId, {windowState: 'normal'});` is not actually finished by the time the call is resolved? 

Let me test on a Linux.

src/tools/pages.ts
```diff
@@ -211,6 +211,18 @@ export const resizePage = defineTool({
   handler: async (request, response, context) => {
     const page = context.getSelectedPage();
 
+    const browser = page.browser();
+    const windowId = await page.windowId();
```
OrKoN: let's wrap the window management operations in a try/catch and suppress the error to avoid breaking the tool in case there are some cross-platform differences operating with windows. IIRC the android browsers might not have the concept of windows.

0xlakshan: done!

src/tools/pages.ts
OrKoN: Alright, I could not really test due to shortcut difficulties with remote desktop + gnome.

src/tools/pages.ts
```diff
@@ -211,6 +211,22 @@ export const resizePage = defineTool({
...
     const page = context.getSelectedPage();
 
+    try {
+      const browser = page.browser();
+      const windowId = await page.windowId();
+
+      const bounds = await browser.getWindowBounds(windowId);
+
+      if (bounds.windowState === 'fullscreen') {
+        // have to call this twice when window is in fullscreen mode
```
```suggestion
        // Have to call this twice on Ubuntu when the window is in fullscreen mode.
```

tests/tools/pages.test.ts
```diff
@@ -233,6 +233,118 @@ describe('pages', () => {
...
+        );
+        await resizePromise;
+        const dimensions = await page.evaluate(() => {
+          return [window.innerWidth, window.innerHeight];
+        });
+        assert.deepStrictEqual(dimensions, [750, 550]);
+      });
+    });
+
+    it('resize when window state is maximized', async () => {
```
OrKoN, line 303: looks like some of these tests could be flaky https://github.com/ChromeDevTools/chrome-devtools-mcp/actions/runs/20917900887/job/60095926344?pr=748

0xlakshan: Let me do a bit of research on this

0xlakshan: I was able to reproduce this issue. As shown in the CI/CD pipeline, the same test also failed locally. I tried adjusting the dimensions by both increasing and decreasing them. 

After increasing the test dimensions beyond `800×600`, the tests always failed. 

When I lowered the dimensions, the tests consistently passed. But, I was not able to findout the exact dimension threshold at which the test starts failing

OrKoN:
I guess might require setting a larger screen info for the test browser similar to https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/src/browser.ts#L178

0xlakshan:
After testing few times i noticed that the `maximized` and `fullscreen` tests often fail, with dimensions overlapping those of the previous tests. 

I think this happens because Chrome is still resizing from the previous test when the next test starts, and the resize listener resolves on the previous resize event rather than the current one. 

To fix this, I found that the puppeteer's `Page.waitForFunction()` method can be used to wait until the size dimensions are actually being set

OrKoN:
Thanks!

PR #753, opened 2026-01-12
feat: add --profile-directory option to specify Chrome profile

### LINKED ISSUE #694: Support multiple Chrome Profiles

When multiple profiles are open at the same time, there's no way to control which profile will be automated by the MCP server. The MCP server will select the default profile as determined by Chrome. 

A way to explicitly select a profile when running the MCP server.

`--profile-directory` can be used on Chrome to specify the default profile (https://superuser.com/questions/377186/how-do-i-start-chrome-using-a-specified-user-profile), this might be a good option (which we should document if it works).

### Additional context

_No response_

Yasindu20:
## Summary
Adds support for specifying which Chrome profile to use via a new `--profile-directory` CLI option.

## Motivation
Users often need to:
- Test with different browser profiles that have specific extensions installed
- Maintain separate contexts for different projects
- Use profiles with pre-configured settings or logged-in accounts

Currently, there's no way to specify which profile Chrome should use when launched by the MCP server.

## Changes
- ✅ Added `--profile-directory` CLI option with kebab-case alias
- ✅ Updated `BrowserOptions` interface to include `profileDirectory` property
- ✅ Connected CLI argument to Chrome's `--profile-directory` launch flag
- ✅ Added documentation in README.md with usage examples
- ✅ Included in the auto-generated options documentation

## Usage Example
```json
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": [
        "chrome-devtools-mcp@latest",
        "--profile-directory=Profile 1"
      ]
    }
  }
}
```

Or via command line:
```bash
npx chrome-devtools-mcp --profile-directory="Profile 1"
```

- [x] Documentation generated correctly with `npm run docs`
- [x] format with  `npm run format`

## Related Issues
Fixes #694

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -63,6 +63,53 @@ async function getBrowserContextForProfile(
     const contexts = browser.browserContexts();
     logger(`Found ${contexts.length} browser context(s)`);
 
+    for (const context of contexts) {
+      let page;
+      try {
+        page = await context.newPage();
+
+        await page.goto('chrome://version', {
+          waitUntil: 'domcontentloaded',
+          timeout: 3_000,
+        });
+
+        const profilePath: string | null = await page.evaluate(() => {
+          const body = document.querySelector('body');
+          if (!body) return null;
+          const text = (body.innerText || '');
+          const match = text.match(/Profile Path:\s*(.+)/i);
+          return match ? match[1].trim() : null;
+        });
+
+        try {
+          await page.close();
+        } catch {
+          //ignore close errors
+        }
+
+        if (!profilePath) {
+          continue;
+        }
+
+        const actualProfile = getProfileNameFromUserDataDir(profilePath);
+        logger(`Probed context: profilePath=${profilePath} => profileName=${actualProfile}`);
+
+        if (actualProfile === profileDirectory) {
+          logger(`Matched profile directory "${profileDirectory}" to a browser context`);
+          return context;
+        }
+      } catch (error) {
+        logger('Error probing a browser context for profile: ', error);
+        try {
+          if (page && !page.isClosed()) {
+            await page.close();
+          }
+        } catch {
+          // ignore
+        }
+      }
+    }
+
     logger(
       `Profile directory "${profileDirectory}" specified. ` +
         `Using default browser context. Full profile support will be added in a future update.`,


src/cli.ts
```diff
@@ -183,6 +183,13 @@ export const cliOptions = {
     default: true,
     describe: 'Set to false to exclude tools related to network.',
   },
+  profileDirectory: {
+    type: 'string',
+    description:
+      'Specify which Chrome profile to use by its directory name (e.g., "Profile 1", "Default"). Only works with --autoConnect or when launching Chrome.',
```
sebastianbenz:
```suggestion
      'Specify which Chrome profile to use by specifying its directory name (e.g., "Profile 1", "Default") inside a chrome user data directory. Only works with --autoConnect or when launching Chrome via the Chrome DevTools MCP server.',
```

Tried to make it a bit more clear //cc @OrKoN

sebastianbenz:
Did you test this with autoconnect? I just tested this by running Chrome in "Profile 1" and setting `"--profile-directory=Profile 2"`. In this case, the MCP server still connected to `Profile 1`, but I would've expected the connection to fail. Or is this the intended behavior?

Yasindu20:
> Did you test this with autoconnect? I just tested this by running Chrome in "Profile 1" and setting `"--profile-directory=Profile 2"`. In this case, the MCP server still connected to `Profile 1`, but I would've expected the connection to fail. Or is this the intended behavior?

@sebastianbenz Yes, I tested this with autoconnect and saw the same behavior even when Chrome is running with “Profile 1”, the MCP server still connects successfully when --profile-directory=Profile 2 is specified. So, I add a validation step to detect a profile mismatch after connecting and fail fast instead of silently attaching to the wrong profile. The intention is to avoid surprising connections when a specific profile is explicitly requested. Let me know if you’d like me to add anything specific to the docs around the profile behavior.

tacsizkral546154:
- Yasindu, Chrome profilini belirtmek için yeni --profile-directory
seçeneğini
   ekleyen bir çekme isteği başlattı.
   - Sebastian'ın otomatik bağlantı sırasındaki profil uyuşmazlığı
   endişeleri üzerine, Yasindu bir doğrulama adımı ekledi.



13 Oca 2026 Sal 09:42 tarihinde Yasindu Dasanga De Mel <
***@***.***> şunu yazdı:

src/browser.ts
```diff
@@ -126,6 +134,39 @@ export async function ensureBrowserConnected(options: {
       },
     );
   }
+
+  if (options.profileDirectory && options.userDataDir) {
+    try {
+      const portPath = path.join(options.userDataDir, 'DevToolsActivatePort');
```
OrKoN: such file does not exist, did you mean DevToolsPort. A single user data dir hosts multiple profiles so I am not sure if the check handles it correctly.

src/browser.ts
```diff
@@ -126,6 +134,39 @@ export async function ensureBrowserConnected(options: {
       },
     );
   }
+
+  if (options.profileDirectory && options.userDataDir) {
```
OrKoN: I do not think this check should be in the connection code branch.

src/cli.ts
```diff
@@ -198,6 +198,13 @@ export const cliOptions = {
     default: true,
     describe: 'Set to false to exclude tools related to network.',
   },
+  profileDirectory: {
```
OrKoN, line 201: we probably should be using https://pptr.dev/api/puppeteer.browser.browsercontexts to locate the browser context for the profile (currently the default browser context is used)

Yasindu20:
@OrKoN I fixed the DevToolsActivePort from DevToolsActivatePort. Removed the profile validation logic since a single userDataDir can contain multiple profiles, and that check isn’t reliable. Also removed the profile-related checks from the connection path, add it to the function launch(). For profile handling, I switched to referencing browser.browserContexts() and added clear logging that we currently fall back to the default context. Proper profile → context mapping can be improved once Puppeteer exposes a cleaner way to do this.

src/browser.ts
```diff
@@ -43,13 +44,45 @@ function makeTargetFilter() {
...
+  if (!profileDirectory) {
+    return browser.defaultBrowserContext();
+  }
+
+  try {
+    const contexts = browser.browserContexts();
+    logger(`Found ${contexts.length} browser context(s)`);
+
+    logger(
+      `Profile directory "${profileDirectory}" specified. ` +
```
OrKoN, line 114: I think this PR does not add anything useful without the implementation for finding the browser context that matches the profileDirectory name. Passing `--profile-directory` to Chrome is already possible via the `--chrome-arg` param. Would you be up to figuring out how to find the correct browser context? it might right changes on the CDP and Puppeteer side.

Yasindu20:
@OrKoN Puppeteer or CDP don’t currently expose profile directory info for browser contexts. Now I tried to implement it to open a temporary page in each context, reads chrome://version to get the profile path, and matches it to the requested --profile-directory. If no match is found, it falls back to the default context. This works with existing APIs and can be updated later if CDP/Puppeteer expose profile metadata. If need any changes let me know

OrKoN:
> @OrKoN Puppeteer or CDP don’t currently expose profile directory info for browser contexts. Now I tried to implement it to open a temporary page in each context, reads chrome://version to get the profile path, and matches it to the requested --profile-directory. If no match is found, it falls back to the default context. This works with existing APIs and can be updated later if CDP/Puppeteer expose profile metadata. If need any changes let me know

Thanks for explaining. I think it does not add much value and might give the wrong impression that `--profile-directory` would actually force a different profile to be used. The PR also went stale. I think the ideal solution is that we can rely on Chrome and CDP to properly deal with multiple profiles via https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/694.

PR #757, opened 2026-01-12
chore: add usage statistics CLI flag

ergunsh:
This PR initiates the implementation of clearcut telemetry. It introduces the CLI configuration required to opt-out of usage statistics, along with the necessary transparency disclaimers for the user. It's hidden by default during development & while we get the launch approvals.

**Implementation Roadmap:**
This is the first in a series of PRs designed to implement the telemetry system:
1.  **CLI & Opt-out Mechanism (This PR):**
    *   Adds the `--usage-statistics` flag (hidden, default `false` for now).
    *   Adds transparency logging to specify data collection when enabled.
2.  **Logger Scaffolding & Integration:**
    *   Implement `ClearcutLogger` and integrate it into `main.ts`.
    *   Add hooks for `logServerStart` and `logToolInvocation`.
    *   Introduce `ClearcutSender` as an abstraction layer for transport logic.
3.  **Persistence Layer:**
    *   Implement local state management to reliably track "First Time Installation" and "Daily Active" metrics.
    *   Implement sending the daily active and first time installation event based on this local state.
4.  **Watchdog Process Architecture:**
    *   Move `ClearcutSender` execution to a dedicated watchdog process to ensure reliable event transmission even during abrupt server shutdowns.
5.  **Transport, Batching & Retries:**
    *   Finalize `ClearcutSender` with actual HTTP transport logic, including event batching, respecting `next_request_wait_millis` and retries.

PR #758, opened 2026-01-12
chore: implement ClearcutLogger and ClearcutSender dummy

ergunsh:
This PR implements the core logging infrastructure for the telemetry system. It introduces the `ClearcutLogger` class, integrates it into the server lifecycle events, and establishes the `ClearcutSender` abstraction.

**Implementation Roadmap:**
This is the second in a series of PRs designed to implement a robust, privacy-conscious telemetry system:

2.  **Logger Scaffolding & Integration (This PR):**
    *   **`ClearcutSender`**: Introduced a transport abstraction (currently a dummy implementation) that will later handle HTTP requests, batching, and retries.
3.  **Persistence Layer (Next):**
    *   Implement local state management to reliably track "First Time Installation" and "Daily Active" metrics.
4.  **Watchdog Process Architecture (Next):**
    *   Move `ClearcutSender` execution to a dedicated watchdog process to ensure reliable event transmission even during abrupt server shutdowns.


src/telemetry/clearcut-logger.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
OrKoN: 2026

src/telemetry/clearcut-sender.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
OrKoN: 2026

src/telemetry/types.ts
```diff
@@ -0,0 +1,77 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
OrKoN: 2026

src/main.ts
```diff
@@ -207,4 +220,11 @@ await loadIssueDescriptions();
 const transport = new StdioServerTransport();
 await server.connect(transport);
 logger('Chrome DevTools MCP Server connected');
+void clearcutLogger?.logServerStart({
+  browser_url_present: !!args.browserUrl,
```
OrKoN: can we start with all args?

tests/telemetry/clearcut-logger.test.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
OrKoN: 2026

src/main.ts
ergunsh: I'd prefer to go like this for now and enhance the logging once we have the implementation in place. We need to update the logging proto on the server side for these to take effect, and would like to have a process there where we first update the proto, and then update the instrumentation here. Does it sound good, or should I go ahead and do that directly?

OrKoN: if we do not update protos, the data will be dropped right? can we have the generic code on the client to pass args in that case?

OrKoN: as an alternative, can we implement this in a generic way, iterating over a list (current args) of arg names and converting them based on the type?

ergunsh: I'm looking into it!

The logic would work as follows:
* If it's boolean, convert the arg name to snake case and log it directly.
* If it's enum (e.g. an argument with multiple choices, in our case), convert the arg name to snake case,  and map the choice correctly to the enum value numbers we specified in the proto.
* Otherwise, convert the arg name to snake case with `_present` suffix, and log it if it's present.

For the types, we'd need to define it as:
```
interface FlagUsage {
    [key: string]: boolean | number | undefined; 
}
```

The challenge in here is to make sure that `choice` is mapped to the correct enum value defined on the server side. For this, if we assume that the `choices` in here is stable, we can directly log the choice index and update the enum definition on the server side to match the indexes defined on the client side (`index + 1` specifically, and we'd keep `0` for `unspecified`).

WDYT?

OrKoN: IIRC the servers can convert enum text value to the id automatically?

ergunsh: Ah correct! As discussed and confirmed offline, we're able to send the key from an enum and it's automatically resolved to the enum value. Updated the CLI arg logging to be generic and started sending the uppercase enum value for the enums.

On the server side, we're going to have enums matching the uppercase choices for enumerated flags.

PR #759, opened 2026-01-13
build: update verify server.json

OrKoN:
move download to the script, run script without deleting server.json

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -60,6 +60,7 @@ export default defineConfig([
     name: 'TypeScript rules',
     rules: {
       '@local/check-license': 'error',
+      curly: ['error', 'all'],
 
       'no-undef': 'off',
       'no-unused-vars': 'off',
--- a/scripts/verify-server-json-version.ts
+++ b/scripts/verify-server-json-version.ts
@@ -6,31 +6,72 @@
 
 import {execSync} from 'node:child_process';
 import fs from 'node:fs';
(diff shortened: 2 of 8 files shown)


scripts/verify-server-json-version.ts
```diff
@@ -6,31 +6,72 @@
 
 import {execSync} from 'node:child_process';
 import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
 
-const serverJsonFilePath = './server.json';
+const serverJsonFilePath = path.join(process.cwd(), 'server.json');
```
Lightning00Blade, line 12: This should use at least the dirname from the file here. Else it can fail if you execute from the wrong folder.

eslint.config.mjs
```diff
@@ -60,6 +60,7 @@ export default defineConfig([
     name: 'TypeScript rules',
     rules: {
       '@local/check-license': 'error',
+      curly: ['error', 'all'],
```
Lightning00Blade, line 63: Next time please create two PRs.

Lightning00Blade (approved):

scripts/verify-server-json-version.ts
OrKoN: The existing script did not use it, so leaving it to the future work.

eslint.config.mjs
OrKoN: Will do

PR #762, opened 2026-01-13
test: add a retry for the merge queue

OrKoN:
The tests need to pass checks before being added to the merge queue so we can avoid merge queue failures in case there are flaky tests.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -59,7 +59,7 @@ jobs:
 
       - name: Run tests
         shell: bash
-        # Try tests if they fail in the merge queue.
+        # Retry tests if they fail in the merge queue.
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -59,7 +59,8 @@ jobs:
 
       - name: Run tests
         shell: bash
-        run: npm run test:no-build
+        # Try tests if they fail in the merge queue.
```
OrKoN:
```suggestion
        # Retry tests if they fail in the merge queue.
```

scripts/test.mjs
```diff
@@ -17,6 +17,13 @@ const flags = args.filter(arg => arg.startsWith('-'));
 
 const files = [];
 
+let shouldRetry = false;
```
OrKoN, line 20: will attempt replacing with util.parseArgs([config]) from node (from node 20+) in the next PR

PR #764, opened 2026-01-13
chore: Implement install_extension command

nattallius:


src/tools/categories.ts
```diff
@@ -11,6 +11,7 @@ export enum ToolCategory {
   PERFORMANCE = 'performance',
   NETWORK = 'network',
   DEBUGGING = 'debugging',
+  EXTENSION = 'extension',
```
OrKoN: let's call the category `EXTENSIONS = 'extensions'

src/tools/extensions.ts
```diff
@@ -0,0 +1,35 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { logger } from '../logger.js';
+import { zod } from '../third_party/index.js';
+
+import { ToolCategory } from './categories.js';
+import { defineTool } from './ToolDefinition.js';
+
+export const installExtension = defineTool({
```
OrKoN, line 12: let's add a test for the tool. We can use puppeteer that an extension appears on the chrome://extensions page

src/main.ts
```diff
@@ -64,34 +58,37 @@ async function getContext(): Promise<McpContext> {
...
       ? await ensureBrowserConnected({
-          browserURL: args.browserUrl,
-          wsEndpoint: args.wsEndpoint,
-          wsHeaders: args.wsHeaders,
-          // Important: only pass channel, if autoConnect is true.
-          channel: args.autoConnect ? (args.channel as Channel) : undefined,
-          userDataDir: args.userDataDir,
-          devtools,
-        })
+        browserURL: args.browserUrl,
```
OrKoN: this looks not indented correctly, please fix up the PR to pass the linter and regenerate docs.

src/cli.ts
```diff
@@ -198,6 +198,12 @@ export const cliOptions = {
     default: true,
     describe: 'Set to false to exclude tools related to network.',
   },
+  categoryExtension: {
+    type: 'boolean',
+    default: true,
```
OrKoN: should not the default be false until we are ready to launch all tools?

src/main.ts
```diff
@@ -64,34 +58,37 @@ async function getContext(): Promise<McpContext> {
   const ignoreDefaultChromeArgs: string[] = (
     args.ignoreDefaultChromeArg ?? []
   ).map(String);
+  if (args.categoryExtension) {
+    chromeArgs.push('--enable-unsafe-extension-debugging');
```
OrKoN: please use enableExtensions on the puppeteer launch interface

docs/tool-reference.md
```diff
@@ -345,3 +347,15 @@ in the DevTools Elements panel (if any).
 - **verbose** (boolean) _(optional)_: Whether to include all possible information available in the full a11y tree. Default is false.
 
 ---
+
+## Extensions
```
OrKoN: let's filter out the extensions category from the doc generation before we have the complete set of tools.

src/tools/extensions.ts
```diff
@@ -0,0 +1,35 @@
...
+    readOnlyHint: false,
+  },
+  schema: {
+    path: zod
+      .string()
+      .describe('Absolute path to the unpacked extension folder.'),
+  },
+  handler: async (request, response, context) => {
+    const {path} = request.params;
+    try {
```
OrKoN: nit: we do not need a try/catch and logging for the error because we already have try/catch for the tool handling. Additional logs here do not add any new info for debugging (see main.ts).

tests/tools/extensions.test.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
OrKoN:
```suggestion
 * Copyright 2026 Google LLC
```

tests/tools/extensions.test.ts
```diff
@@ -0,0 +1,68 @@
...
+import path from 'node:path';
+import {describe, it, after} from 'node:test';
+
+import {installExtension} from '../../src/tools/extensions.js';
+import {withMcpContext} from '../utils.js';
+
+describe('extension', () => {
+  let tmpDir: string;
+
+  it('installs an extension and verifies it is listed in chrome://extensions', async () => {
```
OrKoN, line 20: let's create a fixtures folder under tests (similar to tests/trace-processing/fixtures) and put the extension test files there so that it is easier to re-use for other tests and manual testing.

OrKoN (commented):
Thanks! Left a few comments!

src/browser.ts
```diff
@@ -201,6 +201,7 @@ export async function launch(options: McpLaunchOptions): Promise<Browser> {
       ignoreDefaultArgs: ignoreDefaultArgs,
       acceptInsecureCerts: options.acceptInsecureCerts,
       handleDevToolsAsPage: true,
+      enableExtensions: true,
```
OrKoN: we should only enable this if the extensions tool category is enabled, right?

tests/tools/fixtures/extension/popup.html
```diff
@@ -0,0 +1,6 @@
+<!doctype html>
+<html
+  ><body
```
OrKoN: Perhaps we should move `>` to the previous line? also in other instances

scripts/generate-docs.ts
```diff
@@ -278,9 +278,15 @@ async function generateToolDocumentation(): Promise<void> {
     // Convert ToolDefinitions to ToolWithAnnotations
     const toolsWithAnnotations: ToolWithAnnotations[] = tools
       .filter(tool => {
+        // Filter out extension tools
```
OrKoN, line 281: let's follow up in the next PRs by adding a condition label to the tool definition instead of adding the filter by tool name here

PR #769, opened 2026-01-14
chore: implement telemetry persistence and daily active logging

ergunsh:
This PR implements the persistence layer for the telemetry system. It introduces `FilePersistence` for local state management and integrates it with `ClearcutLogger` to support "Daily Active"  metric. I have decided not to send `first_time_installation` events since we can deduce them from `daily active` events where the `days_since_last_active` will be `-1` for that case.

**Implementation Roadmap:**
This is the third in a series of PRs designed to implement the telemetry system:
2.  **Logger Scaffolding & Integration ([Merged](https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/758)):**
    *   **`ClearcutSender`**: Introduced a transport abstraction.
3.  **Persistence Layer (This PR):**
    *   **`FilePersistence`**: Implemented a local file-based state manager to persist the `lastActive` timestamp.
    *   **Daily Active Logic**: Integrated persistence into `ClearcutLogger` to automatically detect and log `daily_active` events (with `days_since_last_active` calculation) via `logDailyActiveIfNeeded`.
4.  **Watchdog Process Architecture (Next):**
    *   Move `ClearcutSender` execution to a dedicated watchdog process to ensure reliable event transmission even during abrupt server shutdowns.

### DIFF
--- a/src/telemetry/persistence.ts
+++ b/src/telemetry/persistence.ts
@@ -62,13 +62,13 @@ export class FilePersistence implements Persistence {
   }
 
   async saveState(state: LocalState): Promise<void> {
+    const filePath = path.join(this.#dataFolder, STATE_FILE_NAME);
     try {
       await fs.mkdir(this.#dataFolder, {recursive: true});
-      const filePath = path.join(this.#dataFolder, STATE_FILE_NAME);
       await fs.writeFile(filePath, JSON.stringify(state, null, 2), 'utf-8');
     } catch (error) {
       // Ignore errors during state saving to avoid crashing the server
-      logger('Failed to save telemetry state:', error);
+      logger(`Failed to save telemetry state to ${filePath}:`, error);
     }
   }
 }


src/telemetry/clearcut-logger.ts
```diff
@@ -4,14 +4,22 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import {logger} from '../logger.js';
```
OrKoN, line 7: Please use `ClearcutLogger.ts` for the file name. Overlooked previously.

src/telemetry/persistence.ts
```diff
@@ -0,0 +1,74 @@
...
+    name,
+  );
+}
+
+export interface Persistence {
+  loadState(): Promise<LocalState>;
+  saveState(state: LocalState): Promise<void>;
+}
+
+export class FilePersistence implements Persistence {
```
OrKoN, line 45: how do it work with multiple MCP server instances?

ergunsh: For this initial implementation, I was thinking that this race condition can be an acceptable trade-off considering:
 
1. The critical section (reading state → writing update) is quiet fast and effectively synchronous (we do not need to await the HTTP network call before writing), so the probability of a collision is technically low.

2. This manages telemetry data, not critical application state. If a collision does occur, the worst case is a slight inaccuracy in metrics (e.g., a double count) or a file reset via the error handling block, neither of which impacts the user's actual workflow or server stability.

Though for a proper solution, we can use a 3rd party library [proper-lockfile](https://www.npmjs.com/package/proper-lockfile) or implement a mechanism ourselves directly.

One approach to create a locking mechanism is using `fs.open({lock-path}, 'wx')` to create a lock file, and removing the file to release the lock. If another process tries to do the same operation while the file exists, then it'll receive an `EEXIST` error and it can retry acquiring the lock after some time. In addition to the happy paths, we'd need to handle stale locks -- i.e. if the lock file is created more than 5s ago, then the next process should be allowed to steal it.

Let me know what you think!

src/telemetry/clearcut-logger.ts
ergunsh: Ah, there are also other files along the lines `clearcut-sender`, `flag-utils` and their respective tests. WDYT of me handling this in a separate PR right after this?

src/telemetry/persistence.ts
OrKoN: Let's not add any third party dependencies. Sounds good for active daily tracking but let's think about it more if we do more persistence

src/telemetry/clearcut-logger.ts
OrKoN: Ack

src/telemetry/persistence.ts
```diff
@@ -0,0 +1,74 @@
...
+import process from 'node:process';
+
+import {logger} from '../logger.js';
+
+export interface LocalState {
+  lastActive: string; // ISO 8601 UTC date string
+}
+
+const STATE_FILE_NAME = 'telemetry_state.json';
+function getDataFolder(): string {
```
OrKoN, line 19: would this work under WSL?

src/telemetry/persistence.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  async saveState(state: LocalState): Promise<void> {
+    try {
+      await fs.mkdir(this.#dataFolder, {recursive: true});
+      const filePath = path.join(this.#dataFolder, STATE_FILE_NAME);
+      await fs.writeFile(filePath, JSON.stringify(state, null, 2), 'utf-8');
+    } catch (error) {
+      // Ignore errors during state saving to avoid crashing the server
+      logger('Failed to save telemetry state:', error);
```
OrKoN: let's add path into the string part of the log

src/telemetry/persistence.ts
ergunsh: I couldn't test this out in our VMs since WSL is not available there. However, considering that `process.platform` returns `linux` in a WSL environment, this should correctly work by saving data to the `~/.local/share/chrome-devtools-mcp` folder.

Do you have any specific concerns regarding this (and if you have a windows machine with WSL support, can you help me test it? 😊)

src/telemetry/persistence.ts
ergunsh: Done!

src/telemetry/persistence.ts
OrKoN: Not yet, alright sg for now.

PR #770, opened 2026-01-14
chore: support an html route in evals

OrKoN:
Allows using the server URL in the prompt.

### DIFF
--- a/scripts/eval_gemini.ts
+++ b/scripts/eval_gemini.ts
@@ -6,7 +6,7 @@
 
 import fs from 'node:fs';
 import path from 'node:path';
-import {describe, test} from 'node:test';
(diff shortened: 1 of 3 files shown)


scripts/eval_gemini.ts
```diff
@@ -84,10 +89,24 @@ async function runSingleScenario(
 
   let client: Client | undefined;
   let transport: StdioClientTransport | undefined;
+  let server: TestServer | undefined;
 
   try {
     const scenario = await loadScenario(absolutePath);
 
+    if (scenario.htmlRoute) {
+      server = new TestServer(TestServer.randomPort());
```
Lightning00Blade: Let's start the server only once and reset it after each test.

Lightning00Blade (approved):

PR #766, opened 2026-01-14
chore: add basic eval

OrKoN:
This PR adds a tool based on the node test runner that runs a loop to see what tools a model chooses given a prompt. The expectations are encoding for each prompt. Run `npm run eval` to get results. Currently, only Gemini and needs and API key.

PR #777, opened 2026-01-15
chore: Implement Watchdog process for reliable telemetry 

ergunsh:
This PR implements the watchdog process architecture for the telemetry system. It moves the `ClearcutSender` execution to a dedicated child process, ensuring that events—especially shutdown events—are reliably transmitted even if the main server process terminates abruptly.

Added an e2e test that runs the server, checks the log file and confirms the telemetry logs exist and that the watchdog process is correctly killed after sending the shutdown event once the main process is killed.

**Implementation Roadmap:**
This is the fourth in a series of PRs designed to implement the telemetry system:
2.  **Logger Scaffolding & Integration ([Merged](https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/758)):**
    *   **`ClearcutSender`**: Introduced a transport abstraction.
3.  **Persistence Layer ([Merged](https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/766)):**
    *   **`FilePersistence`**: Implemented a local file-based state manager to persist the `lastActive` timestamp.
    *   **Daily Active Logic**: Integrated persistence into `ClearcutLogger` to automatically detect and log `daily_active` events (with `days_since_last_active` calculation) via `logDailyActiveIfNeeded`.
4.  **Watchdog Process Architecture (This PR):**
    *   **`WatchdogClient`**: Added a client-side wrapper to spawn and communicate with the watchdog process via `stdin`.
    *   **`watchdog/main.ts`**: Created the entry point for the watchdog process. It listens for IPC messages and uses `ClearcutSender` to transmit events.
    *   **Reliable Shutdown**: The watchdog monitors the parent process and guarantees a `shutdown` event is sent when the parent exits or crashes (detecting `stdin` closure).
    *   **Refactoring**: Moved `ClearcutSender` to the `watchdog` directory and updated `ClearcutLogger` to delegate event sending to the `WatchdogClient`.

### DIFF
--- a/tests/e2e/telemetry.test.ts
+++ b/tests/e2e/telemetry.test.ts
@@ -5,11 +5,11 @@
  */
 
 import assert from 'node:assert';
-import {execFile, spawn, type ChildProcess} from 'node:child_process';
+import {spawn, type ChildProcess, type SpawnOptions} from 'node:child_process';
 import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
-import {describe, it, type TestContext as NodeTestContext} from 'node:test';
+import {describe, it} from 'node:test';
 
 const SERVER_PATH = path.resolve('build/src/main.js');
 const WATCHDOG_START_PATTERN = /Watchdog started[\s\S]*?"pid":\s*(\d+)/;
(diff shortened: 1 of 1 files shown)


src/logger.ts
```diff
@@ -31,4 +31,26 @@ export function saveLogsToFile(fileName: string): fs.WriteStream {
   return logFile;
 }
 
+export function saveLogsToFileSync(fileName: string): void {
```
OrKoN: why do we need to save logs synchronously?

ergunsh: When the watchdog process is killing itself, it needs to make sure that the logs are saved before shutdown.

OrKoN: Did you actually experience it not writing logs or is it a precaution?

OrKoN: Can we instead export the logging streams and flush/close them?

src/telemetry/watchdog-client.ts
```diff
@@ -0,0 +1,71 @@
...
+    this.#childProcess = spawner(process.execPath, args, {
+      stdio: ['pipe', 'ignore', 'ignore'],
+      detached: true,
+    });
+    this.#childProcess.unref();
+    this.#childProcess.on('error', err => {
+      logger('Watchdog process error:', err);
+    });
+    this.#childProcess.on('exit', (code, signal) => {
+      logger(`Watchdog exited with code ${code} and signal ${signal}`);
```
OrKoN, line 51: we should restart the watchdog if it crashes? WDYT?

src/logger.ts
ergunsh: Ah, I experienced it not writing logs and it was failing the e2e test I added. We can definitely export the logging streams and flush/close them before closing. I'm on it!

(btw let me know what you think about the e2e test as well! Currently, it relies on the logger to log specific lines on the log file provided since there is no other side effect on the environment yet. Once we have the `fetch` calls, we can test this through a mock server to make sure that the calls happened and that all the processes are exited once the main is finished)

OrKoN: I think we should check if the metrics were sent specifically. I think the cleanest way would be to allow configuring a transport for sending telemetry and passing a sync custom transport in tests.

src/telemetry/watchdog-client.ts
ergunsh: We'd need to implement some back-off mechanism in case the crashes are somehow persistent. WDYT of doing this as a follow-up?

OrKoN: ack

src/logger.ts
ergunsh: I have updated this to continue with async logging, and started flushing logs in the watchdog process before exiting. Thanks for the suggestion!

Can you PTAL again?

For the tests specifically, I find e2e test to be useful for making sure that the telemetry start & shutdown flow works end to end. It confirms:
* Main process starts the watchdog process correctly.
* Main process is able to pass messages to the watchdog process with the `server_start` event.
* Watchdog process is able to listen to the death of the main process.
* Watchdog process is able to send the `server_shutdown` event once the main process is dead.
* Watchdog process is able to kill itself after sending the server shutdown event.

Once we implement the fetch transport, we can move away from asserting on the logs and rather make sure that the fetch calls are successful. I'm thinking of using a mock server here for the events and confirm that they are sent correctly.

src/telemetry/types.ts
OrKoN, line 51: for a follow-up: can we change enums to be strings instead of ints?

tests/e2e/telemetry.test.ts
```diff
@@ -0,0 +1,150 @@
...
+        2000,
+      );
+      assert.ok(deathMatch, 'Parent death not detected');
+    } finally {
+      cleanupTest(ctx);
+    }
+  }
+
+  it('handles SIGKILL', () => runSignalTest('SIGKILL'));
+  it('handles SIGTERM', () => runSignalTest('SIGTERM'));
```
OrKoN: let's add a windows specific test calling "exec(`taskkill /PID ${child.pid} /T /F`);" to kill the subprocess.

src/telemetry/types.ts
ergunsh: Yes! Considering that the Clearcut server is able to interpret the keys of the proto enum from their names as string, we can directly map the string key to the same string value here.

Added this as a follow up and hoping to tackle this week! All the follow-ups so far:
* We need to update `ChromeChannel` definition on the server side to match the `choices` from the CLI flags.
* File / folder renames for the convention. (e.g. IIUC the folder names should be "kebab-case" and file names should be "PascalCase" -- as a side note, it'd be great to add this to the GEMINI.md as well)
* Sending the MCP client name along with the log events.
* (this one) Renaming enums to be strings.

tests/e2e/telemetry.test.ts
OrKoN: let's add POSIX specific test kill the entire process sub-group (`process.kill(-child.pid, 'SIGTERM');`, minus pid)

tests/e2e/telemetry.test.ts
```diff
@@ -145,6 +145,126 @@ describe('Telemetry E2E', () => {
     }
   }
 
+  async function runWindowsTaskkillTest(t: NodeTestContext): Promise<void> {
```
OrKoN: we could just use the original test function and pass the kill callback to it. We could also skip the test when declaring it I believe.

ergunsh: Done!

tests/e2e/telemetry.test.ts
ergunsh: Done!

ergunsh: As we've checked offline, this is not something NodeJS supports easily (there are no signals coming to the app for this and the processes are killed directly).

In theory, we can go for an architecture where we can have a launcher process launching the watchdog and killing itself, cutting the link between the main process & the watchdog process. Though that'd mean updating:
* The IPC messaging mechanism to use named pipes or something similar instead of direct `stdin` connection.
* The shutdown detection mechanism to use polling on the PID of the main process.

This might be an overkill and first I'm going to test the current architecture in most common MCP clients and see how they work in windows!

Removed the taskkill test.

PR #778, opened 2026-01-15
refactor: remove text from the status code for Network requests

Lightning00Blade:
LLM are trained to know the common status code so we can rely on that when displaying the status code and fall back to text in case of errors or pending reponse.

### DIFF
--- a/src/formatters/NetworkFormatter.ts
+++ b/src/formatters/NetworkFormatter.ts
@@ -183,15 +183,11 @@ export class NetworkFormatter {
     const failure = request.failure();
     let status: string;
     if (httpResponse) {
-      const responseStatus = httpResponse.status();
-      status =
-        responseStatus >= 200 && responseStatus <= 299
-          ? `[success - ${responseStatus}]`
-          : `[failed - ${responseStatus}]`;
+      status = httpResponse.status().toString();
     } else if (failure) {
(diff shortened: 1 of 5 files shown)


src/formatters/NetworkFormatter.ts
```diff
@@ -171,17 +171,24 @@ export class NetworkFormatter {
...
     const failure = request.failure();
     let status: string;
-    if (httpResponse) {
-      const responseStatus = httpResponse.status();
-      status =
-        responseStatus >= 200 && responseStatus <= 299
-          ? `[success - ${responseStatus}]`
-          : `[failed - ${responseStatus}]`;
-    } else if (failure) {
+    if (failure) {
```
Lightning00Blade: Moved failure first because if we have this text should be more helpful than the status code.

OrKoN (approved):

src/formatters/NetworkFormatter.ts
Lightning00Blade: Maybe we need add the status code if it exist in this branch as well.

OrKoN: I am in favour of a regular structure. It is not like if the failure comes first the LLM would necessarily pay more attention (might be quite the opposite)

PR #781, opened 2026-01-15
chore: refactor eval script more

OrKoN:
add args parsing and simplify output (support debug logging + headless).

### DIFF
--- a/scripts/eval_gemini.ts
+++ b/scripts/eval_gemini.ts
@@ -6,7 +6,7 @@
 
 import fs from 'node:fs';
 import path from 'node:path';
-import {describe, test, before, after, afterEach} from 'node:test';
+import {parseArgs} from 'node:util';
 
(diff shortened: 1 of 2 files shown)


scripts/eval_gemini.ts
```diff
@@ -253,49 +264,85 @@ async function runSingleScenario(
...
+        default: false,
+      },
+    },
+    allowPositionals: true,
   });
 
-  after(async () => {
-    await server.stop();
-  });
+  const modelId = values.model ?? 'gemini-2.5-flash';
```
Lightning00Blade: Why do we need to double the defaults? We already set the default in the parseArgs, is it TypeScript?

OrKoN: No, types actually look correct although this is not integrated into the build. Removed.

PR #785, opened 2026-01-16
chore: Remove file unused after bundling from published package

zyzyzyryxy:
build/src/third_party/devtools.js is only used during bundling, but we publish it unnecessarily. We should remove it to not cause confusion and slightly decrease package size.


OrKoN (changes requested):
Thanks! instead of this, let's move re-exports from devtools.ts into the index.ts and delete devtools.ts

Issue #786, opened 2026-01-17
Filling large amount of text result in timeout


kth8:
### Description of the bug

While using this MCP server with my agent, I noticed the fill tool would frequently timeout after 5000ms when typing in large amounts of text. 
![img](https://i.imgur.com/B7idAF5.png)
Naturally I asked my agent to implement a workaround. After some trial and error it seems like replacing `handle.asLocator().fill` with `page.keyboard.type()` in `src/tools/input.ts` worked.
![!img](https://i.imgur.com/hBzt1UQ.png)
Is this something the maintainers would be willing to look at and perhaps implement a more robust solution for than what I could came up with? I can provide more details if needed.

### Reproduction

Type a long paragraph using fill

### Expectation

Fully type the text without timing out

### Chrome DevTools MCP version

0.13.0

### Chrome version

143.0.7499.192

### Coding agent version

opencode 1.1.25

### Node version

v22.22.0

### Operating system

Linux

### Extra checklist

- [ ] I want to provide a PR to fix this bug

OrKoN:
`page.keyboard.type()` only works for the keyboard input and does not handle other types of input controls. It also does not check if the element is ready/able to receive input.

OrKoN:
`page.keyboard.type()` times out as well in a test for me. The locator fill() uses page.keyboard.type() underneath so it makes sense. The reason is that each character is types one by one and that takes longer for larger texts.

PR #787, opened 2026-01-17
fix: increase timeouts for long text input

OrKoN:
Fixes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/786

### DIFF
--- a/scripts/test.mjs
+++ b/scripts/test.mjs
@@ -53,7 +53,7 @@ const nodeArgs = [
   'spec',
   '--test-force-exit',
   '--test',
-  '--test-timeout=30000',
+  '--test-timeout=60000',
   ...flags,
   ...files,
 ];
--- a/src/tools/input.ts
+++ b/src/tools/input.ts
@@ -160,11 +160,10 @@ async function fillFormElement(
       await selectOption(handle, aXNode, value);
     } else {
       // Increase timeout for longer input values.
-      const textLengthIn1kChars = Math.floor(value.length / 1_000);
-      const timeoutPer1kChars = 5_000; // ms
+      const timeoutPerChar = 10; // ms
       const fillTimeout =
         context.getSelectedPage().getDefaultTimeout() +
-        textLengthIn1kChars * timeoutPer1kChars;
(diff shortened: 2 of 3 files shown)


kth8:
@OrKoN thanks for the quick fix. I tried your patch but still getting timed out, after 10 seconds this time instead of 5.
image
This is what my agent came up with to increase timeout to 60 seconds:
image
I think `fill_form` also have this behaviour.

OrKoN:
@kth8 I think the agent is not quite right here because the locator.fill() ends up in keyboard.type:

- https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/api/locators/locators.ts#L499
- https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/api/ElementHandle.ts#L1136

We also do not want a 60 sec timeout as it is too long if there are other issues not related to typing. Perhaps we should just have a different tool like `paste_text` for large blocks of texts so it does not require typing.

Are you sure you correctly re-built and used the server from this branch? can you reproduce the timeout by running the tests from this PR on your machine? can you try adding a test that would demonstrate the timeout? I wonder if there is some dependency on the page content?

kth8:
I think your suggest of adding a separate `paste_text` tool is a good option.

I cloned the branch with your patch and ran it via 
```
git clone -b orkon/input-timeout https://github.com/ChromeDevTools/chrome-devtools-mcp.git
npm ci
npm run build
node build/src/index.js --browserUrl=http://192.168.122.239:9223
```
I tried on another site like chatgpt.com and the tool didn't timeout like before but MCP itself did this time, but I think that may be a client (opencode) configuration issue. A `paste_text` would prevent this and save a lot of time.
image

OrKoN:
Thanks for testing! I think we will land this next week (maybe increasing timeouts linearly according to the input length, instead of per block of 1k chars). But I think I will also look into fixing the fill() method in Puppeteer to switch to pasting the entire text at once for large texts.

Lightning00Blade (approved):

Issue #713, opened 2025-12-27
Fix alert dialog detection and handling capabilities?


martinambrus:
### Description of the bug

There is an issue that was auto-closed for this: #529
However, even today, with running this MCP through Claude Code, I'm having trouble with this functionality.
My use case is simple - when an alert dialog appears on page, the process halts and times outs. The actual handle_dialog() call gets executed only AFTER I manually click OK in the browser (when running interactively - in headless mode, I just have to wait for the timeout). The interesting thing is that this call even thinks it handled the OK button itself - it does not report an error (which is otherwise does, when no dialog is on page because I already clicked OK before handle_dialog could execute but after the MCP found out about a dialog being present on page).

Not sure if this feature is broken only for a handful of us or it's something else but I can confirm that the issue is still there.

### Reproduction

Please see #529 

### Expectation

Please see #529 

### MCP configuration

{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": [
        "chrome-devtools-mcp@latest",
        "--channel=stable",
        "--isolated=true",
        "--viewport=1920x1080"
      ]
    }
  }
}

### Chrome DevTools MCP version

current, latest

### Chrome version

143.0.7499.169

### Coding agent version

2.0.76

### Model version

opus, sonnet, haiku - all of them

### Chat log

what you'd expect - handle_dialog only appears after I press OK, everything else works smoothly afterwards (the call even thinks it handled the OK button itself)

### Node version

24.12.0

### Operating system

Windows Subsystem for Linux (WSL)

### Extra checklist

- [ ] I want to provide a PR to fix this bug

OrKoN:
Thanks for reporting, do you have steps to reproduce? Which process is timing out for you? is the dialog caused by a script evaluation tool call or by input?

martinambrus:
Just as you commented, I'm running a set of fixes and while trying a fix via MCP, Claude Code has this very issue. Here is where I'm currently stuck: 

<img width="793" height="251" alt="Image" src="https://github.com/user-attachments/assets/84104f94-e4c2-403b-8842-4a4024699cff" />

Claude is instructed to test its fixes in browser. It does this but as it tries to navigate away, the onbeforeunload event hits and displays a leaving message: 

<img width="1914" height="266" alt="Image" src="https://github.com/user-attachments/assets/0aa1531c-0dbc-400d-8f3d-5e7e9558ca28" />

[...]

PR #788, opened 2026-01-17
fix: handle beforeunload dialogs in navigations

OrKoN:
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/713

### DIFF
--- a/src/tools/pages.ts
+++ b/src/tools/pages.ts
@@ -125,7 +125,7 @@ export const navigatePage = defineTool({
       .enum(['accept', 'decline'])
       .optional()
       .describe(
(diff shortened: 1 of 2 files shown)


src/tools/pages.ts
```diff
@@ -120,6 +121,12 @@ export const navigatePage = defineTool({
       .boolean()
       .optional()
       .describe('Whether to ignore cache on reload.'),
+    handleBeforeUnload: zod
+      .enum(['accept', 'decline'])
+      .optional()
+      .describe(
+        'Whether to auto accept, decline or ignore potential before unload dialogs triggered by this navigation.',
```
Lightning00Blade: Should we state that the default is accepts?

Lightning00Blade (approved):

PR #792, opened 2026-01-19
chore(build): add devtools-formatter-worker.ts bundle

szuend:
The `devtools-formatter-worker` is a separate entrypoint from the MCP entrypoint. It is identical to the upstream `front_end/entrypoints/formatter_worker/formatter_worker-entrypoint.js` and a required part to make source mapping, formatting and other AST-based DevTools features work.

The implementation consists of 3 parts:
1) A trivial `devtools-formatter-worker.ts` file that imports the DevTools entrypoint.
2) A new bundle output, bundling everything that needs to run in the worker.
3) A call that initializes the `FormatterWorkerPool` with the custom entrypoint. Otherwise we'd use the unbundled entrypoint from `chrome-devtools-frontend` not available to us at runtime.

Note that we could improve code splitting in the future as some things like `core/common` are now bundled twice.

### DIFF
--- a/src/third_party/devtools-formatter-worker.ts
+++ b/src/third_party/devtools-formatter-worker.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2025 Google LLC
+ * Copyright 2026 Google LLC
  * SPDX-License-Identifier: Apache-2.0
  */
 


src/third_party/devtools-formatter-worker.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
OrKoN:
```suggestion
 * Copyright 2026 Google LLC
```

PR #794, opened 2026-01-19
chore: extract issue formatter in preparation to structured content

OrKoN:
This makes it easier to add JSON-format as the output for issues. Console formatter will be changed next.

### DIFF
--- a/src/formatters/IssueFormatter.ts
+++ b/src/formatters/IssueFormatter.ts
@@ -1,0 +1,219 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {ISSUE_UTILS} from '../issue-descriptions.js';
+import {logger} from '../logger.js';
+import {DevTools} from '../third_party/index.js';
+
+export interface IssueFormatterOptions {
+  requestIdResolver?: (requestId: string) => number | undefined;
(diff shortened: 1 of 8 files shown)


src/formatters/consoleFormatter.ts
```diff
@@ -43,10 +48,22 @@ export function formatConsoleEventVerbose(
   context?: McpContext,
 ): string {
   const aggregatedIssue = msg.item;
+  if (aggregatedIssue) {
```
nattallius, line 51: maybe let's have the same check for issues message type as in `formatConsoleEventShort` fro consistency
`msg.type === 'issue' && msg.item`

src/formatters/IssueFormatter.ts
```diff
@@ -0,0 +1,219 @@
...
+      affectedResources.push({
+        uid,
+        data: data,
+        request,
+      });
+    }
+    if (affectedResources.length) {
+      bodyParts.push('### Affected resources');
+    }
+    bodyParts.push(
```
nattallius: Maybe it makes sense to move this block inside` if (affectedResources.length) { ...}` cause it only make sense if there are resources to process. And it keeps the related logic together

src/formatters/consoleFormatter.ts
OrKoN: I think `msg.type === 'issue'` is redundant here as typescript types say item is always an AggregatedIssue.

PR #796, opened 2026-01-19
chore: Implement uninstall_extension tool

nattallius:


src/main.ts
```diff
@@ -141,8 +141,8 @@ function registerTool(tool: ToolDefinition): void {
     return;
   }
   if (
-    tool.annotations.category === ToolCategory.EXTENSIONS &&
-    args.categoryExtensions === false
+    tool.annotations.conditions?.includes('experimentalExtensionSupport') &&
```
OrKoN: is this change needed?

nattallius: True, will rollback

OrKoN (approved):

PR #800, opened 2026-01-20
fix: improve performance tool description

OrKoN:
Drive-by: refactor and improve eval scripts.

### DIFF
--- a/scripts/eval_gemini.ts
+++ b/scripts/eval_gemini.ts
@@ -8,11 +8,7 @@ import fs from 'node:fs';
 import path from 'node:path';
 import {parseArgs} from 'node:util';
 
-import {
-  GoogleGenerativeAI,
-  type FunctionDeclaration,
-  SchemaType,
-} from '@google/generative-ai';
+import {GoogleGenAI, mcpToTool} from '@google/genai';
 import {Client} from '@modelcontextprotocol/sdk/client/index.js';
 import {StdioClientTransport} from '@modelcontextprotocol/sdk/client/stdio.js';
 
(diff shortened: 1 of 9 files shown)


scripts/eval_scenarios/emulation_test.ts
```diff
@@ -12,8 +12,8 @@ export const scenario: TestScenario = {
   prompt: 'Emulate offline network conditions.',
   maxTurns: 2,
   expectations: calls => {
-    const emulate = calls.find(c => c.name === 'emulate');
-    assert.ok(emulate, 'Should call emulate tool');
-    assert.strictEqual(emulate.args.networkConditions, 'Offline');
+    assert.strictEqual(calls.length, 1);
```
Lightning00Blade, line 15: How does this calls to maxTurns correlate, are they not the same?

scripts/eval_gemini.ts
```diff
@@ -147,137 +117,48 @@ async function runSingleScenario(
...
+    debugLog(`\n--- Response ---\n${result.text}`);
 
     debugLog('\nVerifying expectations...');
-    expectations(allCalls);
+    scenario.expectations(allCalls);
   } finally {
-    await client?.close();
-    await transport?.close();
+    try {
+      if (client) {
```
Lightning00Blade: Super nit use optional chaining.

scripts/eval_gemini.ts
```diff
@@ -297,12 +178,18 @@ async function main() {
...
+      repeat: {
+        type: 'boolean',
+        default: false,
+      },
     },
     allowPositionals: true,
   });
 
-  const modelId = values.model;
+  const modelId = values.model ?? 'gemini-2.5-flash';
```
Lightning00Blade: Again is this nessisary if we have the default in the parseArgs?

scripts/eval_gemini.ts
```diff
@@ -319,16 +206,31 @@ async function main() {
...
+            console.log(
+              `Running scenario: ${path.relative(ROOT_DIR, scenarioPath)} (Run ${i}/3)`,
+            );
+          }
+          await runSingleScenario(
+            scenarioPath,
+            apiKey,
+            server,
+            modelId,
+            debug ?? false,
```
Lightning00Blade: Do we need this, there is a false default value?

scripts/eval_scenarios/emulation_test.ts
OrKoN: maxTurns is the maximum and the actual one can be lower

Lightning00Blade: I am not sure I understand, if the actual number here for example can be 1 or 2 why are we checking for exact value?

OrKoN: because there should be no other function calls for this prompt. What second function call do you expect?

Lightning00Blade: Then why do we give it max of 2 in the config?

OrKoN: To make sure the model stops by itself after 1 function call.

PR #802, opened 2026-01-21
chore: create a console formatter class

OrKoN:
This cleans up the McpResponse and helps with returning the structured content.


src/formatters/ConsoleFormatter.ts
```diff
@@ -0,0 +1,174 @@
...
+  type TargetUniverse,
+} from '../DevtoolsUtils.js';
+import type * as DevTools from '../third_party/index.js';
+import type {ConsoleMessage} from '../third_party/index.js';
+
+export interface ConsoleFormatterOptions {
+  fetchDetailedData?: boolean;
+  id?: number;
+  devTools?: TargetUniverse;
+  stackTrace?: DevTools.DevTools.StackTrace.StackTrace.StackTrace;
```
szuend: I think the `stackTrace` argument is unused since we calculate it in `#loadDetailedData`?

src/formatters/ConsoleFormatter.ts
```diff
@@ -0,0 +1,174 @@
...
+  #resolvedStackTrace?: DevTools.DevTools.StackTrace.StackTrace.StackTrace;
+  #id?: number;
+
+  private constructor(
+    msg: ConsoleMessage | Error,
+    options?: ConsoleFormatterOptions,
+  ) {
+    this.#msg = msg;
+    this.#id = options?.id;
+    this.#resolvedStackTrace = options?.stackTrace;
```
szuend: Same here, i think this is not used?

szuend (approved):
Thanks!

src/formatters/ConsoleFormatter.ts
OrKoN: it is used by the tests

src/formatters/ConsoleFormatter.ts
OrKoN: this one is too

PR #804, opened 2026-01-21
chore: structured content for console

OrKoN:


src/McpResponse.ts
```diff
@@ -511,6 +517,9 @@ Call ${handleDialog.name} to handle it before continuing.`);
             return message.toString();
           }),
         );
+        structuredContent.consoleMessages = data.items.map(message =>
```
nattallius, line 512: As well as here you don't need to distinguish between formatters  lines above, 514-517. You can just call `message.toString()` 

as well as `formatConsoleData`'s `toStringDetailed` . It looks like you can get rid of it and do like other
`if(data.detailedConsoleMessage) {
response.push(data.toStringDetailed());
}`

nattallius (approved):

src/McpResponse.ts
OrKoN: Good point!

PR #805, opened 2026-01-21
chore: Implement ClearcutSender HTTP transport for telemetry disabled by default

ergunsh:
This PR completes the telemetry system by implementing the transport layer for `ClearcutSender`. It enables actual HTTP communication with the Clearcut backend, handling event batching, rate limiting, and reliable delivery, including robust shutdown handling.

**Key Changes:**
*   **HTTP Transport**: Implemented `fetch`-based transport sending `POST` requests to the Clearcut HTTP server.
*   **Event Batching**: Events are now buffered and flushed periodically (default: 15 minutes) or on shutdown.
*   **Reliability & Rate Limiting**:
    *   **Server-Side Backoff**: Respects `next_request_wait_millis` from server responses to handle rate limiting dynamically.
    *   **Transient Error Retries**: Failed requests (5xx, 429) result in events being requeued for the next flush.
    *   **Request Timeouts**: Enforced 30s timeout on requests to prevent hanging processes.
    *   **Session Rotation**: Automatically rotates session IDs every 24 hours.
*   **Safety & Stability**:
    *   **Buffer Overflow Protection**: Caps the buffer at 1000 events to prevent memory leaks, dropping oldest events if necessary.
    *   **Optimistic Removal**: Prevents race conditions and duplicate events during shutdown by optimistically removing events from the buffer before sending.
*   **Testing Improvements**:
    *   **E2E Robustness**: Updated E2E tests to use a mock web server instead of relying on the logger to log specific lines.

**Implementation Roadmap:**
These changes finalize the planned telemetry architecture:
1.  **CLI & Opt-out Mechanism ([Merged](https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/757))**
2.  **Logger Scaffolding & Integration ([Merged](https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/758))**
3.  **Persistence Layer ([Merged](https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/766))**
4.  **Watchdog Process Architecture ([Merged](https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/769))**
5.  **Transport, Batching & Retries (This PR):**
    *   Finalized `ClearcutSender` with HTTP transport, batching, and server-directed backoff strategies.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -209,7 +209,8 @@ export const cliOptions = {
     // Marked as `false` until the feature is ready to be enabled by default.
     default: false,
     hidden: true,
-    describe: 'Set to false to opt-out of usage statistics collection. Google collects usage data to improve the tool, handled under the Google Privacy Policy (https://policies.google.com/privacy). This is independent from Chrome browser metrics.',
+    describe:
+      'Set to false to opt-out of usage statistics collection. Google collects usage data to improve the tool, handled under the Google Privacy Policy (https://policies.google.com/privacy). This is independent from Chrome browser metrics.',
   },
   clearcutEndpoint: {
     type: 'string',


src/telemetry/watchdog/clearcut-sender.ts
```diff
@@ -7,15 +7,42 @@
...
 import {logger} from '../../logger.js';
-import type {ChromeDevToolsMcpExtension, OsType} from '../types.js';
+import type {
+  ChromeDevToolsMcpExtension,
+  LogRequest,
+  LogResponse,
+  OsType,
+} from '../types.js';
 
+const INCLUDE_PID_HEADER = process.env.INCLUDE_PID_HEADER === 'true';
```
ergunsh: The environment variables are needed for configuring the server on e2e tests. LMK if you have an alternative idea (e.g. passing from CLI args) or if we should prefix these environment variables so that it doesn't conflict with other tools (considering that the server is running in the user machine and these configuration options are only for internal testing).

OrKoN: let's pass these options when creating the sender class and remove env variables. In tests that test the sender, let's pass test params in the constructor. Other tests should not really trigger sendings.

ergunsh: For the e2e tests, should we continue depending on the logger logging the correct things for getting the watchdog process pid and assertions? (that `server_start` and `server_shutdown` is sent?)

OrKoN: for e2e tests for metrics let's use hidden cli args instead of env vars

ergunsh: Done!

src/cli.ts
```diff
@@ -211,6 +211,21 @@ export const cliOptions = {
     hidden: true,
     describe: 'Set to false to opt-out of usage statistics collection.',
```
OrKoN: With this change, let's add the disclaimer and the links to policies to the description of the CLI argument.

ergunsh: Done!

PR #819, opened 2026-01-23
chore: implement list_extensions tool for local unpacked  extensions

nattallius:


src/tools/extensions.ts
```diff
@@ -22,6 +22,7 @@ export const installExtension = defineTool({
   schema: {
     path: zod
       .string()
+      .trim()
```
OrKoN: is it required?

src/tools/extensions.ts
```diff
@@ -48,3 +49,29 @@ export const uninstallExtension = defineTool({
...
+      response.appendResponseLine('No extensions installed.');
+      return;
+    }
+    const table = extensions.map(ext => ({
+      Name: ext.name,
+      ID: ext.id,
+      Version: ext.version,
+      Enabled: ext.isEnabled ? 'Yes' : 'No',
+    }));
+    response.appendResponseLine(JSON.stringify(table, null, 2));
```
OrKoN: let's indicate that the response should list extensions and implement formatting in McpResponse.format(), and data retrieval in McpResponse.handle(). Additionally, let's add the output to the structured content in the MCP response.

src/tools/extensions.ts
```diff
@@ -48,3 +49,29 @@ export const uninstallExtension = defineTool({
...
+    const extensions = context.listExtensions();
+    if (extensions.length === 0) {
+      response.appendResponseLine('No extensions installed.');
+      return;
+    }
+    const table = extensions.map(ext => ({
+      Name: ext.name,
+      ID: ext.id,
+      Version: ext.version,
+      Enabled: ext.isEnabled ? 'Yes' : 'No',
```
OrKoN: let's use `enabled`, `disabled` instead of yes/no

src/utils/ExtensionRegistry.ts
```diff
@@ -0,0 +1,55 @@
...
+  #extensions = new Map<string, InstalledExtension>();
+
+  async registerExtension(id: string, extensionPath: string): Promise<InstalledExtension> {
+    const manifestPath = path.join(extensionPath, 'manifest.json');
+    const manifestContent = await fs.readFile(manifestPath, 'utf-8');
+    const manifest = JSON.parse(manifestContent);
+    const name = manifest.name ?? 'Unknown';
+    const version = manifest.version ?? 'Unknown';
+
+    const extension: InstalledExtension = {
```
OrKoN: nit: probably do not need `: InstalledExtension`

src/utils/ExtensionRegistry.ts
```diff
@@ -0,0 +1,55 @@
...
+      name,
+      version,
+      isEnabled: true,
+      path: extensionPath,
+    };
+    this.add(extension);
+    return extension;
+  }
+
+  add(extension: InstalledExtension): void {
```
OrKoN: do we need add? can we merge register and add?

tests/tools/extensions.test.ts
```diff
@@ -71,4 +72,36 @@ describe('extension', () => {
       );
     });
   });
+  it('lists installed extensions', async () => {
```
OrKoN, line 77: in this tool test let's verify that the response is configured to list extensions. Let's move the testing of the response data to McpResponse tests (including structured data).

src/tools/extensions.ts
OrKoN: for the text formatting let's output each extension data as a new line.

src/tools/extensions.ts
```diff
@@ -48,3 +49,29 @@ export const uninstallExtension = defineTool({
...
+  },
+  schema: {},
+  handler: async (request, response, context) => {
+    const extensions = context.listExtensions();
+    if (extensions.length === 0) {
+      response.appendResponseLine('No extensions installed.');
+      return;
+    }
+    const table = extensions.map(ext => ({
+      Name: ext.name,
```
OrKoN: let's follow the convention of lower `camelCase` for attributes.

src/tools/extensions.ts
nattallius: I believe so, it's better clean the path before using it. also i use that path for reading manifest

OrKoN: what if a file or a directory is ending in a whitespace? we do not do this for any other file path inputs so let's remove this here and make sure we error if the provided path is incorrect.

src/tools/extensions.ts
OrKoN: I think this was not addressed, re-opening

src/McpResponse.ts
```diff
@@ -531,6 +545,27 @@ Call ${handleDialog.name} to handle it before continuing.`);
...
+      } else {
+        const parts = [];
+        for (const extension of data.extensions) {
+          parts.push(
+            [
+              `Name: ${extension.name}`,
+              `ID: ${extension.id}`,
+              `Version: ${extension.version}`,
+              `Status: ${extension.isEnabled ? 'Enabled' : 'Disabled'}`,
+            ].join('\n'),
```
OrKoN: `id=${extendion.id} "${extension.name}" v${extension.version} ${extension.isEnabled ? 'Enabled' : 'Disabled'}`

src/tools/extensions.ts
```diff
@@ -48,3 +48,18 @@ export const uninstallExtension = defineTool({
     response.appendResponseLine(`Extension uninstalled. Id: ${id}`);
   },
 });
+
+export const listExtensions = defineTool({
+  name: 'list_extensions',
+  description:
+    'Lists all installed extensions, including their name, ID, version, and enabled status.',
```
OrKoN:
```suggestion
    'Lists all extensions via this server, including their name, ID, version, and enabled status.',
```

PR #830, opened 2026-01-25
feat: add a skill for using chrome-devtools-mcp

OrKoN:
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/710
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/709

### DIFF
--- a/skills/chrome-devtools/SKILL.md
+++ b/skills/chrome-devtools/SKILL.md
@@ -1,0 +1,44 @@
+---
+name: chrome-devtools
+description: Uses Chrome DevTools via MCP for efficient debugging, troubleshooting and browser automation. Use when debugging web pages, automating browser interactions, analyzing performance, or inspecting network requests.
+---
+
+## Core Concepts
+
+**Browser lifecycle**: Browser starts automatically on first tool call using a persistent Chrome profile. Configure via CLI args in the MCP server configuration: `npx chrome-devtools-mcp@latest --help`.
+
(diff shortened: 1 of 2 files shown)


skills/chrome-devtools/SKILL.md
```diff
@@ -0,0 +1,44 @@
...
+3. Snapshot: `take_snapshot` to understand page structure
+4. Interact: Use element `uid`s from snapshot for `click`, `fill`, etc.
+
+### Efficient data retrieval
+
+- Use `filePath` parameter for large outputs (screenshots, snapshots, traces)
+- Use pagination (`pageIdx`, `pageSize`) and filtering (`types`) to minimize data
+- Set `includeSnapshot: false` on input actions unless you need updated page state
+
+### Tool selection
```
natorion, line 29: Should this list network and performance too? I am fine to start without it, we can add it incrementally.

natorion:
Will this already be bundled and surfaced automatically within the Gemini CLI extension and the Claude Code plugin?

skills/chrome-devtools/SKILL.md
OrKoN: This only includes common tools and patterns that might not be obvious from the tool description to reduce the token size.

OrKoN:
> Will this already be bundled and surfaced automatically within the Gemini CLI extension and the Claude Code plugin?

I can test after I land on the main branch but from the docs both claude and gemini-cli should be picking this up

Issue #473, opened 2025-10-26
Expose accessibility findings from lighthouse


BogdanCerovac:
### Is your feature request related to a problem? Please describe.

As Lighthouse provides some limited automatic accessibility testing I would like to see it in the MCP as well. It is always an option to use other, specialized MCPs for accessibility needs, but I think offering Lighthouse accessibility testing results directly in Chrome DevTools MCP could be beneficial to awareness and even helping with suggestions for basic fixes.

### Describe the solution you'd like

As a minimum, tools would get a section called Accessibility that would have at least two tools:

1. getting the accessibility **score** for a page (as per https://developer.chrome.com/docs/lighthouse/accessibility/scoring) - I am aware that scores are not representing the whole situation as per limitations of automatic tests, but it would be useful to get at least a hint that the score represents.
2. getting **accessibility audit details** - a list of errors and impacted code / DOM elements - so that AI can try to come with fix suggestions.

### Describe alternatives you've considered

Using snapshot tool does not offer insights into automatically detected issues, so the only alternative now is to use a specialized MCP for accessibility.

### Additional context

I am using/testing multiple other accessibility oriented MCPs now, but if I consider people just starting with accessibility - it would be a great start and better to have something than nothing.

natorion:
Background: We are evaluating right now how to best make LH audits and Issues available on the MCP server.

@BogdanCerovac how would you expect your AI agent would make use of this in detail? What prompts would you write that would spark the MCP client calling the relevant MCP tool on our end? What prevents you from adding LH directly to your tool chain und using directly?

BogdanCerovac:
@natorion - nice, thanks.

As a minimum, I would like to get the accessibility score, list all accessibility issues and, if possible, also get some fix recommendations (I am aware that this can be difficult for some cases, but at least give some basic suggestions could help some people to investigate more).

Basically similar to what is shown in the video made for performance issues on
https://developer.chrome.com/blog/chrome-devtools-mcp

But with accessibility in mind, like for example support prompts like:

- What is the accessibility score on https://localhost
- Check for accessibility issues on https://localhost
- Try to fix accessibility issues on https://localhost

I've done some testing and made my own MCP server for accessibility and it worked quite well (limited by the capability of automatic testing ofc.), but I think we should get some basics out of Chrome DevTool MCP for more reach (especially useful for people that are early on their accessibility journey and not (yet) focused on accessibility.

PR #831, opened 2026-01-25
feat: integrate Lighthouse audits

OrKoN:
This PR adds tool to perform navigation and snapshot audits using Lighthouse for Accessibility, SEO and Best practices. For performance audits existing tools can be used.

Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/473


src/tools/lighthouse.ts
```diff
@@ -0,0 +1,177 @@
...
+      outputDirPath,
+      formats = ['json', 'html'],
+    } = request.params;
+
+    const flags: Flags = {
+      onlyCategories: categories,
+      output: formats,
+    };
+
+    if (device === 'desktop') {
```
OrKoN, line 63: TODO: can we re-use these definitions somehow?

OrKoN:
This PR is not ready to land but ready for initial review.

OrKoN:
@zyzyzyryxy could you please review the bundling setup and perhaps you could run an import CL to see if there will be any complaints from linters?

OrKoN:
@paulirish could you please review for the lighthouse integration?

src/tools/lighthouse.ts
```diff
@@ -0,0 +1,166 @@
...
+      flags.formFactor = 'desktop';
+      flags.screenEmulation = {
```
OrKoN, line 65: @paulirish do these params look accurate? is there a way to re-use these from some Lighthouse or dtf configs?

zyzyzyryxy:
> @zyzyzyryxy could you please review the bundling setup and perhaps you could run an import CL to see if there will be any complaints from linters?

~~I run some tests, I think THIRD_PARTY_NOTICES is no longer generating. Was this an intentional change?~~

Scratch that, I used wrong command to create the package.

OrKoN:
@zyzyzyryxy PT(another)L, the notices are now combined into the one file (source is copied to our source).

docs/tool-reference.md
```diff
@@ -1,7 +1,9 @@
 <!-- AUTO GENERATED DO NOT EDIT - run 'npm run docs' to update-->
 
-# Chrome DevTools MCP Tool Reference (~6719 cl100k_base tokens)
+# Chrome DevTools MCP Tool Reference (~6980 cl100k_base tokens)
 
+- **[Audits](#audits)** (1 tools)
```
OrKoN: @natorion should we introduce a new category or should I add it to the debugging category?

natorion: Let's add it as default for now. Lighthouse is a fundamental piece of DT, so it makes sense that its there as default.

OrKoN: It is in the default, I was asking about whether we need a new category called Audits or should we add it to the existing debugging category

natorion: no, I think debugging is fine.

src/tools/lighthouse.ts
paulirish: yeah  you can pass in a 'config', in addition to flags.

see clients/lightrider/lightrider-entry.js 

default-config.js == mobile
but desktop-config.js == desktop. (lr-desktop is for lightrider exclusively)

desktop-config is exposed on the lh main exports as `desktopConfig` 

so you can do that. 

(we dont use this config trick in the devtools frontend case because devtools frontend is separately managing emulation, throttling etc.)

OrKoN: I think you mentioned that we would not want throttling for the subset of audits we expose and it looks like default configs define throttling? Would I need both flags and configs or can I only use configs?

PR #837, opened 2026-01-26
feat: add background parameter to new_page tool

### LINKED ISSUE #826: Feature request: Open `new_page` in background, similar to cmd+click in Chrome

When using `new_page` to open a URL, the new tab is always focused and brought to the front. This might not always be wanted e.g. if it's opened to explore next/later, similar to cmd+click in normal Chrome.

Add a `bringToFront` boolean parameter to the `new_page` tool:

```typescript
{
  url: string,
  timeout?: number,
  bringToFront?: boolean  // New parameter, default true
}
```

When `bringToFront: false`, the new page should be created without focusing it or bringing it to the front.

This parameter name is chosen to mirror select_page. Alternatively it could be `openInBackground` but feel like the consistency is helpful.

Using `new_page` followed by `select_page` with `bringToFront: true` to switch back to the original page. This works but causes a visible flicker as the new tab briefly appears before switching back.

### Additional context

_No response_

domdomegg:
## Summary

- Add `background` boolean parameter to the `new_page` tool
- When set to `true`, the new page opens in the background without bringing it to the front
- Uses Puppeteer's existing `background` option for `browser.newPage()`

Fixes #826

### DIFF
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -170,7 +170,7 @@
 **Parameters:**
 
 - **url** (string) **(required)**: URL to load in a new page.
-- **background** (boolean) _(optional)_: Whether to open the page in the background without bringing it to the front.
+- **background** (boolean) _(optional)_: Whether to open the page in the background without bringing it to the front. Default is false (foreground).
 - **timeout** (integer) _(optional)_: Maximum wait time in milliseconds. If set to 0, the default timeout will be used.
 
 ---


src/tools/pages.ts
```diff
@@ -87,10 +87,16 @@ export const newPage = defineTool({
   },
   schema: {
     url: zod.string().describe('URL to load in a new page.'),
+    background: zod
+      .boolean()
+      .optional()
+      .describe(
+        'Whether to open the page in the background without bringing it to the front.',
```
```suggestion
        'Whether to open the page in the background without bringing it to the front. Default is false (foreground).',
```

domdomegg: Applied!

PR #871, opened 2026-02-03
chore: emit UncaughtError events for Runtime.exceptionThrown

szuend:
Follow-up PRs will implement formatting of `UncaughtError`s and actually collecting them.


szuend:
I'll rename `PageIssueSubscriber` to `PageEventSubscriber` in a follow-up given that it now also listens to `Runtime.exceptionThrown`.

PR #876, opened 2026-02-03
feat: include source-mapped stack trace for uncaught errors

szuend:
Note that we replace the existing "pageerror" handler with our new `UncaughtError` handler. The latter has a structured stack trace attached which we can source map via DevTools.

This regresses the current functionality slightly, as we lose "pageerror"s for workers and oopif.

### DIFF
--- a/tests/tools/console.test.ts
+++ b/tests/tools/console.test.ts
@@ -49,7 +49,7 @@ describe('console', () => {
         await listConsoleMessages.handler({params: {}}, response, context);
         const formattedResponse = await response.handle('test', context);
         const textContent = getTextContent(formattedResponse.content[0]);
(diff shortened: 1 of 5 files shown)


tests/tools/console.test.ts
```diff
@@ -49,7 +49,7 @@ describe('console', () => {
         await listConsoleMessages.handler({params: {}}, response, context);
         const formattedResponse = await response.handle('test', context);
         const textContent = getTextContent(formattedResponse.content[0]);
-        assert.ok(textContent.includes('msgid=1 [error] undefined (0 args)'));
+        assert.ok(textContent.includes('msgid=1 [error] Uncaught  (0 args)'));
```
szuend, line 52: Is this diff acceptable?

OrKoN: yep!

PR #884, opened 2026-02-04
fix: use 1-based line/column and fix wasm offsets in stack frames

szuend:
`UILocation.linkText` does the right thing when formatting a location, e.g. a wasm location will be formatted as `foo.wasm:0x95`.


src/formatters/ConsoleFormatter.ts
```diff
@@ -175,6 +175,7 @@ export class ConsoleFormatter {
 
     return [
       '### Stack trace',
+      'Note: line and column numbers use 1-based indexing',
```
OrKoN: maybe let's add it after the stacktrace

szuend: Done

PR #890, opened 2026-02-05
chore: introduce SymbolizedError class

szuend:
The "SymbolizedError" class represents a fully resolved error: The stack trace is fully resolved and (in the future), the full `Error.cause` chain is also fully resolved.

We'll use the `SymbolizedError` for both "uncaught exceptions" as well as when logging `Error` objects to the console (e.g. `console.log(new Error())`. This means the `resolvedArgs` array in `ConsoleFormatter` will contain one `SymbolizedError` instance for every `Error` object logged.

### DIFF
--- a/src/DevtoolsUtils.ts
+++ b/src/DevtoolsUtils.ts
@@ -227,6 +227,73 @@ const SKIP_ALL_PAUSES = {
   },
 };
 
+/**
+ * Constructed from Runtime.ExceptionDetails of an uncaught error.
+ *
+ * TODO: Also construct from a RemoteObject of subtype 'error'.
+ *
+ * Consists of the message, a fully resolved stack trace and a fully resolved 'cause' chain.
+ */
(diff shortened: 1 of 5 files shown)


szuend:
@OrKoN: Two things I find awkward in this CL: 1) We have to pass the `resolvedStackTraceForTesting` into two places, and 2) the `resolvedStackTrace` only applies to plain console messages, where-as otherwise it's part of `SymbolizedError`.

OrKoN:
Can we have a SymbolizedStackTrace class and re-use it in both?

OrKoN (approved):

szuend:
> Can we have a SymbolizedStackTrace class and re-use it in both?

Not sure I follow, what would be the difference to `DevTools.StackTrace.StackTrace`?

OrKoN:
It would be an API that we can take right into Puppeteer :)

PR #902, opened 2026-02-06
feat: show message and stack trace in details when console.log'ging Error objects

szuend:
This PR improves the `get_console_message` tool when logging `Error` objects. We use the existing `getExceptionDetails` CDP command to retrieve the structured stack trace, we'll then source map.

Example:

```js
try {
  compute();
} catch (e) {
  console.log('Compute failed', e);
}
```

Before:
```txt
### Arguments
Arg #0: Compute failed
Arg #1: {}
```

After:
```txt
### Arguments
Arg #0: Compute failed
Arg #1: ComputeError: Invariant violation
at compute (foo.ts:1:20)
at <anonymous> (main.ts:2:8)
Note: line and column numbers use 1-based indexing
```

### DIFF
--- a/tests/formatters/ConsoleFormatter.test.ts
+++ b/tests/formatters/ConsoleFormatter.test.ts
@@ -34,7 +34,7 @@ const createMockMessage = (
   } as unknown as ConsoleMessage;
 };
 
-describe.only('ConsoleFormatter', () => {
+describe('ConsoleFormatter', () => {
   describe('toString', () => {
     it('formats a console.log message', async t => {
       const message = createMockMessage({


tests/formatters/ConsoleFormatter.test.ts
```diff
@@ -31,7 +34,7 @@ const createMockMessage = (
   } as unknown as ConsoleMessage;
 };
 
-describe('ConsoleFormatter', () => {
+describe.only('ConsoleFormatter', () => {
```
OrKoN:
```suggestion
describe('ConsoleFormatter', () => {
```

OrKoN (approved):

PR #923, opened 2026-02-09
fix: limit stack traces to 50 lines

szuend:
This PR refactors the stack trace formatting logic a bit: We split the `formatStackTrace` function into 2: One for the top-level stack trace and the rest for the recursive ones. We also return lines separately so we can slice the result later.


OrKoN (approved):
Thanks!

PR #939, opened 2026-02-12
docs: Add 'Progressive Complexity' and 'Reference over Value' design principles.

natorion:


docs/design-principles.md
```diff
@@ -8,3 +8,5 @@ Apply them with nuance.
 - **Small, Deterministic Blocks**: Give agents composable tools (Click, Screenshot), not magic buttons.
 - **Self-Healing Errors**: Return actionable errors that include context and potential fixes.
 - **Human-Agent Collaboration**: Output must be readable by machines (structured) AND humans (summaries).
+- **Progressive Complexity**: Tools should be simple by default (high-level actions) but offer advanced optional arguments for power users.
+- **Reference over Value**: for heavy assets (screenshots, traces, videos), return a file path or resource URI, never the raw data stream.
```
OrKoN: I am not sure it always makes sense. For example, returning a screenshot always as a file path would require the client to make an extra LLM eval to process it whereas directly returning it as an MCP-supported image response avoids an extra roundtrip.

PR #941, opened 2026-02-12
feat: add experimental screencast recording tools

### LINKED ISSUE #878: Screen cast recording

Allow the agent to record a video of a page.

https://pptr.dev/api/puppeteer.page.screencast

https://pptr.dev/api/puppeteer.page.screencast

I did have some success getting the agent to take a lot of screenshots and run ffmpeg itself to create the video. 

### Additional context

_No response_

dobesv:
## Summary

Adds `screencast_start` and `screencast_stop` MCP tools that allow agents to record a video of a page using Puppeteer's `page.screencast()` API.

Only enabled if the command line option `--experimental-screencast` is provided

Closes #878.

### `screencast_start`
Starts recording a screencast (video) of the selected page in mp4 format.

| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `path` | string (optional) | temp file | Output file path |

### `screencast_stop`
Stops the active recording and reports the saved file path.

## Design

- **Start/stop pair** following the `performance_start_trace` / `performance_stop_trace` pattern
- **State management** via `getScreenRecorder()` / `setScreenRecorder()` on the Context interface
- **Category**: `DEBUGGING` (alongside `take_screenshot`)
- **ffmpeg dependency**: Clear error message when ffmpeg is not installed
- **Temp file fallback**: When no `path` is provided, creates a temp file with the correct extension

### DIFF
--- a/src/tools/screencast.ts
+++ b/src/tools/screencast.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2025 Google LLC
+ * Copyright 2026 Google LLC
  * SPDX-License-Identifier: Apache-2.0
  */
 
--- a/tests/tools/screencast.test.ts
+++ b/tests/tools/screencast.test.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2025 Google LLC
+ * Copyright 2026 Google LLC
  * SPDX-License-Identifier: Apache-2.0
  */
 


OrKoN:
@dobesv would you be able to sign the CLA for this PR?

dobesv:
I did sign it but it seems to be looking at the co-author line added by the AI agent and saying the agent has to sign a cla as well?

OrKoN:
@dobesv could you update commits to remove the AI agent authorship? otherwise you would need the AI agent to sign the CL.

src/tools/screencast.ts
```diff
@@ -0,0 +1,142 @@
...
+    conditions: ['screencast'],
+  },
+  schema: {
+    filePath: zod
+      .string()
+      .optional()
+      .describe(
+        'The absolute file path, or a file path relative to the current working directory, to save the screencast to. For example, recording.webm. If not specified, a temporary file will be created.',
+      ),
+    format: zod
```
OrKoN: how critical are all those parameters? could we set sensible defaults and minimize the number of tokens needed? e.g., perhaps filePath, scale and speed are the only needed?

dobesv: Good point, I think we can just use filePath for now and see if we feel a need for more later.  Even with scale and speed I'm not really sure the LLM would provide interesting values for those.

dobesv:
I cleaned up the commit and removed extra options other than `path` for now.

src/cli.ts
```diff
@@ -173,6 +173,11 @@ export const cliOptions = {
     describe: 'Whether to enable interoperability tools',
     hidden: true,
   },
+  experimentalScreencast: {
+    type: 'boolean',
+    describe: 'Whether to enable screencast tools',
+    hidden: true,
```
OrKoN: I think we can remove hidden and `npm run docs` to re-generate the docs.

src/cli.ts
```diff
@@ -173,6 +173,11 @@ export const cliOptions = {
     describe: 'Whether to enable interoperability tools',
     hidden: true,
   },
+  experimentalScreencast: {
+    type: 'boolean',
+    describe: 'Whether to enable screencast tools',
```
OrKoN: let's document in the description that ffmpeg is required for this to function.

src/tools/screencast.ts
```diff
@@ -0,0 +1,108 @@
...
+
+async function generateTempFilePath(): Promise<string> {
+  const dir = await fs.mkdtemp(path.join(os.tmpdir(), 'chrome-devtools-mcp-'));
+  return path.join(dir, `screencast.mp4`);
+}
+
+export const startScreencast = defineTool({
+  name: 'screencast_start',
+  description:
+    'Starts recording a screencast (video) of the selected page in mp4 format. Requires ffmpeg to be installed on the system.',
```
OrKoN: I think the tool call itself does not need to document the ffmpeg dependency as it is something for the user to ensure they have rather the agent to install for them?

OrKoN: I think the error message would be sufficient so that we do not need this documentation on each turn.

src/tools/screencast.ts
```diff
@@ -0,0 +1,108 @@
...
+            'Install ffmpeg (https://ffmpeg.org/) and ensure it is available in your PATH.',
+        );
+      }
+      throw err;
+    }
+
+    context.setScreenRecorder({recorder, filePath: resolvedPath});
+
+    response.appendResponseLine(
+      `Screencast recording started. The recording will be saved to ${resolvedPath}. Use screencast_stop to stop recording.`,
```
OrKoN:
```suggestion
      `Screencast recording started. The recording will be saved to ${resolvedPath}. Use ${stopScreencast.name} to stop recording.`,
```

src/cli.ts
OrKoN: I am gonna handle these.

PR #958, opened 2026-02-13
feat: `--slim` mode for maximum token savings

OrKoN:
- navigate
- run js
- take screenshot

That's it. Basic functionality for a low (~368) token price.

### DIFF
--- a/skills/chrome-devtools/SKILL.md
+++ b/skills/chrome-devtools/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: chrome-devtools
-description: Uses Chrome DevTools via MCP for efficient debugging, troubleshooting and browser automation. Use when debugging web pages, automating browser interactions, analyzing performance, or inspecting network requests.
+description: Uses Chrome DevTools via MCP for efficient debugging, troubleshooting and browser automation. Use when debugging web pages, automating browser interactions, analyzing performance, or inspecting network requests. This skill does not apply to `--slim` mode (MCP configuration).
 ---
 
 ## Core Concepts


OrKoN:
@natorion wdyt?

natorion:
A few thoughts:
- I like the concept
- Should it be called "lean" instead of "minimal"?
- Is error reporting of evaluate_script sufficient for any common AI agent to act on the errors? Or will this just produce a frustrating experience?

OrKoN:
I like `--lean`!

OrKoN:
@natorion updated to slim with the reference generation (slim mode consumes 359 tokens).

natorion:
One more thing: Should we mention different tool categories and modes in the general Chrome DevTools Skill?

OrKoN:
> One more thing: Should we mention different tool categories and modes in the general Chrome DevTools Skill?

let me think about it a bit

OrKoN:
> > One more thing: Should we mention different tool categories and modes in the general Chrome DevTools Skill?
> 
> let me think about it a bit

so I think we do not need a skill since the mode is low complexity (most complexity lies in code generation and interpreting screenshots for which generic skills can be used). I also worry that combining two modes in one skill would leave to confusion. Any better ideas?

natorion:
> > > One more thing: Should we mention different tool categories and modes in the general Chrome DevTools Skill?
> > 
> > 
> > let me think about it a bit
> 
> so I think we do not need a skill since the mode is low complexity (most complexity lies in code generation and interpreting screenshots for which generic skills can be used). I also worry that combining two modes in one skill would leave to confusion. Any better ideas?

I was mostly thinking to say: In case you don't find tools mentioned in the skills, make sure that there are tooling categories that are not default or something. This way the agent can start fixing the mcp configs.

OrKoN:
Updated the skill to exclude `--slim`.

PR #960, opened 2026-02-13
feat(list_console_messages): include source-mapped location of messages

kopalg20:
I implemented async fetching of source-mapped locations for console messages and uncaught errors.

- list_console_messages remains fast and non-blocking.

- get_console_message returns detailed stack traces with resolved source-mapped top frames.

- Verified with 3000+ simulated errors; performance remains acceptable.

This addresses the request to show the source-mapped location for errors directly, without having to manually parse the stack trace.

Made the changes in src/formatters/ConsoleFormatter.ts

Screenshot 2026-02-14 012632
Screenshot 2026-02-14 012622

### DIFF
--- a/src/formatters/ConsoleFormatter.ts
+++ b/src/formatters/ConsoleFormatter.ts
@@ -159,8 +159,12 @@ export class ConsoleFormatter {
   }
 
   // The short format for a console message.
-  toString(): string {
-    return `msgid=${this.#id} [${this.#type}] ${this.#text} (${this.#argCount} args)`;
+ toString(): string {
+    const topFrame = this.#getTopFrameLocation();
+    const locationStr = topFrame
+      ? ` (${topFrame.url}:${topFrame.lineNumber}:${topFrame.columnNumber})`
+      : '';
+    return `msgid=${this.#id} [${this.#type}] ${this.#text}${locationStr} (${this.#argCount} args)`;
   }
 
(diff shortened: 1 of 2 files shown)


src/formatters/ConsoleFormatter.ts
```diff
@@ -314,6 +314,8 @@ export class ConsoleFormatter {
   }
 
   toJSONDetailed(): object {
+    const location = this.#getTopFrameLocation();
```
OrKoN: the feature is about adding the first line of the stack to the non-detailed variants (toString(), toJSON()) so that the first source line appears in the list of console message. toJSONDetailed is not used in list_console_messages.

kopalg20: Made required changes. Let me know any further improvements needed.
Also added test cases for the same. @OrKoN

tests/formatters/ConsoleFormatter.test.ts
```diff
@@ -673,3 +673,100 @@ describe('ConsoleFormatter', () => {
     });
   });
 });
+
+describe('ConsoleFormatter - Source Map & Heavy Error Tests', () => {
```
OrKoN, line 677: these tests do not use any source maps. Can we also embed them into existing tests that already verify console message details and cover the list expectations there too?

OrKoN:
Closing as stale.

PR #964, opened 2026-02-14
fix(input): include page list when click opens new tab

### LINKED ISSUE #367: return all pages when click web element
### Description of the bug

Image

### Reproduction

_No response_

### Expectation

_No response_

### MCP configuration

_No response_

### Node version

_No response_

### Chrome version

_No response_

### Coding agent version

_No response_

### Model version

_No response_

### Chat log

_No response_

### Operating system

None

### Extra checklist

SucRunBug:
## Summary

Fixes a UX gap where `click`/`click_at` can open a new tab but the tool response doesn't include updated page context, making it hard for agents to detect that navigation happened in a new page.

Closes #367.

## What changed

- Updated `/Users/yanmai/Downloads/MCP 测试/chrome-devtools-mcp/src/tools/input.ts`:
  - `click`: listens for `popup` during the action and calls `response.setIncludePages(true)` if a new tab/window is opened.
  - `click_at`: same behavior as `click`.
  - Ensures popup listeners are cleaned up in `finally` blocks.

- Updated `/Users/yanmai/Downloads/MCP 测试/chrome-devtools-mcp/tests/tools/input.test.ts`:
  - Added `includes pages if click opens a new tab`
  - Added `includes pages if click at coordinates opens a new tab`

## Why this approach

- Keeps existing behavior unchanged for normal clicks.
- Only includes page list when a new page is actually opened, minimizing response noise.

## Validation

- `npm run typecheck`
- `npx eslint src/tools/input.ts tests/tools/input.test.ts`
- Targeted tests for the new scenarios and related click flows

### DIFF
--- a/src/tools/input.ts
+++ b/src/tools/input.ts
@@ -52,6 +52,12 @@ export const click = defineTool({
   handler: async (request, response, context) => {
     const uid = request.params.uid;
     const handle = await context.getElementByUid(uid);
+    const page = context.getSelectedPage();
+    let popupOpened = false;
+    const popupListener = () => {
+      popupOpened = true;
+    };
+    page.on('popup', popupListener);
     try {
       await context.waitForEventsAfterAction(async () => {
         await handle.asLocator().click({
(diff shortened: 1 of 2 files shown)


OrKoN:
Thanks I think the change makes sense in general. Could you please sign the CLA?

src/tools/input.ts
```diff
@@ -52,6 +52,12 @@ export const click = defineTool({
...
     const uid = request.params.uid;
     const handle = await context.getElementByUid(uid);
+    const page = context.getSelectedPage();
+    let popupOpened = false;
+    const popupListener = () => {
+      popupOpened = true;
+    };
+    page.on('popup', popupListener);
     try {
       await context.waitForEventsAfterAction(async () => {
```
OrKoN, line 62: let's embed it into waitForEventsAfterAction since not only clicks can trigger a popup?

OrKoN:
Closing as stale.

PR #970, opened 2026-02-16
fix: apply pagination to network request output

AI-Reviewer-QS:
## Summary
- Network request pagination was cosmetic-only: the header showed "Showing 1-10 of 30" but all requests were rendered in both text and structured content
- The `format()` method computed pagination on raw requests but then iterated over all pre-built `NetworkFormatter` instances from `handle()`, ignoring the paginated subset
- Fixed by applying pagination to the formatter array directly, matching the pattern already used correctly by console message pagination
- Also removed a redundant re-fetch of raw network requests in `format()` since the formatters from `handle()` already contain the filtered data

## Test plan
- [x] Existing pagination tests updated to verify correct paginated output
- [x] `returns first page by default` - now only includes 10 of 30 requests
- [x] `returns subsequent page when pageIdx provided` - shows correct page slice
- [x] `handles invalid page number by showing first page` - shows first page subset
- [x] All other McpResponse tests continue to pass

### DIFF
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -589,36 +589,21 @@ Call ${handleDialog.name} to handle it before continuing.`);
     }
 
     if (this.#networkRequestsOptions?.include) {
-      let requests = context.getNetworkRequests(
-        this.#networkRequestsOptions?.includePreservedRequests,
-      );
-
-      // Apply resource type filtering if specified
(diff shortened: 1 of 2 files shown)


OrKoN:
@nroscino PTAL

nroscino:
This https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/980 addresses the issue without deleting unnecessary code. Closing.

PR #972, opened 2026-02-16
fix: restrict URL schemes in page navigation to http/https

AI-Reviewer-QS:
## Summary
- `new_page` and `navigate_page` tools pass user-provided URLs directly to `page.goto()` without validating the URL scheme
- This allows navigating to potentially dangerous schemes like `file://`, `chrome://`, or `data:` URLs which could expose local files or internal browser state
- Added URL scheme validation that restricts navigation to `http:`, `https:`, and `about:` schemes
- Unparseable URLs are passed through to let the browser handle them (it will reject invalid ones)

## Test plan
- [ ] Verify `http://` and `https://` URLs work normally in `new_page`
- [ ] Verify `http://` and `https://` URLs work normally in `navigate_page`
- [ ] Verify `file://` URLs are rejected with clear error message
- [ ] Verify `chrome://` URLs are rejected
- [ ] Verify `data:` URLs are rejected
- [ ] Verify `about:blank` still works
- [ ] Verify back/forward/reload navigation types are unaffected

### DIFF
--- a/src/tools/pages.ts
+++ b/src/tools/pages.ts
@@ -11,6 +11,24 @@ import {zod} from '../third_party/index.js';
 import {ToolCategory} from './categories.js';
 import {CLOSE_PAGE_ERROR, defineTool, timeoutSchema} from './ToolDefinition.js';
 
+const ALLOWED_URL_SCHEMES = ['http:', 'https:', 'about:'];
+
+function validateUrlScheme(url: string): void {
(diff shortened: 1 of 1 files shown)


OrKoN:
Thanks for the PR! Could you please file a feature request first? I think we can support all URL schemes allowed by Chrome for automation. cc @natorion @nroscino

PR #973, opened 2026-02-16
fix: redact sensitive headers in network request output

AI-Reviewer-QS:
## Summary
- Network request details expose all HTTP headers in plain text, including sensitive values like `Authorization` bearer tokens, `Cookie` values, and API keys
- This data is passed to the MCP client (typically an LLM) where it could be logged, cached, or inadvertently included in responses
- Added redaction for known sensitive headers (`Authorization`, `Cookie`, `Set-Cookie`, `Proxy-Authorization`) in both the text (`toStringDetailed()`) and structured JSON (`toJSONDetailed()`) output

## Example
Before:
```
- authorization:Bearer sk-secret-token-123
- cookie:session=abc123
```

After:
```
- authorization:<redacted>
- cookie:<redacted>
```

## Test plan
- [ ] Verify non-sensitive headers (Content-Type, Content-Length, etc.) display normally
- [ ] Verify `Authorization` header values are redacted in text output
- [ ] Verify `Cookie` header values are redacted in text output
- [ ] Verify sensitive headers are also redacted in JSON structured content
- [ ] Verify case-insensitive matching works (e.g., `AUTHORIZATION`, `Authorization`)

### DIFF
--- a/src/formatters/NetworkFormatter.ts
+++ b/src/formatters/NetworkFormatter.ts
@@ -10,6 +10,13 @@ import type {HTTPRequest, HTTPResponse} from '../third_party/index.js';
 
 const BODY_CONTEXT_SIZE_LIMIT = 10000;
 
+const SENSITIVE_HEADERS = new Set([
+  'authorization',
+  'cookie',
+  'set-cookie',
+  'proxy-authorization',
+]);
+
 export interface NetworkFormatterOptions {
   requestId?: number | string;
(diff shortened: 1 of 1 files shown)


OrKoN:
Could you please file a feature request first?  Currently, we return all data to facilitate the accurate debugging. If we were to implement it, we should re-use the implementation from chrome-devtools-frontend. cc @natorion @nroscino

Issue #975, opened 2026-02-16
tool fill can not work with google.com


russell0430:
### Description of the bug

I am build agent with chrome-devtool-mcp. but i when i start chat with ' search today's news on google'. Things wield happened.
Agent navigate to `google.com` with tool `new_page`,then call `take_snapshot` to get uid of input element ,then call tool `fill` with the uid and text 'today's news'.  
The response of fill always fail with   **Failed to interact with the element with uid 18_13. The element did not become interactive within the configured timeout.\nCause: Could not find option with text**. then agent implement the task with navigate to `google.com/search?q=today'new`

i look into the issue ，and find out when the input has attribute role equal to `combobox` ，the `fill`tool try to find option with the text of tool input ，which is not found in google search page ，and then throw error as below 。 [fill tool](https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/src/tools/input.ts#L186-L190) ,[select option](https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/src/tools/input.ts#L186-L190)

Is there any way to avoid such thing happen? 

### Reproduction

1. use tool `new_page` to google.com
2. use tool `take_snapshot`to get snapshot and uid of textarea(input) element
3. use tool `fill` to search something 
4. tool call fail 

### Expectation

Expect to search successfully in google.com with mcp

### Chrome DevTools MCP version

latest

### Operating system

None

### Extra checklist

- [ ] I want to provide a PR to fix this bug

OrKoN:
Thanks for reporting! able to reproduce.

PR #979, opened 2026-02-17
fix: check that combobox is actually a select element before filling out options

OrKoN:
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/975

### DIFF
--- a/src/tools/input.ts
+++ b/src/tools/input.ts
@@ -177,6 +177,10 @@ async function selectOption(
   }
 }
 
+function hasOptionChildren(aXNode: TextSnapshotNode) {
+  return aXNode.children.some(child => child.role === 'option');
+}
+
 async function fillFormElement(
   uid: string,
   value: string,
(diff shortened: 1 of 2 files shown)


src/tools/input.ts
```diff
@@ -185,7 +185,11 @@ async function fillFormElement(
   const handle = await context.getElementByUid(uid);
   try {
     const aXNode = context.getAXNodeByUid(uid);
-    if (aXNode && aXNode.role === 'combobox') {
+    if (
+      aXNode &&
+      aXNode.role === 'combobox' &&
+      (await handle.evaluate(el => el.tagName.toLowerCase() === 'select'))
```
Lightning00Blade: Is it possible to check if the child nodes are `options` (from A11Y tree), that way we don't need to do this async.

OrKoN: Good point, let me check, although I wonder if an expanded non-select combobox would still have options in the a11y tree.

Lightning00Blade (approved):

PR #985, opened 2026-02-18
refactor(network): de-duplicate String and JSON formatters

OrKoN:
- string formatters should use the same data as the JSON formatters
- removed an extra space in network status

### DIFF
--- a/src/formatters/NetworkFormatter.ts
+++ b/src/formatters/NetworkFormatter.ts
@@ -23,6 +23,25 @@ export interface NetworkFormatterOptions {
   ) => Promise<{filename: string}>;
 }
 
+interface NetworkRequestConcise {
+  requestId?: number | string;
+  method: string;
+  url: string;
+  status: string;
+  selectedInDevToolsUI?: boolean;
+}
+
+interface NetworkRequestDetailed extends NetworkRequestConcise {
+  requestHeaders: Record<string, string>;
+  requestBody?: string;
+  requestBodyFilePath?: string;
+  responseHeaders?: Record<string, string>;
+  responseBody?: string;
+  responseBodyFilePath?: string;
+  failure?: string;
+  redirectChain?: NetworkRequestConcise[];
+}
+
 export class NetworkFormatter {
   #request: HTTPRequest;
   #options: NetworkFormatterOptions;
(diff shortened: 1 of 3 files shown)


src/formatters/NetworkFormatter.ts
```diff
@@ -273,3 +226,67 @@ function getSizeLimitedString(text: string, sizeLimit: number) {
   }
   return text;
 }
+
+function toString(data: NetworkRequestConcise): string {
```
Lightning00Blade: Please either provide better name here or move to static method.

src/formatters/NetworkFormatter.ts
```diff
@@ -273,3 +226,67 @@ function getSizeLimitedString(text: string, sizeLimit: number) {
...
+  // TODO truncate the URL
+  return `reqid=${data.requestId} ${data.method} ${data.url} ${data.status}${data.selectedInDevToolsUI ? ` [selected in the DevTools Network panel]` : ''}`;
+}
+
+function toStringDetailed(data: NetworkRequestDetailed): string {
```
Lightning00Blade: Ditto.

src/formatters/NetworkFormatter.ts
```diff
@@ -273,3 +226,67 @@ function getSizeLimitedString(text: string, sizeLimit: number) {
...
+  const getFormattedHeaderValue = (
```
Lightning00Blade: Nit: This looks like a pure function, it can live outside this one.

src/formatters/NetworkFormatter.ts
OrKoN: I am not sure how making it a static method makes the name any better :)

src/formatters/NetworkFormatter.ts
OrKoN: I don't want anything else using it?

OrKoN: anyway, moved it outside

src/formatters/NetworkFormatter.ts
Lightning00Blade: It would be `NetworkFormatter.toString(request)` so the call would make the explanation.

src/formatters/NetworkFormatter.ts
Lightning00Blade: Module isolation should be good enough.

src/formatters/NetworkFormatter.ts
OrKoN: we already have NetworkFormatter.toString. new methods to not work on the request but on the JSON objects (described by new interfaces).

PR #989, opened 2026-02-18
test: fix missing sinon.restore

OrKoN:


tests/PageCollector.test.ts
```diff
@@ -302,6 +302,10 @@ describe('ConsoleCollector', () => {
     };
   });
 
+  afterEach(() => {
```
Lightning00Blade, line 305: Can we always put it next to the first describe?

tests/tools/pages.test.ts
```diff
@@ -122,6 +122,10 @@ describe('pages', () => {
     });
   });
   describe('navigate_page', () => {
+    afterEach(() => {
```
Lightning00Blade: Same here.

tests/PageCollector.test.ts
OrKoN: no, because there are multiple describes here. The first one does not use sinon.

tests/tools/pages.test.ts
OrKoN: here we can

PR #993, opened 2026-02-19
feat: add skill to debug and optimize LCP

matthiasrohmer:
This is a draft for new skill to debug and optimize LCP scores, going hand in hand with https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/831.

Base for this skill is key authoritative learning material from web.dev, about debugging and optimizing LCP, mainly:

- http://web.dev/articles/lcp
- http://web.dev/articles/optimize-lcp
- https://web.dev/articles/top-cwv

I gave the content of those articles together with the [tool reference](https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/docs/tool-reference.md), the README and the chrome-devtools core skill to both Claude Opus 4.6 and Gemini 3 Pro and their respective skill creator skills. The final skill in this PR is a merge of multiple runs with both models.

The snippets to run with `evaluate_script` is something only Opus came up with, but I think it's quite nice, though our own `performance_*` tools might give the same insight but likely with more overhead (meaning tokens).

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,21 @@ Add the following config to your MCP client:
 > [!NOTE]  
 > Using `chrome-devtools-mcp@latest` ensures that your MCP client will always use the latest version of the Chrome DevTools MCP server.
 
+If you are intersted in doing only basic browser tasks, use the `--slim` mode:
+
+```json
+{
+  "mcpServers": {
+    "chrome-devtools": {
+      "command": "npx",
+      "args": ["-y", "chrome-devtools-mcp@latest", "--slim", "--headless"]
+    }
+  }
+}
+```
+
+See [Slim tool reference](./docs/slim-tool-reference.md).
+
 ### MCP Client configuration
 
 <details>
@@ -531,6 +546,10 @@ The Chrome DevTools MCP server supports the following configuration option:
   Set to false to opt-out of usage statistics collection. Google collects usage data to improve the tool, handled under the Google Privacy Policy (https://policies.google.com/privacy). This is independent from Chrome browser metrics. Disabled if CHROME_DEVTOOLS_MCP_NO_USAGE_STATISTICS or CI env variables are set.
   - **Type:** boolean
   - **Default:** `true`
+
+- **`--slim`**
+  Exposes a "slim" set of 3 tools covering navigation, script execution and screenshots only. Useful for basic browser tasks.
+  - **Type:** boolean
 
 <!-- END AUTO GENERATED OPTIONS -->
 
--- a/docs/slim-tool-reference.md
+++ b/docs/slim-tool-reference.md
@@ -1,0 +1,41 @@
+<!-- AUTO GENERATED DO NOT EDIT - run 'npm run docs' to update-->
+
+# Chrome DevTools MCP Slim Tool Reference (~368 cl100k_base tokens)
+
+- **[Navigation automation](#navigation-automation)** (1 tools)
+  - [`navigate`](#navigate)
+- **[Debugging](#debugging)** (2 tools)
+  - [`evaluate`](#evaluate)
+  - [`screenshot`](#screenshot)
+
+## Navigation automation
+
+### `navigate`
+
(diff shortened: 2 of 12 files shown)


skills/debug-optimize-lcp/references/elements-and-size.md
```diff
@@ -0,0 +1,27 @@
+# Elements and Size for LCP
```
OrKoN, line 1: References are only sources for the skill and are not actually used? can we move them out of the skills folder?

Can you also estimate the skill size with https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/package.json#L26 ?

skills/debug-optimize-lcp/SKILL.md
```diff
@@ -0,0 +1,112 @@
...
+- **LCPBreakdown** — Shows the four LCP subparts with timing for each.
+- **DocumentLatency** — Server response time issues affecting TTFB.
+- **RenderBlocking** — Resources blocking the LCP element from rendering.
+- **LCPDiscovery** — Whether the LCP resource was discoverable early.
+
+Call `performance_analyze_insight` with the insight name and the insight set ID from the trace results.
+
+### Step 3: Identify the LCP Element
+
+Use `evaluate_script` with the **"Identify LCP Element" snippet** found in [references/lcp-snippets.md](references/lcp-snippets.md) to reveal the LCP element's tag, resource URL, and raw timing data.
```
OrKoN, line 57: I am not sure skills support relative references? in which clients does it work? I would suggest hosting reference somewhere and linking via an absolute URL.

skills/debug-optimize-lcp/references/elements-and-size.md
matthiasrohmer: Depends on the model/agent I think, but both Claude (https://code.claude.com/docs/en/skills#add-supporting-files) and Gemini (https://geminicli.com/docs/cli/skills/#creating-your-own-skills) add the full skill file to the context window and their respective read_file tools have access to the referenced materials.

So if the model thinks the doc in references could be useful it would likely read them. So I think for example extracting the snippets out in to reference files and not proactively have them in the skill context is quite reasonable to reduce initial token size. Do you think differently?

Will measure the skill size and get back!

skills/debug-optimize-lcp/SKILL.md
matthiasrohmer: See comment above, both Claude and Gemini CLI claim to support this:

- https://code.claude.com/docs/en/skills#add-supporting-files
- https://geminicli.com/docs/cli/skills/#creating-your-own-skills

skills/debug-optimize-lcp/references/elements-and-size.md
OrKoN: Ack, in this case, sg!

matthiasrohmer: @natorion, thoughts? We could also merge into one file, which would make it easer to reason about (especially for a human).

OrKoN: I think we should not merge in one file, if the clients are actually expected to be able to follow local references, I think the current selection is good. if we could optimize the token size for the main skill file I think it could be useful

natorion: +1

matthiasrohmer:
Sorry for the delay, formatted now. Here's the token count:

```
Input: skills/debug-optimize-lcp/SKILL.md
Tokens: 1603

Input: skills/debug-optimize-lcp/references/elements-and-size.md
Tokens: 298

Input: skills/debug-optimize-lcp/references/lcp-breakdown.md
Tokens: 504

Input: skills/debug-optimize-lcp/references/lcp-snippets.md
Tokens: 662

Input: skills/debug-optimize-lcp/references/optimization-strategies.md
Tokens: 491
```

Issue #996, opened 2026-02-19
v0.17.2 npm package is missing build directory — cannot start


9c5s:
## Description of the bug

v0.17.2 published to npm is missing the entire `build/` directory. The package contains only 3 files (`LICENSE`, `README.md`, `package.json`), so the entry point `build/src/index.js` cannot be resolved. This is the same class of issue as #324.

## Reproduction

```
$ npm pack chrome-devtools-mcp@0.17.2 --dry-run
npm notice 📦  chrome-devtools-mcp@0.17.2
npm notice Tarball Contents
npm notice 11.4kB LICENSE
npm notice 26.9kB README.md
npm notice 3.3kB  package.json
npm notice total files: 3
```

vs working v0.17.1:

```
$ npm pack chrome-devtools-mcp@0.17.1 --dry-run
npm notice 📦  chrome-devtools-mcp@0.17.1
npm notice Tarball Contents
npm notice 11.4kB  LICENSE
npm notice 26.8kB  README.md
npm notice 6.4kB   build/src/browser.js
npm notice 11.3kB  build/src/cli.js
npm notice 2.6kB   build/src/DevToolsConnectionAdapter.js
npm notice 13.2kB  build/src/DevtoolsUtils.js
...
```

Running v0.17.2 produces:

```
Error: Cannot find module '.../chrome-devtools-mcp/build/src/index.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1448:15)
```

## Expectation

The `build/src/` directory should be included in the published npm package, as it is in v0.17.1 and all prior working versions.

## MCP configuration

```json
{
  "chrome-devtools": {
    "type": "stdio",
    "command": "npx",
    "args": ["-y", "chrome-devtools-mcp@latest"]
  }
}
```

## Chrome DevTools MCP version

0.17.2

## Node version

v24.3.0

## Operating system

Windows

## Workaround

Pin to v0.17.1:

```
npx -y chrome-devtools-mcp@0.17.1
```

OrKoN:
Thanks for reporting and sorry for the trouble. The fix will be out shortly.

OrKoN:
0.17.3 is published

9c5s:
Thank you for your prompt response. I rely on this useful tool all the time.

Issue #995, opened 2026-02-19
v0.17.2 published without build directory


scottarver:
### Description of the bug

  The npm package `chrome-devtools-mcp@0.17.2` (published 2026-02-19) is missing the `build/`
  directory. The package only contains `LICENSE`, `README.md`, and `package.json`.

  The `bin` field in `package.json` points to `build/src/index.js` which doesn't exist in the
  tarball, so `npx chrome-devtools-mcp` fails with:

  sh: 1: chrome-devtools-mcp: not found

<img width="864" height="525" alt="Image" src="https://github.com/user-attachments/assets/406ddd70-be75-4320-9aa2-25d74c108302" />

### Reproduction

 ```bash
  npm pack chrome-devtools-mcp@0.17.2 --dry-run
  # Only shows: LICENSE, README.md, package.json

  npm pack chrome-devtools-mcp@0.16.0 --dry-run
  # Shows full build/ directory with all .js files

### Expectation

<img width="910" height="595" alt="Image" src="https://github.com/user-attachments/assets/120c307b-1ea4-47cf-9aba-a499a424b5e5" />

### MCP configuration

_No response_

### Chrome DevTools MCP version

v0.17.2

### Chrome version

_No response_

### Coding agent version

_No response_

### Model version

_No response_

### Chat log

_No response_

### Node version

_No response_

### Operating system

Windows Subsystem for Linux (WSL)

### Extra checklist

- [ ] I want to provide a PR to fix this bug

OrKoN:
Thanks for reporting and sorry for the trouble. The fix will be out shortly.

OrKoN:
0.17.3 is published

PR #999, opened 2026-02-19
chore: verify npm package content before publishing

OrKoN:
Refs: https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/996 https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/995


zyzyzyryxy:
What about checking if `package.json`'s `bin` path exists?

OrKoN:
> What about checking if `package.json`'s `bin` path exists?

I do not think it is worth it for multiple reasons. But happy to review a PR with that :)

PR #1002, opened 2026-02-20
feat: add a new skill for accessibility debugging and auditing with Chrome DevTools MCP.

natorion:


skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,157 @@
...
+        }));
+    };
+    ```
+4.  Check images for `alt` text.
+
+### 4. Focus & Keyboard Navigation
+
+Testing "keyboard traps" and proper focus management without visual feedback relies on precise scripting.
+
+1.  Use `evaluate_script` to find the currently focused element:
```
OrKoN: I think this might be redundant. The snapshot tells which element has focus, so just sending press_key and automatically getting snapshot back should be sufficient.

skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,157 @@
...
+```javascript
+el => {
+  const rect = el.getBoundingClientRect();
+  return {width: rect.width, height: rect.height};
+};
+```
+
+_Pass the element's `uid` from the snapshot as an argument to the tool._
+
+### 6. Color Contrast
```
OrKoN, line 86: should we enable contrast issues in audits instead?

OrKoN: crbug.com/40735430

natorion: you mean in issues?

OrKoN: issues or Lighthouse that is coming

skills/a11y-debugging/SKILL.md
natorion: Ack, done.

skills/a11y-debugging/SKILL.md
natorion: Hmm, so the first thing should be to check issues and lighthouse if available, then run the script?

OrKoN: I think we should not recommend running complex scripts as there is no way for us to verify that the script actually works well and correctly.

OrKoN: maybe we should add scripts as individual files and then embded them into scripts so that we can run tests on scripts?

OrKoN: I also wonder if using axe-core directly is better?

natorion: I did a change, which shows the script but also references axe core.

OrKoN (approved):

natorion:
The code examples are hit or miss. Let;s see if I can make them better.

natorion:
I now tuned the js snippets a bit, so that gemini 2.5 pro is able to utilize them for effect. I think this looks good for now.

skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,164 @@
+---
+name: a11y-debugging
+description: Uses Chrome DevTools MCP for accessibility (a11y) debugging and auditing based on web.dev guidelines. Use when testing semantic HTML, ARIA labels, focus states, keyboard navigation, tap targets, and color contrast.
+---
+
+## Core Concepts
+
+**Accessibility Tree vs DOM**: Visually hiding an element (e.g., `CSS opacity: 0`) behaves differently for screen readers than `display: none` or `aria-hidden="true"`. The `take_snapshot` tool returns the accessibility tree of the page, which represents what assistive technologies "see", making it the most reliable source of truth for semantic structure.
+
+**Reading web.dev documentation**: If you need to research specific accessibility guidelines (like `https://web.dev/articles/accessible-tap-targets`), you can append `.md.txt` to the URL (e.g., `https://web.dev/articles/accessible-tap-targets.md.txt`) to fetch the clean, raw markdown version. This is much easier to read using the `read_url_content` tool!
```
matthiasrohmer, line 10: Is `read_url_content` agent/model agnostic? Or will they be able to map that to whatever their own implemention is called?

skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Workflow Patterns
+
+### 1. Browser Issues & Audits
+
+Chrome automatically checks for common accessibility problems. Use `list_console_messages` to check for these native audits first:
+
+- `types`: `["issue"]`
+- `includePreservedMessages`: `true` (to catch issues that occurred during page load)
```
matthiasrohmer, line 19: If we instruct to use tools with such high specificity, how do we ensure they stay aligned to potential future API updates?

skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,164 @@
...
+1.  Navigate to the page.
+2.  Use `take_snapshot` to capture the accessibility tree.
+3.  **Check Heading Levels**: Ensure heading levels (`h1`, `h2`, `h3`, etc.) are logical and do not skip levels. The snapshot will include heading roles.
+4.  **Content Reordering**: Verify that the DOM order (which drives the accessibility tree) matches the visual reading order. Use `take_screenshot` to inspect the visual layout and compare it against the snapshot structure to catch CSS floats or absolute positioning that jumbles the logical flow.
+
+### 3. Labels, Forms & Text Alternatives
+
+1.  Locate buttons, inputs, and images in the `take_snapshot` output.
+2.  Ensure interactive elements have an accessible name (e.g., a button should not just say `""` if it only contains an icon).
+3.  **Orphaned Inputs**: Verify that all form inputs have associated labels. Use `evaluate_script` to check for inputs missing `id` (for `label[for]`) or `aria-label`:
```
matthiasrohmer, line 36: I think there's a chance to shorten here. The model will be able to infer `to check for inputs missing `id` (for `label[for]`) or `aria-label`` from the following code snippet anyway.

skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,164 @@
...
+        .map(i => ({
+          tag: i.tagName,
+          id: i.id,
+          name: i.name,
+          placeholder: i.placeholder,
+        }));
+    };
+    ```
+
+````
```
matthiasrohmer, line 59: Orphaned?

skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,164 @@
...
+
+4.  Check images for `alt` text.
+
+### 4. Focus & Keyboard Navigation
```
matthiasrohmer, line 63: Orphaned too?

skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,164 @@
...
+
+Testing "keyboard traps" and proper focus management without visual feedback relies on tracking the focused element.
+
+1.  Use the `press_key` tool with `"Tab"` or `"Shift+Tab"` to move focus.
+2.  Use `take_snapshot` to capture the updated accessibility tree.
+3.  Locate the element marked as focused in the snapshot to verify focus moved to the expected interactive element.
+4.  If a modal opens, focus must move into the modal and "trap" within it until closed.
```
matthiasrohmer, line 70: I wonder if this needs an additional instruction like "Look for buttons and other interactive elements and try to interact with them." - Because how else would a modal open in an agentic loop?

skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,164 @@
...
+
+### 5. Tap Targets and Visuals
+
+According to web.dev, tap targets should be at least 48x48 pixels with sufficient spacing. Since the accessibility tree doesn't show sizes, use `evaluate_script`:
```
matthiasrohmer, line 74: I think I would remove "According to web.dev". Tiny risk an agent would just call web.dev, which doesn't help them as it's just the landing page.

skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,164 @@
...
+
+```javascript
+// Usage in console: copy, paste, and call with element: fn(element)
```
matthiasrohmer, line 77: Comment can be removed.

skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,164 @@
...
+el => {
+ const rect = el.getBoundingClientRect();
+ return {width: rect.width, height: rect.height};
+};
+````
+
+_Pass the element's `uid` from the snapshot as an argument to the tool._
```
matthiasrohmer, line 84: To which tool?

skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,164 @@
...
+To verify color contrast ratios, start by checking for native accessibility issues:
+
+1.  Call `list_console_messages` with `types: ["issue"]`.
+2.  Look for "Low Contrast" issues in the output.
+
+If native audits do not report issues (which may happen in some headless environments) or if you need to check a specific element manually, you can use the following script as a fallback approximation.
+
+**Note**: This script uses a simplified algorithm and may not account for transparency, gradients, or background images. For production-grade auditing, consider injecting `axe-core`.
+
+```javascript
```
matthiasrohmer, line 97: Nit: we could shorten this to `js` for every snippet, for token optimization.

skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,164 @@
...
+el => {
```
matthiasrohmer, line 98: Not actionable for this PR, but is there an opportunity to expose this internal DevTools functionality through CDP? @OrKoN?

skills/a11y-debugging/SKILL.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+_Pass the element's `uid` to test the contrast against WCAG AA (4.5:1 for normal text, 3:1 for large text)._
+
+### 7. Global Page Checks
+
+Verify document-level accessibility settings often missed in component testing:
+
+```javascript
+(() => {
```
matthiasrohmer, line 137: Shouldn't this use `evaluate_script` instead of printing to the console? Because the later would require the list console messages tool.

skills/a11y-debugging/SKILL.md
natorion: As the skills are in the same directory as the tools, I would assume they will change in lockstep.

PR #1007, opened 2026-02-20
refactor: types for JSON output of IssueFormatter

OrKoN:
- makes the id required in types
- requires toString(), toStringDetailed() to use the JSON representation.


src/formatters/IssueFormatter.ts
```diff
@@ -20,6 +20,22 @@ export interface AffectedResource {
   request?: string | number;
 }
 
+interface IssueConcise {
```
nattallius, line 23: To my opinion **Concise** is a niche word, so i would make it more simple,
`IssueShort, IssueSimple`

OrKoN: Thanks, I think concise is a regular word but I will consider changing it. For now it is used in multiple places already so keeping it.

PR #1011, opened 2026-02-21
feat: support any-match text arrays in wait_for

### LINKED ISSUE #916: `wait_for`: ability to wait for multiple optional texts (any of them might match)

I see chrome-devtools-mcp use wait_for quite a lot, and often in good ways. However, one area it struggles is for long-running processes that might either succeed or fail; it will then wait_for "Complete" but the actual message might be "Error". I've seen this drive the wait time up (e.g. devtools MCP will assume it's not done and wait another 300 seconds for a process it thinks might be long-running, but has failed after a second already).

Allow wait_for to wait for any term of a provided list, so that devtools mcp can e.g. wait for "Complete" or "Failed".

Alternative is very long wait times until it timeouts, and then devtools MCP figures out "oh, actually it failed" after 300s or so.

### Additional context

_No response_

ashutoshpw:
## Summary

Enhances wait_for to support waiting on multiple possible texts and resolve when any one appears.

This addresses long-running flows that can end in different UI outcomes (for example, "Complete" or "Error"), avoiding unnecessary 300s waits when only one expected string is provided.

Closes #916.

### wait_for

Waits for text on the selected page, now with any-match support.

| Parameter | Type | Required | Description |
|-----------|------|----------|-------------|
| text | string \| string[] | yes | A single text or a non-empty list of texts. Resolves when any value appears. |
| timeout | integer | no | Maximum wait in ms (0 keeps default behavior). |

---

## Design / Implementation

- Kept backward compatibility: existing single-string text calls continue to work unchanged.
- Added schema support for string | string[] with non-empty array validation.
- Updated context API to accept string | string[].
- Matching logic now normalizes to an array and races all candidates across all frames using both:
    - aria/<text>
    - text/<text>
- Added clearer response output for array input:
    - Element matching one of ["Complete","Error"] found.
- Updated generated tool docs for the new wait_for contract.
- Improved docs generation to render ZodUnion types (so union params are documented correctly, not as unknown).
  
---

## Tests

Added/updated coverage for:

- Schema acceptance of:
    - single string
    - non-empty string array
    - rejection of empty array
- Any-match array success case
- Any-match array when matching text appears later (async/delayed content)
- Existing wait_for behavior remains covered for single-text usage

Executed relevant test suites:

- tests/tools/snapshot.test.ts
- tests/McpContext.test.ts
- tests/index.test.ts 

Confirmation after the change applied:
https://opncd.ai/share/8m6I4r4a

### DIFF
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -1,6 +1,6 @@
 <!-- AUTO GENERATED DO NOT EDIT - run 'npm run docs' to update-->
 
-# Chrome DevTools MCP Tool Reference (~6750 cl100k_base tokens)
+# Chrome DevTools MCP Tool Reference (~6916 cl100k_base tokens)
 
 - **[Input automation](#input-automation)** (8 tools)
   - [`click`](#click)


src/tools/snapshot.ts
```diff
@@ -43,13 +43,17 @@ in the DevTools Elements panel (if any).`,
 
 export const waitFor = defineTool({
   name: 'wait_for',
-  description: `Wait for the specified text to appear on the selected page.`,
+  description: `Wait for the specified text to appear on the selected page. You can provide a single text value or a list of texts; the tool resolves when any text appears.`,
```
OrKoN: This doc is probably redundant as the behavior is documented on the text parameter.

src/tools/snapshot.ts
```diff
@@ -43,13 +43,17 @@ in the DevTools Elements panel (if any).`,
...
   annotations: {
     category: ToolCategory.NAVIGATION,
     readOnlyHint: true,
   },
   schema: {
-    text: zod.string().describe('Text to appear on the page'),
+    text: zod
+      .union([zod.string(), zod.array(zod.string()).nonempty()])
```
OrKoN: cc @Lightning00Blade we need to test if this is supported by major MCP clients

tests/tools/snapshot.test.ts
```diff
@@ -20,6 +20,15 @@ describe('snapshot', () => {
     });
   });
   describe('browser_wait_for', () => {
+    it('accepts string or non-empty array text input', () => {
```
OrKoN: we do not need to test schema parsing, please remove this test

OrKoN:
Thanks for the PR! Mostly looks good, left a couple of comments.

src/tools/snapshot.ts
ashutoshpw: agreed, removed now.

tests/tools/snapshot.test.ts
ashutoshpw: removed now.

src/tools/snapshot.ts
Lightning00Blade: Rather than having this complex type, I think changing this to `string[]` always.

src/tools/snapshot.ts
```diff
@@ -58,9 +62,15 @@ export const waitFor = defineTool({
...
-    response.appendResponseLine(
-      `Element with text "${request.params.text}" found.`,
-    );
+    if (Array.isArray(request.params.text)) {
+      response.appendResponseLine(
+        `Element matching one of ${JSON.stringify(request.params.text)} found.`,
+      );
+    } else {
+      response.appendResponseLine(
+        `Element with text "${request.params.text}" found.`,
```
Lightning00Blade: I think it would be nice to know which one got resolve, but let's do that as follow up.

src/McpContext.ts
```diff
@@ -825,15 +825,25 @@ export class McpContext implements Context {
...
-  waitForTextOnPage(text: string, timeout?: number): Promise<Element> {
+  waitForTextOnPage(
+    text: string | string[],
+    timeout?: number,
+  ): Promise<Element> {
     const page = this.getSelectedPage();
     const frames = page.frames();
+    const texts = Array.isArray(text) ? text : [text];
+
+    if (texts.length === 0) {
```
Lightning00Blade: This check will now not be needed as we can rely on Zod to verify.
for type we can update to `[string,...string[]]`, to make TypeScript happy.

src/tools/snapshot.ts
ashutoshpw: Kept this as `([zod.string(), zod.array(zod.string()).nonempty()])` to keep it backward compatible. Should I just set it to string[] ?

OrKoN: we do not need to keep it backward compatible because tool calls are generated by LLMs based on the latest JSON schemas

src/tools/snapshot.ts
```diff
@@ -49,7 +49,12 @@ export const waitFor = defineTool({
     readOnlyHint: true,
   },
   schema: {
-    text: zod.string().describe('Text to appear on the page'),
+    text: zod
+      .tuple([zod.string()])
```
Lightning00Blade: This looks incorrect. This should be `z.array(z.string()).min(1)` as a tuple implies fixed number of items.
This may require typecasing the `request.params.text,` but that is OK to do as we can rely on Zod to do the runtime verification.

scripts/generate-docs.ts
```diff
@@ -292,6 +293,22 @@ function getZodTypeInfo(schema: ZodSchema): TypeInfo {
         result.items = getZodTypeInfo(def.type);
       }
       break;
+    case 'ZodUnion':
```
Lightning00Blade: Given that we don't use an Union now let's remove it and we can handle the case more in depth once we have a use of it.

src/tools/snapshot.ts
```diff
@@ -59,7 +64,7 @@ export const waitFor = defineTool({
     );
 
     response.appendResponseLine(
-      `Element with text "${request.params.text}" found.`,
+      `Element matching one of ${JSON.stringify(request.params.text)} found.`,
```
Lightning00Blade, line 67: @OrKoN Do you think we need to explicitly say X found, given that we don't tell which one I think we can just provide a simple text here. Can I leave this part to be reviewed by you?

docs/tool-reference.md
```diff
@@ -194,7 +194,7 @@
 
 **Parameters:**
 
-- **text** (string) **(required)**: Text to appear on the page
+- **text** (unknown) **(required)**: Non-empty list of texts. Resolves when any value appears on the page.
```
Lightning00Blade: I think this will be correct once we set the correct Zod definitions.

src/tools/snapshot.ts
OrKoN: I think we can provide a simple text for now but the current version is okay too.

Lightning00Blade: ACK, marking as resolved for the PR.

docs/tool-reference.md
ashutoshpw: Fixed in https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1011/changes/9b3020c40f24be028fcf65a98295015c1c992205

scripts/generate-docs.ts
ashutoshpw: Removed in https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1011/changes/9b3020c40f24be028fcf65a98295015c1c992205

src/tools/snapshot.ts
ashutoshpw: agreed, changed it to `z.array(z.string()).min(1)`

ashutoshpw:
@OrKoN Conflicts resolved (rebased with upstream main and regenerated docs)

PR #1017, opened 2026-02-23
docs: Update troubleshooting for MCP server connection errors

EzioMercer:
Added solutions for MCP server connection issues on Windows 10, including using cmd and absolute path for npx.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -103,6 +103,7 @@ Possible workarounds include:
 ### Windows 10: Error during discovery for MCP server 'chrome-devtools': MCP error -32000: Connection closed
 
 - **Solution 1** Call using `cmd` (For more info https://github.com/modelcontextprotocol/servers/issues/1082#issuecomment-2791786310)
+
   ```json
   "mcpServers": {
       "chrome-devtools": {
@@ -110,7 +111,8 @@ Possible workarounds include:
         "args": ["/c", "npx", "-y", "chrome-devtools-mcp@latest"]
       }
     }
-  ``` 
+  ```
+
   > **The Key Change:** On Windows, running a Node.js package via `npx` often requires the `cmd /c` prefix to be executed correctly from within another process like VSCode's extension host. Therefore, `"command": "npx"` was replaced with `"command": "cmd"`, and the actual `npx` command was moved into the `"args"` array, preceded by `"/c"`. This fix allows Windows to interpret the command correctly and launch the server.
 
 - **Solution 2** Instead of another layer of shell you can write the absolute path to `npx`:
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -98,3 +98,27 @@ Possible workarounds include:
...
+- **Solution 2** Instead of another layer of shell you can write the absolute path to `npx`:
+  ```json
+  "mcpServers": {
+      "chrome-devtools": {
+        "command": "C:\\nvm4w\\nodejs\\npx.ps1",
+        "args": ["-y", "chrome-devtools-mcp@latest"]
+      }
+    }
+  ```
+  Note: The path above is an example. You must adjust it to match the actual location of `npx` on your machine. Depending on your setup, the file extension might be `.cmd`, `.bat`, or `.exe` rather than `.ps1`. Also, ensure you use double backslashes (`\\`) as path delimiters, as required by the JSON format.
```
matthiasrohmer: Can you move this note before the code snippet please? 🙂 I would claim 99% of people copy it blindly and never read the note otherwise.

I have no Windows machine to cross check, but would "Use `where.exe npx` to understand where your NPM installation is located` work?

docs/troubleshooting.md
```diff
@@ -98,3 +98,27 @@ Possible workarounds include:
      `npx chrome-devtools-mcp --browser-url http://127.0.0.1:9222`
 
 - **Use Powershell or Git Bash** instead of WSL.
+
+### Windows 10: Error during discovery for MCP server 'chrome-devtools': MCP error -32000: Connection closed
```
matthiasrohmer, line 103: I wonder if we should add another note here, like:

>Note: The solutions provided below apply to a native Windows setup, they can't fix issues related to the Windows Subsystem for Linux (WSL)

matthiasrohmer (approved):
Thanks for taking the time to improve this guidance, much appreciated!

I left one small suggestion on order, and if you have time can you kindly check me on the `where.exe` bit?

docs/troubleshooting.md
EzioMercer: As you wish :) I will move it
Currently I'm writing from mobile phone, so can't check `where.exe`. I will check it ASAP

docs/troubleshooting.md
EzioMercer: TBH I think it is obvious that we are talking about Windows itself but if you really want I can change the title from `Windows 10: ...` to `Windows 10 (not WSL)`: ...`

docs/troubleshooting.md
EzioMercer: @matthiasrohmer I checked the `where.exe` and here is the output:

```
C:\nvm4w\nodejs\npx
C:\nvm4w\nodejs\npx.cmd
C:\nvm4w\nodejs\npx.ps1
C:\Users\Username\AppData\Roaming\npm\npx
C:\Users\Username\AppData\Roaming\npm\npx.cmd
C:\Users\Username\AppData\Roaming\npm\npx.ps1
```

I think this will confuse people even more

PR #1018, opened 2026-02-23
Feat/isolated context page routing

passtas:


passtas:
I'll create new one or reopen once #991 is merged, as this one depends on it

PR #1020, opened 2026-02-23
chore: implement a daemon process

OrKoN:
Implements a daemon process needed for the CLI. The daemon has commands to start/stop itself as well as an ability to forward data to the managed MCP server. Local sockets or named pipes are used depending on the platform.

The .sock and .pid files are placed under XDG_RUNTIME_DIR/chrome-devtools-mcp or /tmp/chrome-devtools-mcp for short socket paths.

It uses Puppeteer's PipeTransport to avoid re-implementing \0 terminated messages.

Note the code is not used anywhere at the moment.

### DIFF
--- a/src/daemon/daemon.ts
+++ b/src/daemon/daemon.ts
@@ -1,0 +1,214 @@
+#!/usr/bin/env node
+
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import fs from 'node:fs/promises';
+import {createServer, type Server} from 'node:net';
+import process from 'node:process';
+
+import {Client} from '@modelcontextprotocol/sdk/client/index.js';
+import {StdioClientTransport} from '@modelcontextprotocol/sdk/client/stdio.js';
+
+import {logger} from '../logger.js';
+import {PipeTransport} from '../third_party/index.js';
+
+import {
+  getSocketPath,
+  handlePidFile,
+  INDEX_SCRIPT_PATH,
+  IS_WINDOWS,
+} from './utils.js';
+
+const pidFile = handlePidFile();
+const socketPath = getSocketPath();
+
+let mcpClient: Client | null = null;
+let mcpTransport: StdioClientTransport | null = null;
+let server: Server | null = null;
+
+async function setupMCPClient() {
+  console.log('Setting up MCP client connection...');
+
+  const args = process.argv.slice(2);
+  // Create stdio transport for chrome-devtools-mcp
+  mcpTransport = new StdioClientTransport({
+    command: process.execPath,
+    args: [INDEX_SCRIPT_PATH, ...args],
+    env: process.env as Record<string, string>,
+  });
+  mcpClient = new Client(
+    {
+      name: 'chrome-devtools-cli-daemon',
+      // TODO: handle client version (optional).
+      version: '0.1.0',
+    },
+    {
+      capabilities: {},
+    },
+  );
+  await mcpClient.connect(mcpTransport);
+
+  console.log('MCP client connected');
+}
+
+interface McpContent {
+  type: string;
+  text?: string;
+}
+
+interface McpResult {
+  content?: McpContent[] | string;
+  text?: string;
+}
+
+type DaemonMessage =
+  | {
+      method: 'stop';
+    }
+  | {
+      method: 'invoke_tool';
+      tool: string;
+      args?: Record<string, unknown>;
+    };
+
+async function handleRequest(msg: DaemonMessage) {
+  try {
+    if (msg.method === 'invoke_tool') {
+      if (!mcpClient) {
+        throw new Error('MCP client not initialized');
+      }
+      const {tool, args} = msg;
+
+      const result = (await mcpClient.callTool({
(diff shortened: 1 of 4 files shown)


src/daemon/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  // 1. Try XDG_RUNTIME_DIR (Linux standard, sometimes macOS)
+  if (process.env.XDG_RUNTIME_DIR) {
+    return path.join(process.env.XDG_RUNTIME_DIR, appName, 'server.sock');
+  }
+
+  // 2. macOS/Unix Fallback: Use /tmp/
+  // We use /tmp/ because it is much shorter than ~/Library/Application Support/
+  // and keeps us well under the 104-character limit.
+  return path.join('/tmp', `${appName}-${uid}.sock`);
```
Lightning00Blade: Should we use `os.tmpdir()`?

src/daemon/daemon.ts
```diff
@@ -0,0 +1,258 @@
...
+
+import {Client} from '@modelcontextprotocol/sdk/client/index.js';
+import {StdioClientTransport} from '@modelcontextprotocol/sdk/client/stdio.js';
+
+import {logger} from '../logger.js';
+import {PipeTransport} from '../third_party/index.js';
+
+import {getSocketPath, handlePidFile, INDEX_SCRIPT_PATH} from './utils.js';
+
+const IS_WINDOWS = os.platform() === 'win32';
```
Lightning00Blade: This should be in Utils and reused there i think

Lightning00Blade: Let's move this to utils and re-use it there.

src/daemon/daemon.ts
```diff
@@ -0,0 +1,258 @@
...
+  mcpTransport = new StdioClientTransport({
+    command: process.execPath,
+    args: [INDEX_SCRIPT_PATH, ...args],
+    env: process.env as Record<string, string>,
+  });
+  mcpClient = new Client(
+    {
+      name: 'chrome-devtools-cli-daemon',
+      // TODO: handle client version (optional).
+      version: '0.1.0',
```
Lightning00Blade, line 47: You can sue the special comments from release please to maintain this.
Plus you may need to and the file to the list of watch files.

Or extract the version file in a separate place for reusability. 

(Good separate task actually)

src/daemon/daemon.ts
```diff
@@ -0,0 +1,258 @@
...
+      const result = (await mcpClient.callTool({
+        name: tool,
+        arguments: args || {},
+      })) as McpResult | McpContent[];
+
+      // Extract text content from MCP response
+      let textContent = '';
+
+      // Check if result is an array of content blocks
+      if (Array.isArray(result)) {
```
Lightning00Blade: Can we extract this if/else logic in a separate function. I think it't very hard to follow.

src/daemon/daemon.ts
```diff
@@ -0,0 +1,258 @@
...
+  };
+
+  // Remove existing socket file if it exists (only on non-Windows)
+  if (!IS_WINDOWS) {
+    try {
+      await fs.unlink(socketPath);
+    } catch {
+      // ignore errors.
+    }
+    return await startServer();
```
Lightning00Blade: This is common for both if lets move if out of it.

src/daemon/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+import os from 'node:os';
+import path from 'node:path';
+import process from 'node:process';
+
+const _dirname = import.meta.dirname;
+
+export const DAEMON_SCRIPT_PATH = path.join(_dirname, 'daemon.js');
+export const INDEX_SCRIPT_PATH = path.join(_dirname, '..', 'index.js');
+
+const appName = 'chrome-devtools-mcp';
```
Lightning00Blade: Nit: Const need to be UPPER_CASE

src/daemon/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import process from 'node:process';
+
+const _dirname = import.meta.dirname;
```
Lightning00Blade: Please use directly without this intermediate variable.

src/daemon/daemon.ts
```diff
@@ -0,0 +1,258 @@
...
+  } else {
+    return startServer();
+  }
+}
+
+async function cleanup() {
+  console.log('Cleaning up daemon...');
+
+  if (mcpClient) {
```
Lightning00Blade: Non-blocking: Maybe we can safe some if else with using `?` over the file.

src/daemon/daemon.ts
OrKoN: I think we need to put the version into a separate file as a separate task.

src/daemon/daemon.ts
OrKoN: let me just stringify the whole result for now. I plan to revisit this and it was mostly meant as a placeholder.

PR #1021, opened 2026-02-23
chore: add -f argument for count token tool to read from file instead

matthiasrohmer:
Makes it a little easier to repeatedly run the skill counter on the contents of a file, instead of copy-pasting.


matthiasrohmer:
... or are you all using `cat | ...`? 😄 If yes happy to adapt.

Issue #1019, opened 2026-02-23
Support isolatedContext page-scoped tool routing for multi-agent workflows


passtas:
## Problem

PR #991 adds `isolatedContext` to `new_page`, enabling agents to create pages with isolated cookies, storage, and sessions. However, all subsequent tools (`take_screenshot`, `take_snapshot`, `navigate_page`, `evaluate_script`, `emulate`, `click`, `fill`, `press_key`, etc.) resolve their target page via the global `getSelectedPage()`. This means agents must call `select_page` before each operation, which introduces a race condition in multi-agent workflows.

The internal mutex serializes individual tool calls, but it doesn't prevent interleaving between two sequential calls from the same agent:

| Step | Mutex processes | `#selectedPage` after | Correct? |
|------|----------------|----------------------|----------|
| 1 | Agent A: `select_page(2)` | Page 2 | Yes |
| 2 | Agent B: `select_page(3)` | Page 3 | Yes |
| 3 | Agent A: `take_screenshot()` | Page 3 (captures wrong page) | No, wanted Page 2 |

Each call is atomic, but Agent A has no way to hold the mutex across both calls. Agent B's `select_page` mutates the shared `#selectedPage` state in between.

## Use cases
(genuinely using it like that)
 
**Parallel site exploration for specification writing.** A team of agents explores a large e-commerce site to produce functional specifications for each domain (navigation, product listing, product detail, checkout, account, etc.). A single agent doing this sequentially takes hours. With 3-5 parallel agents, each assigned a domain and operating in its own isolated context, the same work completes in a fraction of the time. Each agent needs to navigate, take snapshots, fill forms, and interact with elements independently.

**Cross-session cache and state testing.** Two agents investigate caching issues by browsing the same site simultaneously: one logged in as a premium user, the other as a guest. Isolated contexts give them separate cookies and sessions, but they need to take screenshots, evaluate scripts, and compare DOM state in parallel to catch bugs where one user sees another's cached data.

**Responsive design auditing.** Multiple agents emulate different viewports (mobile, tablet, desktop) on the same site in parallel. Each agent operates in its own isolated context with a different viewport configuration, taking screenshots and snapshots across the same set of pages to compare layout differences.

**Multi-locale testing.** Agents browse separate locale-specific versions of a site (en-US, de-DE, ja-JP) in parallel, each in its own isolated context with different language/geo settings, verifying that translations, currency formatting, and regional content are correct.

[...]

PR #1022, opened 2026-02-23
feat: add pageId routing for parallel multi-agent workflows

passtas:
## Summary

Adds optional `pageId` routing to page-scoped tools, gated behind `--experimental-page-id-routing`. When enabled, multi-agent callers can target a specific page without relying on global selection state. Fully backward-compatible: without the flag, behavior is unchanged.

### Key changes

- **`pageScoped` annotation**: tools declare `pageScoped: true`; the server merges `pageId` into their schema at registration time (when the flag is on)
- **`McpPage` wrapper**: consolidates per-page state (numeric id, isolated context name, focus tracking) into a single class
- **Request-scoped page routing**: `resolvePageById()` resolves the target page, `setRequestPage()` threads it through the handler so tools like `getSelectedPage()` see the right page
- **`assertPageIsFocused`**: keyboard/input tools validate that the target page holds browser focus, returning an actionable error ("call select_page first") instead of silently dispatching to the wrong page
- **`--experimental-page-id-routing` CLI flag** (hidden): gates schema injection and request-scoped routing so the feature can be tested before graduating
- **Eval scenarios**: `page_id_routing_test` and `page_focus_keyboard_test` with `serverArgs` support in the eval harness

Addresses #1019

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -14,6 +14,7 @@ import {
   UniverseManager,
   urlsEqual,
 } from './DevtoolsUtils.js';
+import {McpPage} from './McpPage.js';
 import type {ListenerMap, UncaughtError} from './PageCollector.js';
 import {NetworkCollector, ConsoleCollector} from './PageCollector.js';
 import type {DevTools} from './third_party/index.js';
@@ -38,48 +39,29 @@ import {takeSnapshot} from './tools/snapshot.js';
 import {CLOSE_PAGE_ERROR} from './tools/ToolDefinition.js';
 import type {Context, DevToolsData} from './tools/ToolDefinition.js';
 import type {TraceResult} from './trace-processing/parse.js';
+import type {
+  EmulationSettings,
+  GeolocationOptions,
+  TextSnapshot,
+  TextSnapshotNode,
+} from './types.js';
 import {
   ExtensionRegistry,
   type InstalledExtension,
 } from './utils/ExtensionRegistry.js';
 import {WaitForHelper} from './WaitForHelper.js';
 
-export interface TextSnapshotNode extends SerializedAXNode {
-  id: string;
-  backendNodeId?: number;
-  loaderId?: string;
-  children: TextSnapshotNode[];
-}
+export type {
+  EmulationSettings,
+  GeolocationOptions,
+  TextSnapshot,
+  TextSnapshotNode,
+} from './types.js';
 
 export interface ExtensionServiceWorker {
   url: string;
   target: Target;
   id: string;
-}
-
-export interface GeolocationOptions {
-  latitude: number;
-  longitude: number;
-}
-
-export interface TextSnapshot {
-  root: TextSnapshotNode;
-  idToNode: Map<string, TextSnapshotNode>;
-  snapshotId: string;
-  selectedElementUid?: string;
-  // It might happen that there is a selected element, but it is not part of the
-  // snapshot. This flag indicates if there is any selected element.
-  hasSelectedElement: boolean;
-  verbose: boolean;
-}
-
-interface EmulationSettings {
-  networkConditions?: string | null;
-  cpuThrottlingRate?: number | null;
-  geolocation?: GeolocationOptions | null;
-  userAgent?: string | null;
-  colorScheme?: 'dark' | 'light' | null;
-  viewport?: Viewport | null;
 }
 
 interface McpContextOptions {
@@ -129,18 +111,14 @@ export class McpContext implements Context {
 
   // Maps LLM-provided isolatedContext name → Puppeteer BrowserContext.
   #isolatedContexts = new Map<string, BrowserContext>();
-  // Reverse lookup: Page → isolatedContext name (for snapshot labeling).
-  // WeakMap so closed pages are garbage-collected automatically.
-  #pageToIsolatedContextName = new WeakMap<Page, string>();
   // Auto-generated name counter for when no name is provided.
   #nextIsolatedContextId = 1;
 
   #pages: Page[] = [];
   #extensionServiceWorkers: ExtensionServiceWorker[] = [];
 
-  #pageToDevToolsPage = new Map<Page, Page>();
+  #mcpPages = new Map<Page, McpPage>();
   #selectedPage?: Page;
-  #textSnapshot: TextSnapshot | null = null;
   #networkCollector: NetworkCollector;
   #consoleCollector: ConsoleCollector;
   #devtoolsUniverseManager: UniverseManager;
@@ -149,10 +127,8 @@ export class McpContext implements Context {
   #isRunningTrace = false;
   #screenRecorderData: {recorder: ScreenRecorder; filePath: string} | null =
     null;
-  #emulationSettingsMap = new WeakMap<Page, EmulationSettings>();
-  #dialog?: Dialog;
+  #focusedPagePerContext = new Map<BrowserContext, Page>();
 
-  #pageIdMap = new WeakMap<Page, number>();
   #nextPageId = 1;
 
   #extensionServiceWorkerMap = new WeakMap<Target, string>();
@@ -163,8 +139,6 @@ export class McpContext implements Context {
 
   #locatorClass: typeof Locator;
   #options: McpContextOptions;
-
-  #uniqueBackendNodeIdToMcpId = new Map<string, string>();
 
   private constructor(
     browser: Browser,
@@ -207,6 +181,10 @@ export class McpContext implements Context {
     this.#networkCollector.dispose();
     this.#consoleCollector.dispose();
     this.#devtoolsUniverseManager.dispose();
+    for (const mcpPage of this.#mcpPages.values()) {
+      mcpPage.dispose();
+    }
+    this.#mcpPages.clear();
     // Isolated contexts are intentionally not closed here.
     // Either the entire browser will be closed or we disconnect
     // without destroying browser state.
@@ -242,24 +220,32 @@ export class McpContext implements Context {
     return this.#networkCollector.getIdForResource(request);
   }
 
-  resolveCdpElementId(cdpBackendNodeId: number): string | undefined {
+  resolveCdpElementId(
+    cdpBackendNodeId: number,
+    page?: Page,
+  ): string | undefined {
     if (!cdpBackendNodeId) {
       this.logger('no cdpBackendNodeId');
       return;
     }
-    if (this.#textSnapshot === null) {
+    const snapshots = page
+      ? [this.#mcpPages.get(page)?.textSnapshot].filter(Boolean)
+      : [...this.#mcpPages.values()].map(mp => mp.textSnapshot).filter(Boolean);
+    if (!snapshots.length) {
       this.logger('no text snapshot');
       return;
     }
     // TODO: index by backendNodeId instead.
-    const queue = [this.#textSnapshot.root];
-    while (queue.length) {
-      const current = queue.pop()!;
-      if (current.backendNodeId === cdpBackendNodeId) {
-        return current.id;
-      }
-      for (const child of current.children) {
-        queue.push(child);
+    for (const snapshot of snapshots) {
+      const queue = [snapshot!.root];
+      while (queue.length) {
+        const current = queue.pop()!;
+        if (current.backendNodeId === cdpBackendNodeId) {
+          return current.id;
+        }
+        for (const child of current.children) {
+          queue.push(child);
+        }
       }
     }
     return;
@@ -305,7 +291,6 @@ export class McpContext implements Context {
         this.#isolatedContexts.set(isolatedContextName, ctx);
       }
       page = await ctx.newPage();
-      this.#pageToIsolatedContextName.set(page, isolatedContextName);
     } else {
       page = await this.browser.newPage({background});
     }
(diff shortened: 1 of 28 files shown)


OrKoN (changes requested):
I think the PR does not extend the focused page emulation and the dialog handling for all pages. Without it, tools might hang and not work as expected on background tabs.

passtas:
@OrKoN, I prototyped fixes for both and wanted to validate the approach before going further (see https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1022/changes/8edf6e8ef53cb7087a7439e7e67ac6a4a4105752).

**Dialog**: Replaced the singleton `#dialog` with `#pageDialogs = new Map<Page, Dialog>()`. Handlers are attached per-page via a `#setupPage()` method called when pages are first discovered. `getDialog` and `clearDialog` take an optional page param, defaulting to selected page for backward compat.

`handle_dialog` now resolves page via `pageIdSchema`.

**Focus**: Added `#focusedPagePerContext = new Map<BrowserContext, Page>()`. The issue is that with parallel agents interleaving `selectPage()` calls,  `#selectedPage` bounces between contexts so the wrong page gets defocused (or none does). Now `selectPage()` looks up the last focused page per-context,  so it always defocuses the correct same-context page. Entries are cleaned up in `closePage()`.

Both follow the same pattern: singleton field becomes a Map keyed by Page or BrowserContext, existing API stays backward-compatible via optional params, cleanup on page close.

Before I extend this further, wanted your take on introducing these Maps. The next candidate would be `#textSnapshot` which has the same singleton problem (one agent's uid references go stale when another agent snapshots a different page). That's a bigger change though, so I'd rather confirm you're good with the direction on these simpler cases first.

OrKoN:
I will need to think about it more but I think there could be the following opportunities:

1) instead of adding pageId schema in each tool definition file, we could have an annotation saying that a tool is page-scoped and then add the schema centrally. We can also avoid looking up the right page in the tool logic by passing request.page centrally.

2) I think we might be at the point where we have outgrown McpContext. Instead we would need to introduce an McpPage wrapper over Puppeteer's page to hold page related data instead of adding maps and simplify the McpContext interface by moving some actions to McpPage and removing the need for so many maps.

3) tools that already accept a uid or any other id should not need a page id, instead we need to make sure that IDs are unique across all pages.

OrKoN:
Another thing to keep in mind is that although performance tracing tools are scoped to a page it is not possible to start more than one tracing in the entire browser at the same time (even if pages are different).

src/McpContext.ts
```diff
@@ -556,41 +566,86 @@ export class McpContext implements Context {
     return page.getDefaultNavigationTimeout();
   }
 
+  // Linear scan over per-page snapshots. The page count is small (typically
+  // 2-10) so a reverse index isn't worthwhile given the uid-reuse lifecycle
+  // complexity it would introduce.
   getAXNodeByUid(uid: string) {
-    return this.#textSnapshot?.idToNode.get(uid);
+    for (const mcpPage of this.#mcpPages.values()) {
```
passtas, line 612: @OrKoN, here I just have done the linear scan and in other places as well. Would it make sense to have the separate index of `uidToPage` or it is OK as is for now?

OrKoN: I think it's fine for now but we might want to encode the pageId within the uid, e.g., `<pageId>_<objId>` eventually to have constant time lookups

src/tools/ToolDefinition.ts
```diff
@@ -108,22 +117,33 @@ export type Context = Readonly<{
...
   getAXNodeByUid(uid: string): TextSnapshotNode | undefined;
-  emulate(options: {
-    networkConditions?: string | null;
-    cpuThrottlingRate?: number | null;
-    geolocation?: GeolocationOptions | null;
-    userAgent?: string | null;
-    colorScheme?: 'dark' | 'light' | 'auto' | null;
-    viewport?: Viewport | null;
-  }): Promise<void>;
+  assertUidOnSelectedPage(uid: string): void;
```
OrKoN: this interface should only contain methods used by tools

src/McpContext.ts
OrKoN: I think console and network tools need to be marked page-scoped as well?

src/tools/ToolDefinition.ts
```diff
@@ -108,22 +117,33 @@ export type Context = Readonly<{
   recordedTraces(): TraceResult[];
   storeTraceRecording(result: TraceResult): void;
   getSelectedPage(): Page;
```
OrKoN, line 120: probably can delete this method to find all usages?

src/main.ts
```diff
@@ -204,9 +209,13 @@ function registerTool(tool: ToolDefinition): void {
         logger(`${tool.name} context: resolved`);
         await context.detectOpenDevToolsWindows();
         const response = args.slim ? new SlimMcpResponse() : new McpResponse();
```
OrKoN: we could make response instances also be scoped to a page to avoid passing the page from tools

src/McpContext.ts
```diff
@@ -122,16 +104,12 @@ export class McpContext implements Context {
...
   #isolatedContexts = new Map<string, BrowserContext>();
-  // Reverse lookup: Page → isolatedContext name (for snapshot labeling).
-  // WeakMap so closed pages are garbage-collected automatically.
-  #pageToIsolatedContextName = new WeakMap<Page, string>();
   // Auto-generated name counter for when no name is provided.
   #nextIsolatedContextId = 1;
 
   #pages: Page[] = [];
-  #pageToDevToolsPage = new Map<Page, Page>();
+  #mcpPages = new Map<Page, McpPage>();
```
OrKoN, line 120: I think it needs to be a weak map (which we probably cannot have due to the need to iterate over contents) or we need to listen to the close https://pptr.dev/api/puppeteer.pageevent and clean this map

src/tools/ToolDefinition.ts
passtas: Yeah, I wasn't so sure about the placement of it either. From one perspective it need to have access to the `#mcpPages`, from other it is used only within the tool. Not sure where to put it yet, but let me think about.

OrKoN: I mostly meant methods below (like getters) that were re-added.

src/tools/ToolDefinition.ts
```diff
@@ -173,12 +200,40 @@ export function defineTool<
...
+  wrapPageScopedHandler(definition);
   return definition;
 }
 
+function wrapPageScopedHandler<Schema extends zod.ZodRawShape>(
+  definition: ToolDefinition<Schema>,
+) {
+  if (definition.annotations.pageScoped) {
+    const originalHandler = definition.handler;
+    definition.handler = async (request, response, context) => {
```
passtas, line 209: @OrKoN, this another "dirty workaround" of mine. I was meant to ask how would you like to proceed here.
Reality is that this handler is only needed for the test, as in existing tests we are not passing `page `(obvisously).

The question: would you be happy for me to modify those existing tests (these will touch most of the tools)?

src/McpContext.ts
passtas: I'll check if it will work.

src/tools/ToolDefinition.ts
OrKoN: I think let's keep this to make the change manageable. I think we can also introduce the McpPage refactor as a separate PR. Eventually we should just pass the page in tests explicitly and test implicit page selection in a separate test.

src/McpContext.ts
passtas: 🤦
Good catch, let me check.

On the pageId to be part of the uid - should be easy enough, those loops concern me a fair bit.

src/tools/ToolDefinition.ts
passtas: Oh shoot, my bad. I think they reappeared during rebase conflict resolution. Thanks for catching.

src/tools/input.ts
```diff
@@ -284,6 +291,7 @@ export const drag = defineTool({
     includeSnapshot: includeSnapshotSchema,
   },
   handler: async (request, response, context) => {
+    context.assertUidOnSelectedPage(request.params.from_uid);
```
passtas: @OrKoN, what do you think about this assertion. Basically I was doing what you can call "chaos" test, and subagents tried to hover/input etc. on the non-selected tabs within their context. So of course they failed because there wasn't focus on those pages.

This will return guiding error to the agent to self-correct.

Alternative approach though (a bit more complicated) is that we have ability to emulate focus even on non-selected pages, so if agent intention was to hover over element on the subpage - it still can do it. 

I just didn't want to change to much of the current single-agent behaviour, but it could be nice improvement I think (theoretically that would mean that those subagents can work with different tabs even in single context) :)

Maybe not in scope for this PR, but could be nice continuation of work in this direction. What do you think?

OrKoN: I think context.getElementByUid should require a target page and the error should be thrown there

src/tools/ToolDefinition.ts
passtas: Fixed. The rebase onto main picked up #1033. And also I removed the remaining two: resolvePageById (only used in main.ts) and resolveCdpElementId (only used in McpResponse and McpContext internally).

src/tools/ToolDefinition.ts
passtas: Tried it, still a few tools uses it in slim mode and here for that test coverage. I think it worth planning bigger refactoring to reorg tests and do some other house-keeping items, so we can safely get rid of it.

src/main.ts
passtas: I went with a lighter approach: a request-scoped `#requestPage` field on McpContext. `main.ts` sets it before the handler, and all page-dependent context methods resolve via `#resolveTargetPage()` which returns `#requestPage ?? getSelectedPage()`. This gives the same result (response auto-resolves the correct page) without changing the Response interface. A full page-scoped response would pair well with the McpPage refactor you mentioned.

src/McpContext.ts
passtas: Added cleanup in `createPagesSnapshot()`: after building the page list, it prunes orphaned `#mcpPages` entries (calls dispose() and deletes) and stale  `#focusedPagePerContext` entries for pages that no longer exist.

src/tools/input.ts
passtas: Done. Removed `assertUidOnSelectedPage` entirely. Merged its logic into `getElementByUid`. Although introduced new assertion for the elements that doesn't use `getElementByUid` (also gap from my previous version)

src/McpContext.ts
passtas: Scoped the console and network. Checked others - added some corrections.

src/McpContext.ts
```diff
@@ -225,8 +205,16 @@ export class McpContext implements Context {
     return context;
   }
 
+  setRequestPage(page?: Page): void {
+    this.#requestPage = page;
```
OrKoN, line 212: This is not very nice that we rely on the context state to pass request specific data. Please add a TODO that this needs to refactored.

passtas: Agree, whole thing require refactoring tbh. Alternative was to thread around 11 method signatures with pageId. I reckoned I'll reuse selected page style of implementation.

src/main.ts
```diff
@@ -206,31 +211,41 @@ function registerTool(tool: ToolDefinition): void {
...
-        } = {
-          content,
-        };
-        success = true;
-        if (args.experimentalStructuredContent) {
-          result.structuredContent = structuredContent as Record<
-            string,
-            unknown
-          >;
+        const page = tool.annotations.pageScoped
```
OrKoN: let's add a CLI flag `--experimental-page-id-routing` (false by default) to control adding pageIdSchema and resolving page by id. I think I would like to do more refactoring and more testing before we can enable it by default.

src/McpContext.ts
OrKoN: Right, that's why I am only asking to add a TODO comment in this place so it is not forgotten.

src/main.ts
OrKoN: I think this is the last action item for this PR, and we can land it if the pageId is disabled by default

OrKoN:
I think there is a merge conflict caused by https://github.com/ChromeDevTools/chrome-devtools-mcp/commit/a8bf3e585682c3126dfd378e9f98b5dc7ab6045d To solve changes from main.ts should be moved to server.ts

passtas:
> I think there is a merge conflict caused by [a8bf3e5](https://github.com/ChromeDevTools/chrome-devtools-mcp/commit/a8bf3e585682c3126dfd378e9f98b5dc7ab6045d) To solve changes from main.ts should be moved to server.ts

Yeah, was just solving that :) 

Just FYI: there was one flaky test failure "should should detect open DevTools pages" only on Windows with Node 20. https://github.com/ChromeDevTools/chrome-devtools-mcp/actions/runs/22402585807/job/64955339393?pr=1022

Can you please rerun tests to see if it was flakiness or I actually broke it somehow?

passtas:
> > I think there is a merge conflict caused by [a8bf3e5](https://github.com/ChromeDevTools/chrome-devtools-mcp/commit/a8bf3e585682c3126dfd378e9f98b5dc7ab6045d) To solve changes from main.ts should be moved to server.ts
> 
> Yeah, was just solving that :)
> 
> Just FYI: there was one flaky test failure "should should detect open DevTools pages" only on Windows with Node 20. https://github.com/ChromeDevTools/chrome-devtools-mcp/actions/runs/22402585807/job/64955339393?pr=1022
> 
> Can you please rerun tests to see if it was flakiness or I actually broke it somehow?

This time it failed on Windows Node 23 :), certainly flaky.

@OrKoN, how would you like to proceed here?

OrKoN (approved):
LGTM thanks!

pirate:
related: @OrKoN @passtas you may be interested in my chromium CDP feature request to make foreground page tracking easier / avoid needing to maintain internal state in CDP clients to track it: https://issues.chromium.org/issues/497896141

Also in case it helps: At browser-use we made page IDs simply the last 4 characters of `targetIds`, which helped keep them stable / avoid needing in-memory mappings.

Issue #406, opened 2025-10-17
Add memory analysis tools for heap snapshots


natorion:
### Is your feature request related to a problem? Please describe.

_**FEEDBACK  WANTED**_

### 1. Problem Statement

The MCP server currently provides excellent tools for performance tracing (e.g., `performance_start_trace`), which allows an AI agent to analyze runtime performance, LCP, TBT, etc.

However, there is **no equivalent capability for memory debugging**. An AI agent cannot investigate memory leaks, detached DOM nodes, or object allocation patterns. This is a significant gap, as memory issues are a critical class of web performance problems.

A full heap snapshot file is far too large for an AI agent to process, so we need tools that run the analysis on the server and return a small, actionable JSON summary.

### 2. Proposed Feature

This feature would introduce a new set of tools to the MCP server, enabling an AI to request heap analysis and comparison.

#### Tool 1: Analyze Current Heap

* **`memory_analyze_heap()`**
    * **Action:** Takes a snapshot of the current heap.
    * **Analysis:** *Immediately* analyzes it on the server for common memory issues.
    * **Returns:** A structured JSON summary of the *current* state.
    * **Example JSON Response:**
        ```json
        {
          "total_heap_size": 12345678,
          "detached_dom_nodes": 15,
          "top_objects_by_retained_size": [
            { "constructor": "(string)", "count": 1024, "retained_size": 2048576 },
            { "constructor": "MyCustomObject", "count": 50, "retained_size": 1024567 },
            { "constructor": "(array)", "count": 800, "retained_size": 800123 }
          ]
        }
        ```

#### Tool 2: Diff/Comparison Workflow

To find leaks caused by a specific action, a comparison workflow is needed.

1.  **`memory_start_diff()`**
    * **Action:** Takes a "baseline" heap snapshot and stores it temporarily on the server.
    * **Returns:** A `diff_id` (e.g., `"diff-abc-123"`).

2.  **`memory_stop_diff(diff_id: string)`**
    * **Action:** Takes a "final" heap snapshot.
    * **Analysis:** Compares this new snapshot *on the server* against the baseline snapshot identified by `diff_id`.
    * **Returns:** A structured JSON summary of the *differences* (the delta).
    * **Example JSON Response:**
        ```json
        {
          "diff_id": "diff-abc-123",
          "heap_size_delta": 512345,
          "new_detached_dom_nodes": 5,
          "object_deltas": [
[...]

PR #1023, opened 2026-02-23
feat: add take_memory_snapshot tool

OrKoN:
Refs: https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/406


natorion:
> 1. Do we want this behind a flag?
> 2. @matthiasrohmer do you think we can create a skill right away?

Fundamental panel, so I think we should have this on per default. Let's put it behind the performane category though.

docs/tool-reference.md
```diff
@@ -28,10 +28,11 @@
 - **[Network](#network)** (2 tools)
   - [`get_network_request`](#get_network_request)
   - [`list_network_requests`](#list_network_requests)
-- **[Debugging](#debugging)** (5 tools)
+- **[Debugging](#debugging)** (6 tools)
   - [`evaluate_script`](#evaluate_script)
   - [`get_console_message`](#get_console_message)
   - [`list_console_messages`](#list_console_messages)
+  - [`take_memory_snapshot`](#take_memory_snapshot)
```
natorion: let's move it to performance

matthiasrohmer:
@OrKoN, asked @devnook to follow up with a skill this week.

PR #1024, opened 2026-02-23
fix: optimize token usage

OrKoN:
~10% savings with cl100k_base

from 4908 to 4046 with gemini.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -72,6 +72,8 @@ const server = new McpServer(
     name: 'chrome_devtools',
     title: 'Chrome DevTools MCP server',
     version: VERSION,
+    description:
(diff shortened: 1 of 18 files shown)


OrKoN:
@natorion @matthiasrohmer could you help with testing?

src/main.ts
```diff
@@ -72,6 +72,8 @@ const server = new McpServer(
     name: 'chrome_devtools',
     title: 'Chrome DevTools MCP server',
     version: VERSION,
+    description:
+      'All tools operate on the currently selected page. Many tools take a `uid` of an element from a page content snapshot.',
```
OrKoN, line 76: I am afraid this is not actually visible to MCP clients :(

PR #1028, opened 2026-02-24
refactor: optimize slim tool descriptions and params

OrKoN:
Minor changes to shorten descriptions and improve ease of use.

Note that `--slim` mode is not released yet and this is an improvement for it.

### DIFF
--- a/src/tools/slim/tools.ts
+++ b/src/tools/slim/tools.ts
@@ -12,7 +12,7 @@ import {defineTool} from '../ToolDefinition.js';
 
 export const screenshot = defineTool({
   name: 'screenshot',
-  description: `Take a screenshot of the active page.`,
+  description: `Takes a screenshot`,
   annotations: {
     category: ToolCategory.DEBUGGING,
     // Not read-only due to filePath param.
@@ -32,13 +32,13 @@ export const screenshot = defineTool({
 
 export const navigate = defineTool({
   name: 'navigate',
-  description: `Load URL in the browser`,
+  description: `Loads a URL`,
   annotations: {
     category: ToolCategory.NAVIGATION,
     readOnlyHint: false,
   },
   schema: {
-    url: zod.string().describe('Page URL'),
+    url: zod.string().describe('URL to navigate to'),
   },
   handler: async (request, response, context) => {
     const page = context.getSelectedPage();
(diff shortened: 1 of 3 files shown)


src/tools/slim/tools.ts
```diff
@@ -12,7 +12,7 @@ import {defineTool} from '../ToolDefinition.js';
 
 export const screenshot = defineTool({
   name: 'screenshot',
-  description: `Take a screenshot of the active page.`,
+  description: `Takes a screenshot`,
```
Lightning00Blade, line 15: Side note - we should enabled - https://typescript-eslint.io/rules/no-unnecessary-template-expression

OrKoN: why does it matter?

src/tools/slim/tools.ts
```diff
@@ -68,29 +68,21 @@ export const navigate = defineTool({
...
   handler: async (request, response, context) => {
     const page = context.getSelectedPage();
-    const fn = await page.evaluateHandle(`(${request.params.fn})`);
     try {
-      const result = await page.evaluate(async fn => {
-        // @ts-expect-error no types.
-        return JSON.stringify(await fn());
-      }, fn);
-      response.appendResponseLine(result);
+      const result = await page.evaluate(request.params.script);
```
Lightning00Blade, line 82: Was this an intentional change. I remember we did it as there were some issues.

src/tools/slim/tools.ts
OrKoN: if you ask me it only makes diffs worse once a string actually needs quotes or templates, so I am fine with always using `\``

src/tools/slim/tools.ts
OrKoN: yes, models cannot call the functions correctly without a ton of examples

src/tools/slim/tools.ts
Lightning00Blade: ACK, but we lose consistency.

src/tools/slim/tools.ts
Lightning00Blade: This should be a separate PR. Both the description and the title talk about text changes.
This is functionality. This makes it hard to discover in Changelogs. 
Also the history seems to be unavailable after the recent changes.

OrKoN: Discussed offline

PR #1031, opened 2026-02-24
chore: enable dynamic tool definition based on CLI arguments

nroscino:
This PR allows tools to be created dynamically based on the provided cli arguments


OrKoN:
`chore: ` because it is not ready for launching

src/main.ts
```diff
@@ -260,6 +260,7 @@ function registerTool(tool: ToolDefinition): void {
   );
 }
 
+const tools = createTools(args);
```
OrKoN, line 258: let's move `args.slim ? slimTools : tools` handling into createTools

OrKoN:
@nroscino we need to update generate docs script as well

OrKoN:
@nroscino could you please resolve the conflict?

PR #1035, opened 2026-02-24
chore: list service workers

nroscino:
This PR adds the extension service workers to the output of listPages. This is available only behind "--category-extensions" flag

### DIFF
--- a/src/tools/pages.ts
+++ b/src/tools/pages.ts
@@ -21,7 +21,7 @@ export const listPages = defineTool(args => {
     },
     schema: {},
     handler: async (_request, response) => {
-      response.setIncludePages(true, args?.categoryExtensions);
+      response.setIncludePages(true);
     },
   };
 });
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import type {ParsedArguments} from './cli.js';
 import {ConsoleFormatter} from './formatters/ConsoleFormatter.js';
 import {IssueFormatter} from './formatters/IssueFormatter.js';
 import {NetworkFormatter} from './formatters/NetworkFormatter.js';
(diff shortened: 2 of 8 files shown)


src/tools/pages.ts
```diff
@@ -11,17 +11,22 @@ import {zod} from '../third_party/index.js';
...
+    description: `Get a list of pages ${args?.categoryExtensions ? 'including extension service workers' : ''} open in the browser.`,
+    annotations: {
+      category: ToolCategory.NAVIGATION,
+      readOnlyHint: true,
+    },
+    schema: {},
+    handler: async (_request, response) => {
+      response.setIncludePages(true);
+      if (args?.categoryExtensions) {
+        response.setIncludeExtensionServiceWorkers(true);
```
OrKoN: let's set `setIncludeExtensionServiceWorkers` within `setIncludePages` in McpResponse so that all page listings include the service workers?

nroscino: should I also pass the flag and move the `if(categoryExtensions)` in the `setIncludePages` or we want them always to show up regardless of the flag in this case?

OrKoN: yeah, categoryExtensions should still apply

nroscino: Done!

src/tools/pages.ts
```diff
@@ -11,17 +11,19 @@ import {zod} from '../third_party/index.js';
...
+  return {
+    name: 'list_pages',
+    description: `Get a list of pages ${args?.categoryExtensions ? 'including extension service workers' : ''} open in the browser.`,
+    annotations: {
+      category: ToolCategory.NAVIGATION,
+      readOnlyHint: true,
+    },
+    schema: {},
+    handler: async (_request, response) => {
+      response.setIncludePages(true, args?.categoryExtensions);
```
OrKoN: let's pass it when we create the Response instances?  This still requires each tool pass the arg which is not needed.

PR #1037, opened 2026-02-25
chore: implement daemon client

OrKoN:
The client is meant to start the daemon if it is not running or connect to the existing one. Long-running client connections are not needed yet.

Removed previous direct daemon test as we better test it via the client, the intended way to manage the daemon.


src/daemon/daemon.ts
```diff
@@ -17,14 +18,32 @@ import {logger} from '../logger.js';
...
-  handlePidFile,
   INDEX_SCRIPT_PATH,
   IS_WINDOWS,
 } from './utils.js';
 
-const pidFile = handlePidFile();
+const pid = getDaemonPid();
+if (pid) {
+  try {
+    process.kill(pid, 0); // Throws if process doesn't exist
```
Lightning00Blade: Why are we double checking this we do it at `src/daemon/utils.ts:85`

src/daemon/client.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+export async function startDaemon(mcpArgs: string[] = []) {
+  if (isDaemonRunning()) {
+    console.log('Daemon is already running');
+    return;
+  }
+
+  logger('Starting daemon...');
+  const child = spawn(process.execPath, [DAEMON_SCRIPT_PATH, ...mcpArgs], {
```
Lightning00Blade, line 56: Does this need shell: true to work on Windows?

src/daemon/client.ts
```diff
@@ -0,0 +1,125 @@
...
+    child.on('error', err => {
+      reject(err);
+    });
+    child.on('exit', code => {
+      logger(`Child exited with code ${code}`);
+      reject(new Error(`Daemon process exited prematurely with code ${code}`));
+    });
+
+    waitForFile(getPidFilePath())
+      .then(() => resolve())
```
Lightning00Blade:
```suggestion
      .then(resolve)
```

src/daemon/client.ts
```diff
@@ -0,0 +1,125 @@
...
+
+import type {DaemonMessage} from './types.js';
+import {
+  DAEMON_SCRIPT_PATH,
+  getSocketPath,
+  getDaemonPid,
+  getPidFilePath,
+} from './utils.js';
+
+export function isDaemonRunning() {
```
Lightning00Blade: Why are we rewrapping the function here? Maybe we would have the 
Find Pid and is Pid running kinda separation in the logic.

src/daemon/client.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    const timer = setTimeout(() => {
+      fs.unwatchFile(filePath);
+      reject(
+        new Error(`Timeout: file ${filePath} not found within ${timeout}ms`),
+      );
+    }, timeout);
+
+    fs.watchFile(filePath, {interval: 500}, curr => {
```
Lightning00Blade, line 39: Should this be derivative from the timeout example `Math.max(timeout / 10, 10)`

src/daemon/client.ts
```diff
@@ -0,0 +1,125 @@
...
+        fs.unwatchFile(filePath); // Always clean up your listeners!
+        resolve();
+      }
+    });
+  });
+}
+
+export async function startDaemon(mcpArgs: string[] = []) {
+  if (isDaemonRunning()) {
+    console.log('Daemon is already running');
```
Lightning00Blade: Q: Do we have a fix/task for when the daemon is started with different args?

src/daemon/client.ts
```diff
@@ -0,0 +1,125 @@
...
+      reject(new Error('Timeout waiting for daemon response'));
+    }, SEND_COMMAND_TIMEOUT);
+
+    const transport = new PipeTransport(socket, socket);
+    transport.onmessage = async (message: string) => {
+      clearTimeout(timer);
+      logger('onmessage', message);
+      resolve(JSON.parse(message));
+    };
+    socket.on('error', error => {
```
Lightning00Blade, line 102: Should we listen on the `close` event as well?

src/daemon/daemon.ts
OrKoN: We actually do not need the one in src/daemon/utils.ts:85

src/daemon/client.ts
OrKoN: I do not think so although I have not tested with many arguments yet. The tests pass on Windows.

src/daemon/client.ts
OrKoN: Restrcutured.

src/daemon/client.ts
OrKoN: I do not think so, it is not striclty speaking related

src/daemon/client.ts
OrKoN: Yep!

src/daemon/client.ts
OrKoN: Yes, I will test this once I implement the CLI

src/daemon/daemon.ts
Lightning00Blade: This try block is not needed now.

OrKoN: why not?

src/daemon/daemon.ts
```diff
@@ -17,14 +18,33 @@ import {logger} from '../logger.js';
...
+  isDaemonRunning,
 } from './utils.js';
 
-const pidFile = handlePidFile();
+const pid = getDaemonPid();
+if (isDaemonRunning(pid)) {
+  try {
+    process.kill(pid, 0); // Throws if process doesn't exist
+    logger('Another daemon process is running');
+    process.exit(1);
```
Lightning00Blade: Should this be `exit(0)`. In practice it did "start" the daemon, it just reused an existing one.
Example for this is when you start a service on Linux multiple times it's a noop and returns a 0.

OrKoN: Discussed offline, that this would be the responsibility of the client and not of the server. Checks here are just safety precautions to not corrupt a different process data.

src/daemon/daemon.ts
OrKoN: Resolved, removed the redundant process-is-running check.

PR #1040, opened 2026-02-25
chore: add additional logs to clearcut sender

OrKoN:


src/telemetry/watchdog/ClearcutSender.ts
```diff
@@ -183,6 +184,7 @@ export class ClearcutSender {
     isPermanentError?: boolean;
     nextRequestWaitMs?: number;
   }> {
+    logger(`Sending batch ${events.length}`);
```
Lightning00Blade: Nit
```suggestion
    logger(`Sending batch of ${events.length}`);
```

PR #1049, opened 2026-02-26
chore: implement trigger extension action

nroscino:
This PR implements the ability to trigger an extension action by passing it an extension id. 
It uses puppeteer internals and the canary chrome version in tests since the TriggerAction CDP command is still not available in the current stable release.


tests/tools/extensions.test.ts
```diff
@@ -128,4 +129,39 @@ describe('extension', () => {
...
+          'triggerExtensionAction should be called with correct params',
+        );
+        assert.ok(
+          response.responseLines[0].includes(
+            `Extension action triggered. Id: ${extensionId}`,
+          ),
+          'Response should indicate action triggered',
+        );
+      },
+      {channel: 'chrome-canary'},
```
OrKoN: we cannot use the canary channel as it might not be installed on the system. Instead of we should download a Canary version using @puppeteer/browsers in CI and use a different executable path for those.

tests/tools/extensions.test.ts
```diff
@@ -128,4 +129,39 @@ describe('extension', () => {
       assert.ok(reinstalled, 'Extension should be present after reload');
     });
   });
+
+  it('triggers an extension action', async () => {
+    await withMcpContext(
+      async (response, context) => {
+        const triggerSpy = sinon.spy(context, 'triggerExtensionAction');
```
OrKoN: if the action is mocked, do we need `withMcpContext`? let's use a real extension test for this

tests/tools/extensions.test.ts
```diff
@@ -128,4 +135,72 @@ describe('extension', () => {
...
+          type: 'text';
+          text: string;
+        };
+        assert.ok(
+          textContent.text.includes(swUrl),
+          'Response should contain extension service worker url',
+        );
+      },
+      {
+        channel: process.env.CANARY_EXECUTABLE_PATH
```
OrKoN: we do not need to provide the channel if the path is provided

tests/utils.ts
```diff
@@ -47,10 +48,19 @@ let context: McpContext | undefined;
 
 export async function withBrowser(
   cb: (browser: Browser, page: Page) => Promise<void>,
-  options: {debug?: boolean; autoOpenDevTools?: boolean} = {},
+  options: {
+    debug?: boolean;
+    autoOpenDevTools?: boolean;
+    channel?: string;
```
OrKoN: remove channel?

tests/tools/extensions.test.ts
```diff
@@ -128,4 +135,69 @@ describe('extension', () => {
...
+        const swUrl = swTarget.url();
+
+        response.resetResponseLineForTesting();
+        await listPageDef.handler(
+          {params: {}, page: context.getSelectedMcpPage()},
+          response,
+          context,
+        );
+        result = await response.handle(listPageDef.name, context);
+        textContent = result.content.find(c => c.type === 'text') as {
```
OrKoN: let's use snapshot tests instead?

tests/tools/extensions.test.ts
```diff
@@ -128,4 +135,69 @@ describe('extension', () => {
       assert.ok(reinstalled, 'Extension should be present after reload');
     });
   });
+  it('triggers an extension action', async () => {
+    await withMcpContext(
+      async (response, context) => {
+        await installExtension.handler(
```
OrKoN: let's use `installExtension` from the context directly instead of using handlers for the setup.

tests/tools/extensions.test.js.snapshot
```diff
@@ -0,0 +1,13 @@
...
+# list_pages response
+## Pages
+1: about:blank [selected]
+## Extension Service Workers
+`;
+
+exports[`extension > triggers an extension action 2`] = `
+# list_pages response
+## Pages
+1: about:blank [selected]
```
nroscino: Missing header

tests/utils.ts
```diff
@@ -85,6 +101,8 @@ export async function withMcpContext(
     debug?: boolean;
     autoOpenDevTools?: boolean;
     performanceCrux?: boolean;
+    channel?: string;
```
OrKoN: remove channel as well?

tests/utils.ts
```diff
@@ -42,15 +44,29 @@ export function getImageContent(content: CallToolResult['content'][number]): {
   throw new Error(`Expected image content but got ${content.type}`);
 }
 
+export function extractId(response: McpResponse) {
```
OrKoN:
```suggestion
export function extractExtensionId(response: McpResponse) {
```

tests/tools/extensions.test.ts
```diff
@@ -128,4 +125,66 @@ describe('extension', () => {
...
     });
   });
+  it('triggers an extension action', async t => {
+    await withMcpContext(
+      async (response, context) => {
+        const extensionId = await context.installExtension(
+          EXTENSION_WITH_SW_PATH,
+        );
+
+        response.resetResponseLineForTesting();
```
OrKoN: not needed?

tests/tools/extensions.test.ts
```diff
@@ -128,4 +125,66 @@ describe('extension', () => {
...
+          context,
+        );
+        let result = await response.handle(listPageDef.name, context);
+        let textContent = result.content.find(c => c.type === 'text') as {
+          type: 'text';
+          text: string;
+        };
+        t.assert.snapshot?.(
+          textContent.text.replaceAll(extensionId, '<extension-id>'),
+        );
```
OrKoN: do we need these steps?

nroscino: Testing that the service worker is there after clicking doesn't guarantee that it was really activated by the action, it's possible that it can be active from the installation as well. Modifying the extension example could lead to this test passing while the action being actually broken. I think in general this might not be the best way to test the service worker activation via the click on the action.

A better test would probably be to either:

- Testing that the extension page connected to the action shows
- Listen on the event "onClicked" on the `chrome.action`, log something there and then access the logs on the service worker to verify the log is there. In this case we should remove the extension page from the extension.

Do you have any other ideas to make a better test?

OrKoN: You could verify with Puppeteer that an extension page got created on action. You could also verify with Puppeteer that the target does not exist before the action.

OrKoN (approved):

tests/tools/extensions.test.ts
nroscino: I like the idea! Changed it, thank you!

mathiasbynens:
~~For next time, please mark features like this as `feat:` rather than `chore:` so that it shows up properly in release notes & changelogs.~~

mathiasbynens:
> For next time, please mark features like this as `feat:` rather than `chore:` so that it shows up properly in release notes & changelogs.

Ignore this — it’s not a released feature yet, so `chore:` is actually the right thing. 

Once the CDP APIs are available in Stable Chrome, and docs have been updated, we can separately enable the feature by default (example: #1100). _That_ PR should then be marked as `feat:`.

PR #1052, opened 2026-02-26
chore: evaluate script on service workers

nroscino:
This PR introduce script evaluation on service workers behind the flag --category-extension.

### DIFF
--- a/src/tools/pages.ts
+++ b/src/tools/pages.ts
@@ -5,7 +5,7 @@
  */
 
 import {logger} from '../logger.js';
-import type {Dialog} from '../third_party/index.js';
+import type {CdpPage, Dialog} from '../third_party/index.js';
 import {zod} from '../third_party/index.js';
 
 import {ToolCategory} from './categories.js';
@@ -385,8 +385,7 @@ export const getTabId = definePageTool({
   },
   handler: async (request, response, context) => {
     const page = context.getPageById(request.params.pageId);
-    // @ts-expect-error _tabId is internal.
-    const tabId = page.pptrPage._tabId;
+    const tabId = (page.pptrPage as unknown as CdpPage)._tabId;
     response.setTabId(tabId);
   },
 });


src/McpContext.ts
```diff
@@ -510,7 +508,7 @@ export class McpContext implements Context {
     }
     if (page.isClosed()) {
       throw new Error(
-        `The selected page has been closed. Call ${listPages().name} to see open pages.`,
+        `The selected page has been closed. Call list_pages to see open pages.`,
```
OrKoN: did you mean to revert these?

src/tools/script.ts
```diff
@@ -4,23 +4,28 @@
...
 import {ToolCategory} from './categories.js';
+import type {Context} from './ToolDefinition.js';
 import {definePageTool} from './ToolDefinition.js';
 
-export const evaluateScript = definePageTool({
-  name: 'evaluate_script',
-  description: `Evaluate a JavaScript function inside the currently selected page. Returns the response as JSON,
+export type Evaluatable = Page | Frame | WebWorker;
+
+export const evaluateScript = definePageTool(cliArgs => {
```
OrKoN: this change would make it a non-page tool unfortunately and we should handle the optional pageId manually.

nroscino: I implemented the checks manually! PTAL

src/tools/script.ts
```diff
@@ -30,57 +35,129 @@ Example with arguments: \`(el) => {
...
+      const args: Array<JSHandle<unknown>> = [];
+      try {
+        const frames = new Set<Frame>();
+        for (const el of request.params.args ?? []) {
+          const handle = await context.getElementByUid(el.uid, page);
+          frames.add(handle.frame);
+          args.push(handle);
+        }
+
+        const evaluatable = await getEvaluatable(
```
OrKoN: we do not need to go through the page code path at all for service worker evaluation. We do not need to resolve the page and we do not need to handle uid args. In fact, if any are provided with the serviceWorkerId we should throw an error. I think you can create a dedicated if branch for serviceWorkerId and just extract eval functions into helper functions if needed.

src/McpContext.ts
```diff
@@ -622,7 +621,7 @@ export class McpContext implements Context {
       const mcpPage = this.#mcpPages.get(page);
       if (!mcpPage?.textSnapshot) {
         throw new Error(
-          `No snapshot found for page ${mcpPage?.id ?? '?'}. Use ${takeSnapshot.name} to capture one.`,
+          `No snapshot found for page ${mcpPage?.id ?? '?'}. Use take_snapshot to capture one.`,
```
OrKoN: revert?

src/McpContext.ts
```diff
@@ -663,9 +662,7 @@ export class McpContext implements Context {
       }
     }
     if (!anySnapshot) {
-      throw new Error(
-        `No snapshot found. Use ${takeSnapshot.name} to capture one.`,
-      );
+      throw new Error(`No snapshot found. Use take_snapshot to capture one.`);
```
OrKoN: ditto

src/tools/ToolDefinition.ts
```diff
@@ -5,6 +5,7 @@
  */
 
 import type {ParsedArguments} from '../cli.js';
+import type {ExtensionServiceWorker} from '../McpContext.js';
```
passtas: maybe move this to type to `../types.ts` for consistency to reuse between multiple consumers and avoid idiomatic circular dependencies?

OrKoN: circular dependencies for types are not an issue.

passtas: yep it still work functionally, but making my eye tick a bit :) 
I was specifically creating that shared types to avoid that confusion in the McpPage, as then McpPage would use something from McpContext and backwards. While it is just an idiomatic thing, at this stage would improve consistency with that approach and make it a tiny bit more expressive.

src/tools/script.ts
nroscino: Done! I also added a check on the pageId to avoid having both params at once instead of silently prefer the service worker. PTAL

src/tools/ToolDefinition.ts
nroscino: Done!

PR #1058, opened 2026-02-27
feat: add 13 JavaScript Debugger tools via CDP Debugger domain

soul-cat:
Add full debugger capabilities: enable/disable, breakpoints (set/remove/list), pause state inspection, stepping (over/into/out), call frame evaluation, script source retrieval, and script listing.

New file: src/tools/debugger.ts
Modified: McpPage (per-page debugger state + CDP event listeners), McpContext (13 debugger method implementations),
ToolDefinition (Context interface extension),
tools.ts (tool registration), types.ts (debugger interfaces)


OrKoN:
Thanks for the PR, could you please open a feature request instead? right now we do not have plans to offer debugger domain tools.

PR #1062, opened 2026-02-27
refactor: clean up more of the context getters

OrKoN:
- moves more code to the McpPage
- deleted Context specific tests

### DIFF
--- a/src/tools/pages.ts
+++ b/src/tools/pages.ts
@@ -54,7 +54,7 @@ export const selectPage = defineTool({
     context.selectPage(page);
     response.setIncludePages(true);
     if (request.params.bringToFront) {
-      await page.bringToFront();
+      await page.pptrPage.bringToFront();
     }
   },
 });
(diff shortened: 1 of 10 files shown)


src/tools/pages.ts
```diff
@@ -386,7 +386,7 @@ export const getTabId = definePageTool({
   handler: async (request, response, context) => {
     const page = context.getPageById(request.params.pageId);
     // @ts-expect-error _tabId is internal.
-    const tabId = page._tabId;
+    const tabId = page.pptrPage._tabId;
```
Lightning00Blade, line 389: Should we typecast this as `CdpPage` that should poke directly into the .d.ts of the file.

OrKoN: As long as we have a test it is fine, it's experimental.

OrKoN: https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1064

PR #1065, opened 2026-02-27
chore: extension pages list

nroscino:
This PR introduces extension pages in the listPages tool under the flag --category-extension.


src/McpContext.ts
```diff
@@ -589,6 +590,36 @@ export class McpContext implements Context {
       this.#options.experimentalIncludeAllPages,
     );
 
+    const allTargets = this.browser.targets();
```
OrKoN, line 593: do you know why `browser.pages()` does not return the extension pages? can we change it to return extension pages instead?

nroscino: I did a bit of investigation for this. 
Basically, when the target is created initially the type assigned to it is of type "other", therefore an instance of the type `OtherTarget` is created.

This snippet:
```typescript
browser.on("targetcreated", (target: Target) => {
    console.log("a target has been created: ", target.type());
});

// This outputs: 
// a target has been created: other
```

If later on we query for the `browser.targets()` and get the sidepanel or popup target, when we check the type is now of type "page", but the instance of the class is now the "OtherTarget". This leads to the call to the `page()` method that is not from `PageTarget` but from `OtherTarget`. `OtherTarget.page()` is taken from the parent implementation which simply returns a "null" value. 

Should we address this first in puppeteer and simplify the logic here afterwards?

OrKoN: I see, let's pass _isPageTarget https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/ConnectOptions.ts#L103 to Puppeteer instead to automatically treat it as a normal page?

nroscino: I don't think it's feasible. This is run when the target is still of type "other" correctly and the `url()` call returns an empty string in the snippet I posted above. 
Later on an update of the target happens (https://github.com/puppeteer/puppeteer/blob/a5ba71ab242a078b259188f091677386400d4e79/packages/puppeteer-core/src/cdp/TargetManager.ts#L124) in chrome and our listener changes the target infos to `type=page` and assign also an url as well. We can try to check on the listener for that event if there's a type change and create a new instance of type `PageTarget` to replace the `OtherTarget` instance we have.

OrKoN: The target should be re-created as a page target if you provide the callback https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/cdp/Browser.ts#L296

nroscino: Yes I understood that, my question was how to determine if the target I am receiving in the callback is actually an extension page to make sure to return a page type only for them. 

This is the information I have on target creation:
```js
{
  targetId: "<id>",
  type: 'other',
  title: '',
  url: '',
  attached: true,
  canAccessOpener: false,
  browserContextId: "<id>"
}
```
so in my callback I don't know what to do to make sure that I am converting just extension pages.

OrKoN: ah got it, is this data the `_getTargetInfo()` data?

nroscino: no this was on the internal of puppeteer, on Browser.ts on the `#createTarget`, before the call to the possible `_isPageTarget`.

OrKoN:
I removed it from the queue as I want to get the release in (which I tested manually). I will re-add the PR after the release happens.

tests/tools/pages.test.ts
```diff
@@ -101,6 +152,53 @@ describe('pages', () => {
         );
       });
     }
+
+    it('list pages for side panels with --category-extensions', async t => {
```
OrKoN, line 156: looks like this can be flaky https://github.com/ChromeDevTools/chrome-devtools-mcp/actions/runs/22737384192/job/65944081412#step:8:772

Issue #918, opened 2026-02-09
emulate tool schema uses anyOf, rejected by Claude API


ioku24:
## Problem

The `emulate` tool fails when used with Claude Code (and the Claude API in general) because the tool's input schema contains `anyOf`, which the Claude API does not support.

**Error:**
```
tools.N.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf
```

## Root Cause

In `build/src/tools/emulation.js`, the `geolocation` and `viewport` fields use `.nullable().optional()` on Zod objects:

```js
geolocation: zod
    .object({
        latitude: zod.number().min(-90).max(90),
        longitude: zod.number().min(-180).max(180),
    })
    .nullable()   // <-- produces anyOf
    .optional()

viewport: zod
    .object({
        width: zod.number().int().min(0),
        height: zod.number().int().min(0),
        deviceScaleFactor: zod.number().min(0).optional(),
        isMobile: zod.boolean().optional(),
        hasTouch: zod.boolean().optional(),
        isLandscape: zod.boolean().optional(),
    })
    .nullable()   // <-- produces anyOf
    .optional()
```

The bundled `zod-to-json-schema` converter in `third_party/index.js` (line ~22174) translates `.nullable()` on objects to:

```json
{
  "anyOf": [
    { "type": "object", "properties": { ... } },
    { "type": "null" }
  ]
}
```

Note: `.nullable()` on primitives (like `userAgent: zod.string().nullable()`) is fine — the converter uses `type: ["string", "null"]` for those. The issue only affects `.nullable()` on object schemas.

## Impact

- **Version affected:** 0.16.0 (latest via npx)
- **Tool affected:** `emulate` only (the other 25+ tools work fine)
- **Who's affected:** Anyone using this MCP server with Claude Code or the Claude API

## Suggested Fix

Replace `.nullable().optional()` with `.optional()` and handle `null` clearing logic at the tool handler level. For example:

```js
// Before:
geolocation: zod.object({ ... }).nullable().optional()

// After:
geolocation: zod.object({ ... }).optional()
// Handle "clear geolocation" via a separate boolean flag or empty object convention
```

Alternatively, post-process the generated JSON Schema to replace `anyOf: [{object}, {type: "null"}]` with just the object schema marked as optional.

## Environment

- `chrome-devtools-mcp` v0.16.0 (via npx)
- Claude Code CLI
- macOS (Apple Silicon)

[...]

PR #1073, opened 2026-03-02
fix: simplify emulation and script tools

OrKoN:
- reduces the token usage.
- makes emulation and script compatible with CLI.
- documents .nullable() and .object() restrictions for the future.
- the emulation tools do not have nullable anymore and undefined would clear the emulation instead. The model is thus required to provide all emulation settings at once.

Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/918

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,3 +119,10 @@ export const scenario: TestScenario = {
   },
 };
 ```
+
+## Restrictions on JSON schema
+
+- no .nullable(), no .object() types.
+- represent complex object as a short formatted string.
+
+TODO: implement eslint for schema https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1076
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -306,115 +306,95 @@ export class McpContext implements Context {
 
   async emulate(
     options: {
-      networkConditions?: string | null;
-      cpuThrottlingRate?: number | null;
-      geolocation?: GeolocationOptions | null;
-      userAgent?: string | null;
-      colorScheme?: 'dark' | 'light' | 'auto' | null;
-      viewport?: Viewport | null;
+      networkConditions?: string;
+      cpuThrottlingRate?: number;
+      geolocation?: GeolocationOptions;
+      userAgent?: string;
+      colorScheme?: 'dark' | 'light' | 'auto';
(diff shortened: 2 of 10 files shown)


src/tools/ToolDefinition.ts
```diff
@@ -310,3 +310,40 @@ export const timeoutSchema = {
...
+  }
+  const [dimensions, ...tags] = arg.split(',');
+  const isMobile = tags.includes('mobile');
+  const hasTouch = tags.includes('touch');
+  const isLandscape = tags.includes('landscape');
+  const [width, height, dpr] = dimensions.split('x').map(Number);
+  return {
+    width,
+    height,
+    deviceScaleFactor: dpr ?? undefined,
```
Lightning00Blade: This looks wrong, Let's typecast only what is needed - like 322.
The reset should be correct and we don't need `?? undefined` in any place.

src/McpContext.ts
```diff
@@ -306,115 +306,95 @@ export class McpContext implements Context {
...
-      if (options.cpuThrottlingRate === null) {
-        await page.emulateCPUThrottling(1);
-        delete newSettings.cpuThrottlingRate;
-      } else {
-        await page.emulateCPUThrottling(options.cpuThrottlingRate);
-        newSettings.cpuThrottlingRate = options.cpuThrottlingRate;
-      }
+    if (!options.networkConditions) {
+      await page.emulateNetworkConditions(null);
+      delete newSettings.networkConditions;
```
Lightning00Blade, line 324: Should we prefer `= undefined` it's cheaper.

CONTRIBUTING.md
```diff
@@ -119,3 +119,8 @@ export const scenario: TestScenario = {
   },
 };
 ```
+
+## Restrictions on JSON schema
+
+- no .nullable(), no .object() types.
```
Lightning00Blade, line 125: Let's add a task to add a EsLint rule it's should be straight forward.

src/McpContext.ts
OrKoN: absolutely not critical in my opinion

CONTRIBUTING.md
OrKoN: Created a task https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1076

PR #1077, opened 2026-03-02
chore: add a script to generate CLI

OrKoN:
The script will be used to generate CLI for the MCP server at `src/bin/cliDefinitions.ts`. 
It uses the live server JSON schemas to make it more compatible with the actual MCP server at runtime. The code definitions are used to extend the CLI with tool categories.


scripts/generate-cli.ts
```diff
@@ -0,0 +1,168 @@
...
+      description: tool.description || '',
+      category,
+      args,
+    };
+  }
+
+  const lines: string[] = [];
+  lines.push(`/**
+ * @license
+ * Copyright 2026 Google LLC
```
Lightning00Blade: Should this be the current year?

scripts/generate-cli.ts
```diff
@@ -0,0 +1,168 @@
...
+    },
+    {
+      capabilities: {},
+    },
+  );
+
+  await client.connect(transport);
+  try {
+    const toolsResponse = await client.listTools();
+    const tools = toolsResponse.tools || [];
```
Lightning00Blade, line 49: Super nit: let's throw here if there are no tools.

scripts/generate-cli.ts
```diff
@@ -0,0 +1,168 @@
...
+    const args: Record<string, CliOption> = {};
+    for (const opt of options) {
+      args[opt.name] = opt;
+    }
+    const category = toolNameToCategory.get(tool.name);
+    if (!category) {
+      throw new Error(`Tool ${tool.name} has no category.`);
+    }
+    commands[tool.name] = {
+      description: tool.description || '',
```
Lightning00Blade: Nit: If we expect all to have description let's throw if there is not one.

PR #1079, opened 2026-03-02
chore: implement chrome-devtools CLI

OrKoN:
Note that the implementation for response handling and daemon start/stop is still a placeholder. The proper handling will be done in follow-ups. The `generateCustomHelp` function is also fully experimental and it is likely to get removed before the release.  The goal of this PR is to scaffold the e2e workflow for early manual testing. Automated tests will be also added in a follow-up.

Note this does not expose the CLI as a binary on the package.


src/bin/chrome-devtools.ts
```diff
@@ -0,0 +1,148 @@
...
+  'start',
+  'Start or restart chrome-devtools-mcp',
+  y => y.help(false), // Disable help for start command to avoid parsing issues with passed args
+  async () => {
+    if (isDaemonRunning()) {
+      await stopDaemon();
+    }
+    // Extract args after 'start'
+    const startIndex = process.argv.indexOf('start');
+    const args = startIndex !== -1 ? process.argv.slice(startIndex + 1) : [];
```
Lightning00Blade, line 46: Do I understand correctly, here we don't use Yargs to avoid parsing back to string?

src/bin/cliDefinitions.ts
```diff
@@ -0,0 +1,741 @@
+/**
```
Lightning00Blade, line 1: Should this live in a `generated` folder?

src/bin/chrome-devtools.ts
OrKoN: Yes, to avoid potential serialization and escaping issues. I think for now we can forward as is.

src/bin/cliDefinitions.ts
OrKoN: There is no requirement to put generated files into a `generated` folder. I am fine with the current location but we can move later especially if more stuff will be generated.

PR #1082, opened 2026-03-03
chore: forward disclaimers

OrKoN:
`chrome-devtools start` should forward underlying disclaimers. This is achieved by piping stderr until the start indicator is seen.


OrKoN:
will be adding a bunch of CLI e2e tests soon

PR #1084, opened 2026-03-03
chore: fix arg forwarding in CLI

OrKoN:
Drive-by: e2e tests for major scenarios.
Drive-by: make stopDaemon more robust.


OrKoN:
Forwarding of stderr proven to be problematic, I changed the approach to log disclaimers separately in the CLI.

PR #1085, opened 2026-03-03
test: installing chrome through test script

nroscino:
Fixes local test failing for triggerAction on extensions.

### DIFF
--- a/scripts/test.mjs
+++ b/scripts/test.mjs
@@ -92,7 +92,7 @@ async function runTests(attempt) {
 }
 
 const chromePath = installChrome('146.0.7680.31');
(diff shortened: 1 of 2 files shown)


scripts/test.mjs
```diff
@@ -78,6 +91,9 @@ async function runTests(attempt) {
   });
 }
 
+const chromePath = installChrome('146.0.7680.31');
+process.env.TRIGGER_ACTION_EXECUTABLE_PATH = chromePath;
```
OrKoN: let's call it `CHROME_M146_EXECUTABLE_PATH`

PR #1089, opened 2026-03-03
chore: add a skill for chrome-devtools

OrKoN:
2060 tokens


OrKoN:
See https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/docs/cli.md Will update the skill later.

natorion (approved):

OrKoN:
P.S. I will land this once we are closer to making a release since skills are available immediately.

PR #1103, opened 2026-03-04
docs: add error quick-reference table to troubleshooting skill

natorion:
## Summary
- Add a quick-reference table at the top of Step 2 in the troubleshooting skill mapping common error messages/symptoms to likely causes and immediate first actions
- Covers 8 common scenarios sourced from the existing skill content and `docs/troubleshooting.md`
- Lets the model resolve frequent issues instantly without needing to WebFetch remote docs

## Test plan
- [ ] Verify the table renders correctly in GitHub markdown preview
- [ ] Confirm all error patterns in the table match real error messages from the codebase/docs


OrKoN:
@natorion do you still want to land this? can we close it if not?

PR #1106, opened 2026-03-04
chore: add workflow to sync webperf skills from webperf-snippets

nucliweb:
## Motivation

WebPerf skills originate in [nucliweb/webperf-snippets](https://github.com/nucliweb/webperf-snippets). This PR proposes a GitHub Actions workflow to automate keeping them in sync, so new releases are picked up automatically without manual intervention.

## How it works

- Checks the latest release from `nucliweb/webperf-snippets`
- Compares it against the last synced version stored in `.github/webperf-snippets-version`
- If a new release is available:
  - Downloads and extracts the skills into `skills/`
  - Adapts tool references from the standalone format (`mcp__chrome-devtools__tool_name`) to the MCP server internal format (`tool_name`)
  - Opens a PR with the changes
- Skips gracefully if already up to date or if a PR for that version already exists

## Schedule

The workflow runs every Monday at 09:00 UTC — this is a proposal and the cadence can be adjusted or replaced with a different trigger. It also supports manual runs via `workflow_dispatch`.

## Example execution

You can see the workflow in action here: nucliweb/chrome-devtools-mcp#2

### DIFF
--- a/.github/workflows/sync-webperf-skills.yml
+++ b/.github/workflows/sync-webperf-skills.yml
@@ -1,0 +1,128 @@
+name: Sync WebPerf Skills
+
+on:
+  schedule:
+    - cron: '0 9 * * 1' # Every Monday at 9:00 UTC
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  sync:
+    name: Sync skills from webperf-snippets
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Get latest release from webperf-snippets
+        id: upstream
+        run: |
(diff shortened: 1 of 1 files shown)


natorion:
In general I think it would be great to have the skills available within the DT MCP, thank you @nucliweb! I have a few remarks though:

- I don't think we want an automated import of 3rd party content.
- I think it makes sense first to migrate the skills over to this repo 1-by-1 - in particular because I also expect we will have to make some changes to optimize them for the MCP.

@OrKoN @devnook @matthiasrohmer for thoughts too please.

OrKoN:
Yes, we cannot mix third party code (skills in this case) with our own. I also was thinking that if you are open to upstreaming the skills we can maintain them here (+ we already have some skills around performance) so it probably makes sense to have skills be consistent and in the same style (instead of a second set of skills).

nucliweb:
Ok, make sense, I'll send another PR with the skills.


@natorion @OrKoN, WDYT about the structure and naming?

- webperf
- webperf-core-web-vitals
- webperf-interaction
- webperf-loading
- webperf-media
- webperf-resources

PR #1107, opened 2026-03-05
chore: handle images in the CLI

OrKoN:
- save base64 data as an image to a tmp file.
- change `--format` to `--output-format` to avoid conflicts.

### DIFF
--- a/src/utils/files.ts
+++ b/src/utils/files.ts
@@ -1,0 +1,26 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+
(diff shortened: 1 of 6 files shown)


src/utils/files.ts
```diff
@@ -0,0 +1,26 @@
...
+import os from 'node:os';
+import path from 'node:path';
+
+export async function saveTemporaryFile(
+  data: Uint8Array<ArrayBufferLike>,
+  filename: string,
+): Promise<{filepath: string}> {
+  try {
+    const dir = await fs.mkdtemp(
+      path.join(os.tmpdir(), 'chrome-devtools-mcp-'),
```
szuend, line 17: Is the tralining `-` intentional here?

OrKoN: yes, this is a prefix https://nodejs.org/api/fs.html#fspromisesmkdtempprefix-options

Issue #775, opened 2026-01-15
Connecting to Chrome with suspended/frozen tabs fails (usually with --auto-connect)


ProjectBetelgeuse:
**Update**: see a workaround here https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/775#issuecomment-3762914013

### Description of the bug

I have try to use the opencode and connect it to the MCP. I set it to autoConnect. Updated to Chrome 144 and enabled the Remote debugging. But it always timeout and I don't see any change in the chrome. It output  chrome-devtools_new_page [url=https://developers.chrome.com]. But it didn't open any new page and timed out. 

Also in the doc there is a inconsistance out the autoConnect version support. At some place it said 145+ but for the tutorial of how to autoConnect it said 144+. 

### Reproduction

1. Use "opencode"
2. Add MCP to to it. 
`    "chrome-devtools": {
      "type": "local",
      "command": ["npx", "-y","chrome-devtools-mcp@latest", "--autoConnect", "--channel=beta"],
    }` or
`    "chrome-devtools": {
      "type": "local",
      "command": ["npx", "-y","chrome-devtools-mcp@latest", "--autoConnect", "--channel=stable"],
    }` or
`    "chrome-devtools": {
      "type": "local",
      "command": ["npx", "-y","chrome-devtools-mcp@latest", "--autoConnect"],
    }`
3. Use chrome 144 and toggle the remote debugging. 
4. Tell AI in the opencode to check the performance of https://developers.chrome.com

### Expectation

For chrome 144+. The autoConnect should work by follow the doc in the readme. 

### MCP configuration

    "chrome-devtools": {
      "type": "local",
      "command": ["npx", "-y","chrome-devtools-mcp@latest", "--autoConnect", "--channel=stable"],
    }

### Chrome DevTools MCP version

v0.13.0

### Chrome version

144.0.7559.60

### Coding agent version

opencode 1.1.21

### Model version

Claude opus 4.5

### Node version

24.11.0

### Operating system

macOS

### Extra checklist

- [ ] I want to provide a PR to fix this bug

OrKoN:
@natorion @matthiasrohmer could you please help reproduce this?

OrKoN:
@ProjectBetelgeuse can you confirm that opencode actually installs the latest chrome-devtools-mcp@latest version and that you started the server in chrome://inspect/#remote-debugging

ProjectBetelgeuse:
[...]

PR #1109, opened 2026-03-05
fix: filter targetFilter to page-level targets to prevent connection timeouts

kevmo314:
## Summary

When connecting to a browser with many open tabs via `--browserUrl` or `--wsEndpoint`, the MCP server can time out with `Network.enable timed out` because Puppeteer attaches to **every** target (iframes, service workers, shared workers, webviews, background pages) and sends CDP initialization commands (`Network.enable`, `Page.enable`, `Runtime.enable`, etc.) to each one.

Frozen or suspended background tabs don't respond to these commands, causing the connection to hang until the protocol timeout (180s) is reached. A typical browsing session can easily have 50-60+ targets but only ~7 actual pages.

## Fix

Restricts `targetFilter` in `makeTargetFilter()` to only attach to `page` and `other` type targets. This is safe because:

- **Iframes** within pages are still accessible via `page.frames()` — they don't need to be separate targets
- **Network/console events** work through the page-level CDP session
- **Extension service workers** (`--categoryExtensions`) are disabled by default and were already filtered out by the `chrome-extension://` URL prefix filter

## Before/After

With ~58 browser targets (7 pages, 24 iframes, 13 service workers, 9 workers, 4 background pages, 1 webview):

- **Before**: `puppeteer.connect()` hangs for 180s then times out
- **After**: `puppeteer.connect()` attaches to ~7-9 targets and completes quickly

## Test plan

- [x] `tsc --noEmit` passes
- [ ] Verify `list_pages` returns all open pages
- [ ] Verify `select_page` + page interaction tools work
- [ ] Verify network request listing works on selected page
- [ ] Verify console message collection works
- [ ] Test with `--categoryExtensions` to confirm extension tools still work when launched (not connected)

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -28,6 +28,14 @@ function makeTargetFilter() {
   ]);
 
   return function targetFilter(target: Target): boolean {
+    // Only attach to page-level targets. Iframes, service workers, shared
+    // workers, webviews, and background pages are not needed for MCP page
+    // interactions and attempting to initialize them (Network.enable, etc.)
+    // on frozen or suspended targets can cause connection timeouts.
+    const type = target.type();
+    if (type !== 'page' && type !== 'other') {
+      return false;
+    }
     if (target.url() === 'chrome://newtab/') {
       return true;
     }


src/browser.ts
```diff
@@ -28,6 +28,14 @@ function makeTargetFilter() {
   ]);
 
   return function targetFilter(target: Target): boolean {
+    // Only attach to page-level targets. Iframes, service workers, shared
```
OrKoN, line 31: this is not not true, those targets are needed.

OrKoN:
Thanks for the PR! are you sure you are not hitting this issue with frozen tabs instead https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/775 ?

OrKoN:
> Iframes within pages are still accessible via page.frames() — they don't need to be separate targets

This is false as well.

kevmo314:
> Thanks for the PR! are you sure you are not hitting this issue with frozen tabs instead #775 ?

Ah it's possible this is the real issue I'm hitting, thanks for the link!

PR #1115, opened 2026-03-05
chore: configure cli args

OrKoN:
This PR configures CLI args for the start command. It removes arguments that do not make sense.


src/bin/chrome-devtools.ts
```diff
@@ -32,6 +32,35 @@ async function start(args: string[]) {
 
 const defaultArgs = ['--viaCli', '--experimentalStructuredContent'];
 
+const startCliOptions = {
```
Lightning00Blade, line 35: Is this used anywhere?

OrKoN: Lost in rebase, fixed.

PR #1117, opened 2026-03-05
chore: generate cli automatically

OrKoN:


package.json
```diff
@@ -13,7 +13,7 @@
     "typecheck": "tsc --noEmit",
     "format": "eslint --cache --fix . && prettier --write --cache .",
     "check-format": "eslint --cache . && prettier --check --cache .;",
-    "docs": "npm run build && npm run docs:generate && npm run format",
+    "docs": "npm run build && npm run docs:generate && npm run cli:generate && npm run format",
```
Lightning00Blade: Should this be part of the docs? It feels off being here as is generated code.

OrKoN: I can rename `docs` to `generate?`

OrKoN: renamed to `gen`

Lightning00Blade: Thanks!

PR #1120, opened 2026-03-05
refactor: move main files

OrKoN:
- `chrome-devtools-mcp.js` is the `npx chrome-devtools-mcp`
- `chrome-devtools.js` is the new CLI
- `-cli-options.js` is the corresponding options
- all these files are in the bin folder to indicate they are executable


src/index.ts
```diff
@@ -1,34 +1,281 @@
-#!/usr/bin/env node
-
 /**
  * @license
- * Copyright 2025 Google LLC
```
Lightning00Blade, line 5: Revert.

OrKoN: why? the file is moved

Lightning00Blade (approved):
LGTM + comment

Let's mark this as `refactor: change main files` as we touch the package.json original content.

src/index.ts
Lightning00Blade: From my understanding, you move a new file in place of the another. 
Technically is only a code update so one must keep the original license (year included).
Leaving to you to decide if this is valid.

OrKoN: if this was done in 2 PRs, there would be no need to update the header (like in the case of src/bin/chrome-devtools-mcp-cli-options.ts) since git correctly recognizes that as a move. The 2025 is from the old index.ts content, and the new file content (server.ts) is labelled 2026. I believe we only need to update the year when we create a new file with a different content.

PR #1121, opened 2026-03-05
test: improve cli test

OrKoN:
Use async spawn, and make the logs visible in CI.

### DIFF
--- a/tests/e2e/chrome-devtools.test.ts
+++ b/tests/e2e/chrome-devtools.test.ts
@@ -5,129 +5,147 @@
  */
 
 import assert from 'node:assert';
-import {spawnSync} from 'node:child_process';
+import {spawn} from 'node:child_process';
 import path from 'node:path';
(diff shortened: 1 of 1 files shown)


tests/e2e/chrome-devtools.test.ts
```diff
@@ -5,129 +5,147 @@
...
 
+async function runCli(
+  args: string[],
+): Promise<{status: number | null; stdout: string; stderr: string}> {
+  return new Promise((resolve, reject) => {
+    const child = spawn('node', [CLI_PATH, ...args]);
+    let stdout = '';
+    let stderr = '';
+    child.stdout.on('data', chunk => {
+      stdout += chunk;
```
Lightning00Blade, line 22: I think we need 
```suggestion
      stdout += chunk.toString();
```

OrKoN: not needed, see `let test = "" + {toString() {return 'test'; }}; console.log(test)`

PR #1132, opened 2026-03-06
test: fix flaky test

OrKoN:


tests/tools/pages.test.ts
```diff
@@ -162,13 +162,18 @@ describe('pages', () => {
 
           assert.ok(extensionId);
 
-          const _sidePanelPage = await context.newPage();
-          await _sidePanelPage.pptrPage.goto(
+          const sidePanelPage = await context.newPage();
+          await sidePanelPage.pptrPage.goto(
```
OrKoN, line 166: @nroscino I missed this during the review but I think we should test using an actual page created by extensions and not the regular page navigates to an extension URL to make sure it works. Could you please address that?

PR #1133, opened 2026-03-06
chore: report current cli args

OrKoN:
This CL records the args in the daemon and reports them back to the client via the status call.


OrKoN:
cc @natorion

PR #1100, opened 2026-03-04
feat: experimental Chrome DevTools CLI

OrKoN:
This PR enables an alternative CLI interface for Chrome DevTools MCP. `chrome-devtools` can be used in shell scripts to automate the Chrome DevTools MCP session running as a daemon process. Currently, a single session is shared between all `chrome-devtools` calls. All tools available by default are included into the CLI.

Examples:

```bash
❯ chrome-devtools navigate_page --url https://example.com
Successfully navigated to https://example.com.
## Pages
1: https://example.com/ [selected]

❯ chrome-devtools lighthouse_audit
## Lighthouse Audit Results
Mode: navigation
Device: desktop
URL: https://example.com/
### Category Scores
- Accessibility: 96 (accessibility)
- Best Practices: 96 (best-practices)
- SEO: 80 (seo)
### Audit Summary
Passed: 30
Failed: 4
Total Timing: 4107.22ms
### Reports
- /var/folders/hq/m1wr43z9665g5pghys7gkc2w00r23k/T/chrome-devtools-mcp-l5LRIT/report.json
- /var/folders/hq/m1wr43z9665g5pghys7gkc2w00r23k/T/chrome-devtools-mcp-3o6rcM/report.htm
```

```sh
❯ chrome-devtools lighthouse_audit --format=json | jq
{
  "lighthouseResult": {
    "summary": {
      "mode": "navigation",
      "device": "desktop",
      "url": "https://example.com/",
      "scores": [
        {
          "id": "accessibility",
          "title": "Accessibility",
          "score": 0.96
        },
        {
          "id": "best-practices",
          "title": "Best Practices",
          "score": 0.96
        },
        {
          "id": "seo",
          "title": "SEO",
          "score": 0.8
        }
      ],
      "audits": {
        "failed": 4,
        "passed": 30
      },
      "timing": {
        "total": 3754.87
      }
    },
    "reports": [
      "/var/folders/hq/m1wr43z9665g5pghys7gkc2w00r23k/T/chrome-devtools-mcp-nbLj82/report.json",
      "/var/folders/hq/m1wr43z9665g5pghys7gkc2w00r23k/T/chrome-devtools-mcp-VTp35b/report.html"
    ]
  }
}
```

To test:

```sh
npm ci
npm run clean
npm run bundle
npm pack
npm i chrome-devtools-mcp-0.18.1.tgz -g # to install globally
```

PR #1137, opened 2026-03-06
chore: work around the SDK's MCP client issues

OrKoN:
MCP SDK client does not set windowsHide when spawning the process unless it is on windows under Electron (not sure why). But I think we can safely pretend to be "electron".

Refs https://github.com/modelcontextprotocol/typescript-sdk/issues/1638
Refs https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1100


OrKoN:
cc @natorion

Issue #1140, opened 2026-03-07
Support deviceScaleFactor in --viewport parameter


Aalivexy:
### Is your feature request related to a problem? Please describe.

Currently, we need to use two separate parameters to set viewport with scaling:

- --viewport=1920x1080 for dimensions
- --chrome-arg=--force-device-scale-factor=2 for zoom level

### Describe the solution you'd like

Extend the --viewport parameter parsing to support an optional third value for deviceScaleFactor:

- Current format: 1920x1080 (width x height)
- Proposed format: 1920x1080x2 (width x height x deviceScaleFactor)

### Describe alternatives you've considered

Use emulate tool after startup: Set deviceScaleFactor via the emulate tool

Aalivexy:
https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1141#pullrequestreview-4043635314

PR #1141, opened 2026-03-07
feat: support device scale factor in --viewport

Aalivexy:
## Summary

This PR updates the server CLI --viewport flag to reuse the existing viewport syntax used by emulation tools.

`--viewport` now accepts both:

- `1280x720`
- `1280x720x2`

When a device scale factor is provided, it is applied during browser startup in addition to the initial viewport size.

## Changes

- add CLI parsing support for `<width>x<height>x<deviceScaleFactor>`
- preserve the existing `<width>x<height>` format
- apply `deviceScaleFactor` during initial browser launch
- update CLI documentation and examples
- add tests for CLI parsing and browser launch behavior

reslove #1140

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -63,6 +63,8 @@ interface McpContextOptions {
   experimentalIncludeAllPages?: boolean;
   // Whether CrUX data should be fetched.
   performanceCrux: boolean;
+  // Viewport emulation applied before the context was created.
+  initialViewport?: Viewport;
 }
 
 const DEFAULT_TIMEOUT = 5_000;
@@ -83,6 +85,16 @@ function getNetworkMultiplierFromString(condition: string | null): number {
       return 10;
   }
   return 1;
+}
+
+function normalizeViewportSettings(viewport: Viewport): Viewport {
+  return {
+    deviceScaleFactor: 1,
+    isMobile: false,
+    hasTouch: false,
+    isLandscape: false,
+    ...viewport,
+  };
 }
 
 export class McpContext implements Context {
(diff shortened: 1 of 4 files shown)


src/McpContext.ts
```diff
@@ -85,6 +87,16 @@ function getNetworkMultiplierFromString(condition: string | null): number {
   return 1;
 }
 
+function normalizeViewportSettings(viewport: Viewport): Viewport {
+  return {
+    deviceScaleFactor: 1,
```
Lightning00Blade, line 92: If we have this as one won't at some point try to set it to 1 which is wrong if the user has a out of the box set to 2 or something else?

Aalivexy: I'm not sure I understand what you mean. It does not override an explicit user value. `normalizeViewportSettings()` applies defaults first and then spreads `...viewport`, so an explicit `deviceScaleFactor` (for example `2`) wins over the fallback `1`.

Also, in the launch-time path we only persist `initialViewport` when the CLI viewport explicitly includes a DPR (`--viewport ...x2`). If no DPR was provided, we do not record an initial viewport override at all, so we do not replace the browser/OS default DPR with `1`.

Aalivexy: Are you suggesting that I should leave it as `undefined` when it is not explicitly set?

Lightning00Blade: First sorry for the slow reply.
Second both MCP and underlying Puppeteer have the wrong behaviour - they should not modify the DPR if user does not request that.

Given that this is already the case and this only moved from other place (missed that initially).
I will resolve, and will follow up with what issue I was observing separately, sorry for the confusion.

OrKoN:
@Lightning00Blade could you please review?

PR #1148, opened 2026-03-09
docs(troubleshooting): add symptom for missing tools due to read-only mode

natorion:


natorion:
> LGTM + we should update the text in `src/tools/ToolDefinition.ts:39` to explicitly call this out for use as maintainers, as we may hide things that don't make sense there.
> 
> Happy to do a followup after this is merged.

Thank, that would be great.

As discussed: The failing tests are related to infra, not this CL, so I will merge.

PR #1162, opened 2026-03-11
feat: add a skill for detecting memory leaks using take_memory_snapshot tool

### LINKED ISSUE #1056: Add a skill that describes how to debug memory leaks

https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/docs/tool-reference.md#take_memory_snapshot let's the ai agent take a heap snapshot. ai agents don' really know what to do with it, they suggest that the human opens devtools UI to analyze it. That is unfortunate.

Create a skill in https://github.com/ChromeDevTools/chrome-devtools-mcp/tree/main/skills that explains to the ai agent how find issues in the heap snapshot provide by `take_memory_snapshot`.

-

### Additional context
Rough notes:

`the idea is basically to take a heap snapshot in the beginning (startup), one after some possibly leaky action has been performed, and a final one after everything should have been cleaned up. then compare the 3 in diff mode.`
Relevant links:
- https://developer.chrome.com/docs/devtools/memory-problems?hl=en, in particular https://developer.chrome.com/docs/devtools/memory-problems?hl=en#discover_detached_dom_tree_memory_leaks_with_heap_snapshots
- https://developer.chrome.com/docs/devtools/memory-problems/heap-snapshots?hl=en
- https://github.com/facebook/memlab

devnook:
This PR add a new skill for detecting and removing memory leaks using take_memory_snapshot tool.

Skill was based on authoritative material from own sources:
* https://developer.chrome.com/docs/devtools/memory-problems
* https://web.dev/articles/effectivemanagement
as well as examples of memory leaks from public domain.

It can be tested with https://github.com/devnook/mem-leak-example. Example prompt
* A page running at localhost:3000 has memory issues when you click from home to contact tab and back. Can you fix it?   

The skill prefers using comparison tools like memlab instead of custom evaluate_script snippets to avoid directly reading of .heapsnpashot files to save context space and tokens.

Fixes #1056

### DIFF
--- a/skills/memory-leak-debugging/SKILL.md
+++ b/skills/memory-leak-debugging/SKILL.md
@@ -11,7 +11,7 @@ This skill provides expert guidance and workflows for finding, diagnosing, and f
 
 - **Prefer `memlab`:** Do NOT attempt to read raw `.heapsnapshot` files directly, as they are extremely large and will consume too many tokens. Always recommend and use `memlab` to process snapshots and identify leak traces.
 - **Isolate the Leak:** Determine if the leak is in the browser (client-side) or Node.js (server-side).
-- **Common Culprits:** Look for detached DOM nodes, unhandled closures, global variables, event listeners not being removed, and caches growing unbounded. *Note: Detached DOM nodes are sometimes intentional caches; always ask the user before nulling them.*
+- **Common Culprits:** Look for detached DOM nodes, unhandled closures, global variables, event listeners not being removed, and caches growing unbounded. _Note: Detached DOM nodes are sometimes intentional caches; always ask the user before nulling them._
 
 ## Workflows
 
@@ -42,7 +42,9 @@ When you have found a leak trace (e.g., via `memlab` output), you must identify 
 If `memlab` is not available, you MUST use the fallback script in the references directory to compare two `.heapsnapshot` files and identify the top growing objects and common leak types.
 
 Run the script using Node.js:
+
 ```bash
 node compare_snapshots.js <baseline.heapsnapshot> <target.heapsnapshot>
 ```
+
 The script will analyze and output the top growing objects by size and highlight the 3 most common types of memory leaks (e.g., Detached DOM nodes, closures, Contexts) if they are present.
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -37,6 +37,7 @@ export default defineConfig([
             'puppeteer.config.cjs',
             'eslint.config.mjs',
             'rollup.config.mjs',
+            'skills/memory-leak-debugging/references/compare_snapshots.js',
           ],
         },
       },
--- a/skills/memory-leak-debugging/references/compare_snapshots.js
+++ b/skills/memory-leak-debugging/references/compare_snapshots.js
@@ -1,4 +1,10 @@
-const fs = require('fs');
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as fs from 'node:fs';
 
 function parseSnapshot(filePath) {
   console.log(`Loading ${filePath}...`);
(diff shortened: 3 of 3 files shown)


skills/memory-leak-debugging/SKILL.md
```diff
@@ -0,0 +1,37 @@
...
+description: Diagnoses and resolves memory leaks in JavaScript/Node.js applications. Use when a user reports high memory usage, OOM errors, or wants to analyze heapsnapshots or run memory leak detection tools like memlab.
+---
+
+# Memory Leak Debugging
+
+This skill provides expert guidance and workflows for finding, diagnosing, and fixing memory leaks in JavaScript and Node.js applications.
+
+## Core Principles
+
+- **Prefer `memlab`:** Do NOT attempt to read raw `.heapsnapshot` files directly, as they are extremely large and will consume too many tokens. Always recommend and use `memlab` to process snapshots and identify leak traces.
```
natorion, line 12: How about suggesting that if memlab is not available, use jq to parse the heapsnapshot instead of ingesting the whole thing?

skills/memory-leak-debugging/references/common-leaks.md
```diff
@@ -0,0 +1,33 @@
+# Common Memory Leaks
+
+When analyzing a retainer trace from `memlab`, look for these common patterns in the codebase:
```
mlippautz, line 3: I have not worked with `memlab` before. 

Would it make sense as a follow up to work on some native mcp utilities to navigate heap snapshots directly?

I know that people on my team have been vibe coding stand-alone utitlities for working with snapshots based on the existing heap snapshot readers in Chrome. Basically, they extracted the format into a stand-alone JS tool, and then even asked for a rewrite of the whole parser in Rust.

skills/memory-leak-debugging/references/common-leaks.md
```diff
@@ -0,0 +1,33 @@
...
+
+## 1. Detached DOM Nodes
+
+A DOM node is removed from the document tree but is still referenced by a JavaScript variable.
+
+**Fix:** Ensure variables holding DOM references are set to `null` when the node is removed, or limit their scope.
```
mlippautz: This can also be just a cache of some sorts. We know websites (also 1P) that just cache whole navigation trees on the side that are detached. In that sense, while detachedness is a good signal, it's not always correct to just null the reference.

skills/memory-leak-debugging/references/common-leaks.md
```diff
@@ -0,0 +1,33 @@
...
+
+## 2. Uncleared Event Listeners
+
+Event listeners attached to global objects (like `window` or `document`) or long-living objects prevent garbage collection of the objects referenced in their callbacks.
```
mlippautz: This is probably the number 1 reason for unintended leaks of whole iframes (windows). The callbacks always refer to global objects (via constructors, that are specified in JS) which cause the whole iframe to leak. 

This leaks are generally big in terms of memory and actionable.

skills/memory-leak-debugging/SKILL.md
```diff
@@ -0,0 +1,37 @@
...
+- Use tools like `click`, `navigate_page`, `fill`, etc., to manipulate the page into the desired state.
+- Revert the page back to the original state after interactions to see if memory is released.
+- Repeat the same user interactions 10 times to amplify the leak.
+- Use `take_memory_snapshot` to save `.heapsnapshot` files to disk at baseline, target (after actions), and final (after reverting actions) states.
+
+### 2. Using Memlab to Find Leaks (Recommended)
+
+Once you have generated `.heapsnapshot` files using `take_memory_snapshot`, use `memlab` to automatically find memory leaks.
+
+- Read [references/memlab.md](references/memlab.md) for how to use `memlab` to analyze the generated heapsnapshots.
```
mlippautz, line 31: Not sure if memlab can do that but for manual debugging (using DevTools) one can get the retaining path of objects. This path is often very useful for finding the culprit of the leak.

E.g., if retainers reach back to internals it may show platform leaks. If retainers show a transition to an attached DOM tree, then that's where a cache of some sort is held.

mlippautz:
This generally looks good and I am curious to actually try it out :)

skills/memory-leak-debugging/references/common-leaks.md
natorion: I think we can add a js file as a reference to the skill.

Lightning00Blade:
Note: some tooling like Bun provide a way to export this information as Markdown - https://bun.com/docs/project/benchmarking#markdown-output-2

natorion:
> Note: some tooling like Bun provide a way to export this information as Markdown - https://bun.com/docs/project/benchmarking#markdown-output-2

Seems like the source for this is at https://github.com/oven-sh/bun/blob/a305e99af5d9dbe74ffabfa5a3c0cfdf815ac71a/src/bun.js/bindings/BunHeapProfiler.cpp#L568. Can we embed that as a pythons script or TS script somehow?

skills/memory-leak-debugging/SKILL.md
devnook: I added the compare_snapshots.js script as a fallback

skills/memory-leak-debugging/references/common-leaks.md
devnook: I added the compare_snapshots.js script as a fallback

skills/memory-leak-debugging/references/common-leaks.md
devnook: I added the cache warning to the skill recommendations

skills/memory-leak-debugging/references/common-leaks.md
devnook: Upgraded to nr1 in the leaks list

skills/memory-leak-debugging/SKILL.md
devnook: I'm not sure how to incorporate this into the skill. Do you have an example where this could be tested (to compare cache vs leak scenario)?

devnook:
> Note: some tooling like Bun provide a way to export this information as Markdown - https://bun.com/docs/project/benchmarking#markdown-output-2

Is there an advantage to .md format instead of raw .heapsnaphot files?

devnook:
I added the compare_snapshots.js script as a fallback to memlab to compare snapshot and rewrote some of the skills guidance to include remarks from the review. PTAL

Lightning00Blade:
> Is there an advantage to .md format instead of raw .heapsnaphot files?

From the post I am seeing it's ease of use with LLM, as it can directly use the data without needing to do other steps.
Though I have not see a actual benchmark that confirms this.

skills/memory-leak-debugging/SKILL.md
mlippautz: I think this is probably left for follow up work as I am not sure the external tools are capable of doing this programatically.

devnook:
Added a possible follow up issue for retaining path: https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1224

PR #1176, opened 2026-03-14
feat: add extraHttpHeaders emulation to emulate tool

### LINKED ISSUE #1175: Feature request: set_extra_http_headers tool for Network.setExtraHTTPHeaders
### Problem

There is currently no way to set custom HTTP headers on **all** requests (including the initial document navigation and `<script>` tag loads) via the MCP tools.

The `navigate_page` tool supports `initScript`, which can intercept `fetch()` and `XMLHttpRequest` calls at runtime. However, `initScript` runs **after** the document is already fetched — it cannot set headers on:
- The initial HTML document request (navigation)
- `<script src="...">` tag loads (parsed from HTML)  
- CSS, image, and font loads

### Proposed Solution

Add a `set_extra_http_headers` tool that calls Puppeteer's `page.setExtraHTTPHeaders()` (which uses CDP `Network.setExtraHTTPHeaders` under the hood).

**Tool definition:**

```
name: set_extra_http_headers
description: Set extra HTTP headers to be sent with every request. Headers persist until cleared. Pass an empty object to clear.
schema:
  headers:
    type: object
    description: "Header name-value pairs, e.g. {\"X-Custom\": \"value\"}"
```

**Behavior:**
- Headers apply to ALL subsequent requests on the selected page (document, scripts, fetch, XHR, images, etc.)
- Headers persist across navigations until explicitly cleared with `{}`
- Fits naturally in the existing `network` category alongside `list_network_requests` and `get_network_request`

### Use Case

We use `chrome-devtools-mcp` for visual regression testing of a Shopify storefront widget. The widget's JS bundle is served via a `<script>` tag in the HTML. To test a PR branch, the storefront server needs to see `X-Promise-Tophat: <branch>` on the **document request** to serve the branch-specific bundle URL.

Currently we work around this by using a separate tool (`agent-browser`) that supports `--headers` via CDP. Having `set_extra_http_headers` in chrome-devtools-mcp would eliminate this workaround.

### Implementation Notes

The implementation should be straightforward since the MCP already has access to the Puppeteer `page` object:

```typescript
// In tools/network.ts
await page.setExtraHTTPHeaders(params.headers);
```

This is a one-liner on top of existing Puppeteer infrastructure.

pedrodurek:
## Summary

Extend the existing `emulate` tool with an `extraHTTPHeaders` parameter that calls Puppeteer's `page.setExtraHTTPHeaders()` (which uses CDP `Network.setExtraHTTPHeaders` under the hood).

Closes #1175

## Approach

Per [feedback from @natorion](https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1175#issuecomment-4097587153), this integrates into the existing `emulate` tool rather than adding a standalone tool. The `emulate` tool is already the central hub for page-level state modifications (userAgent, viewport, networkConditions, geolocation, colorScheme), and custom HTTP headers fit naturally alongside them. This also avoids increasing the MCP tool count and LLM token overhead.

## Changes

- **`src/types.ts`** — Added `extraHTTPHeaders?: Record<string, string>` to `EmulationSettings`
- **`src/tools/emulation.ts`** — Added `extraHTTPHeaders` as an optional zod parameter on the `emulate` tool
- **`src/McpContext.ts`** — Added handler logic in the `emulate()` method:
  - Calls `page.setExtraHTTPHeaders()` when `extraHTTPHeaders` is provided
  - Clears from settings when an empty `{}` is passed
  - Preserves existing headers when the param is **omitted** (unlike other emulation settings that reset when omitted) — prevents `emulate({colorScheme: "dark"})` from accidentally clearing previously-set headers
- **`tests/tools/emulation.test.ts`** — Added 5 test cases:
  1. Sets extra headers on requests
  2. Clears headers with `{}`
  3. Headers persist across navigations
  4. Does not affect other emulation settings
  5. Reports correctly per-page (new page has no headers)

## Use Case

This enables setting custom HTTP headers on **all** requests — including the initial document navigation and `<script>` tag loads — which `initScript` cannot do since it runs after the document is already fetched.

## Usage

```js
// Set headers
emulate({ extraHTTPHeaders: { "X-Custom": "value", "Authorization": "Bearer token" } })

// Clear headers
emulate({ extraHTTPHeaders: {} })

// Combine with other emulation settings
emulate({ extraHTTPHeaders: { "X-Branch": "feature-1" }, userAgent: "MyBot/1.0" })
```


OrKoN (changes requested):
Thanks for the PR! I think we should align with the suggestion in this comment https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1175#issuecomment-4097587153 and add this to the emulation tool.

pedrodurek:
@OrKoN I moved the functionality to the emulation tool. Let me know if further adjustments are needed.

src/tools/emulation.ts
```diff
@@ -65,6 +65,12 @@ export const emulate = definePageTool({
       .describe(
         `Emulate device viewports '<width>x<height>x<devicePixelRatio>[,mobile][,touch][,landscape]'. 'touch' and 'mobile' to emulate mobile devices. 'landscape' to emulate landscape mode.`,
       ),
+    extraHTTPHeaders: zod
+      .record(zod.string(), zod.string())
```
OrKoN: Unfortunately, this would not work with CLI interface out of the box. Let's change this to be a string parameter and the string should contain a serialized JSON object containing headers. This would allow use to safely parse multi line headers and will be compatible with the CLI interface. Empty string can be then used to unset the headers.

OrKoN (changes requested):
Thanks! mostly looks good, one remaining comment about the format of the field.

pnpm-lock.yaml
OrKoN, line 1: please remove the pnpm lock file

OrKoN (changes requested):
Thanks it looks good, but it appears some compilation and test issues have to be addressed.

src/telemetry/tool_call_metrics.json
```diff
@@ -85,6 +85,10 @@
       {
         "name": "viewport_length",
         "argType": "number"
+      },
+      {
+        "name": "extra_h_t_t_p_headers_length",
```
OrKoN: cc @yulunz is this generating the way we want it?

yulunz: huh, @pedrodurek would it be possible to call this `extraHttpHeadersLength`?

In other places we have been following this format: `McpClient`, `McpResponse` etc (although MCP is otherwise all capitalized).

src/McpContext.ts
```diff
@@ -379,6 +380,18 @@ export class McpContext implements Context {
       newSettings.viewport = viewport;
     }
 
+    if (options.extraHTTPHeaders !== undefined) {
+      if (options.extraHTTPHeaders === null) {
```
yulunz: nit: it seems null here is used to mean that the string was empty. Also the transform function already returns undefined. I wonder if there is a better way. For example, can it just return an empty object when the string is empty? So here we can have:

```
if (options.extraHTTPHeaders !== undefined) {
  await page.setExtraHTTPHeaders(options.extraHTTPHeaders);
  newSettings.extraHTTPHeaders = options.extraHTTPHeaders;
  if (Object.entries(options.extraHTTPHeaders).length === 0) {
    delete newSettings.extraHTTPHeaders;
  }
}
```

this is pretty self-explanatory and also takes care of the case when the string passed is an empty object.

src/telemetry/tool_call_metrics.json
```diff
@@ -85,6 +85,16 @@
...
+      },
+      {
+        "name": "extra_http_headers_length",
+        "argType": "number",
+        "isDeprecated": true
+      },
+      {
+        "name": "extra_h_t_t_p_headers_length",
+        "argType": "number",
+        "isDeprecated": true
```
OrKoN: @nroscino I think this was never merged and we should delete this entry for extra_h_t_t_p_headers_length

yulunz: Thanks for fixing this! And yeah, you can just revert all changes in this file, and re-run `npm run gen` which will regenerate the correct outcome.

nroscino: Done! Let me know if I should address anything else!

OrKoN (approved):

src/telemetry/tool_call_metrics.json
OrKoN: Thanks!

PR #1189, opened 2026-03-15
chore: add ESLint guard against direct third-party imports bypassing bundle

### LINKED ISSUE #1123: Testing blind spot around bundling
We have a testing blind spot: it is possible to include an import to a npm module directly bypassing the third_party/index.js which means the import will fail once bundled. Although we test the bundle, our dev dependencies are still present in the test environment (and they are need for tests themselves). 

We should have an eslint rule for this + better that our e2e test runs in a separate folder where we unpack npm package that we previously packed with `npm pack`.

Example of this is https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1111 (found through manual release testing with `npm pack`)

cc @zyzyzyryxy @Lightning00Blade

mvanhorn:
## Summary

- Adds a custom ESLint rule `@local/no-direct-third-party-imports` that flags value imports of bundled third-party packages (`@modelcontextprotocol/sdk`, `puppeteer-core`, `@puppeteer/browsers`, `yargs`, `debug`, `zod`, `core-js`) when used outside of `src/third_party/`
- Type-only imports (`import type`) are allowed since they are erased at compile time and don't affect the bundle
- The rule is scoped to `src/**/*.ts` so development scripts and tests are unaffected

This prevents the class of bugs where a direct npm import works during development (devDependencies installed) but breaks in the published package (only bundled code ships). PR #1111 was an example of this exact issue caught through manual `npm pack` testing.

Closes #1123

## Test plan

- [x] Verified `npx eslint --no-cache src/` passes with no violations on the current codebase
- [x] Verified the rule correctly catches a test file with `import {Client} from '@modelcontextprotocol/sdk/client/index.js'`
- [x] Verified the rule allows `import type {Flags} from 'lighthouse'` (type-only import)
- [x] Verified the rule does not fire inside `src/third_party/index.ts` (the barrel itself)
- [x] Verified scripts/ and tests/ are unaffected (rule scoped to `src/**/*.ts`)


zyzyzyryxy:
Thank you for your contribution!

I see you hardcoded the list of bundled packages, I am thinking if we could get that list dynamically.

We generate build/src/third_party/bundled-packages.json during bundling, but that we cannot rely on in eslint. We could maybe have a test to check if they are in sync instead.

Could eslint parse src/third_party/*.ts for imports there to populate that list? Not sure if that could work.

In both cases, we would need some special case handling around puppeteer anyways.

If either of those is impractical, certainly having hardcoded list is still better than nothing.

mvanhorn:
Replaced hardcoded list with `discoverBundledPackages()` that reads `src/third_party/*.ts` at ESLint load time and extracts package names from import statements. Correctly discovers all 8 packages including scoped packages and side-effect imports.

scripts/eslint_rules/no-direct-third-party-imports-rule.js
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * See https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1123
+ */
+
+import {readdirSync, readFileSync} from 'node:fs';
+import {join, dirname} from 'node:path';
+import {fileURLToPath} from 'node:url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const THIRD_PARTY_DIR = join(__dirname, '..', '..', 'src', 'third_party');
```
Lightning00Blade: I think we can safely use `import.meta.dirname`

scripts/eslint_rules/no-direct-third-party-imports-rule.js
```diff
@@ -0,0 +1,138 @@
...
+    },
+    schema: [],
+    messages: {
+      noDirectImport:
+        'Do not import "{{source}}" directly. Use the re-export from "src/third_party/index.js" instead so the import survives bundling. (Type-only imports are fine.)',
+    },
+  },
+  defaultOptions: [],
+  create(context) {
+    const filename = context.getFilename();
```
Lightning00Blade: I think this is deprecated and removed in EsLint 10.
https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/#context-methods-becoming-properties

eslint.config.mjs
```diff
@@ -135,6 +135,13 @@ export default defineConfig([
     },
   },
   {
+    name: 'Bundle import guard',
```
Lightning00Blade: Nit
```suggestion
    name: 'Source files',
```

mvanhorn:
Thanks for the merge!

mvanhorn:
Thanks for the quick review and merge!

PR #1191, opened 2026-03-15
chore: add eslint rule to validate tool schema patterns

### LINKED ISSUE #1076: Create eslint for validating schema used in tools
We should create an eslint to disallow `.nullable` and `.object` attributes.

See for more context https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1073#discussion_r2872188785

mvanhorn:
## Summary

- Add `@local/no-zod-nullable-object` ESLint rule that disallows `.nullable()` and `.object()` usage in tool schema files (`src/tools/**/*.ts`)
- Fix the existing `zod.object()` violation in `fill_form` by converting the elements array from objects to `"uid=value"` formatted strings
- Remove the TODO from CONTRIBUTING.md and link the rule name to the documented restriction

## Context

Per the [PR #1073 review](https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1073#discussion_r2872188785) and the CONTRIBUTING.md guidelines, tool schemas should not use `.nullable()` or `.object()` types. Complex objects should be represented as short formatted strings instead.

The rule catches:
- `zod.object({...})` / `z.object({...})` - flags direct zod object schema usage
- `.nullable()` - flags nullable schema usage on any expression

Closes #1076

## Test plan

- [ ] `npx eslint src/tools/` passes with no violations
- [ ] Creating a test file with `zod.object()` or `.nullable()` in `src/tools/` triggers the rule
- [ ] TypeScript compilation passes (`npx tsc --noEmit`)
- [ ] Existing `fill_form` test updated to use new string format

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -138,7 +138,6 @@ export default defineConfig([
   {
     name: 'Tools definitions',
     files: ['src/tools/**/*.ts'],
-    ignores: ['src/tools/input.ts'],
     rules: {
       '@local/enforce-zod-schema': 'error',
     },
--- a/src/tools/input.ts
+++ b/src/tools/input.ts
@@ -322,6 +322,7 @@ export const fillForm = definePageTool({
   schema: {
     elements: zod
       .array(
+        // eslint-disable-next-line @local/enforce-zod-schema
         zod.object({
           uid: zod.string().describe('The uid of the element to fill out'),
           value: zod.string().describe('Value for the element'),


wolfib:
Thank you! 

The lint rule LGTM. I don't feel comfortable with the API change to `fill_form` (yet), we probably want to evaluate with actual agents that it doesn't cause them any problems, etc.

How about not touching the `fill_form` tool in this PR and instead excluding the corresponding file from being checked against the new lint rule?

scripts/eslint_rules/no-zod-nullable-object-rule.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
Lightning00Blade: Nit: 2026

scripts/eslint_rules/no-zod-nullable-object-rule.js
```diff
@@ -0,0 +1,59 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export default {
+  name: 'no-zod-nullable-object',
```
Lightning00Blade: Nit: Let's call it something like - `enforce-zod-schema`, as we may add/remove things from it.

scripts/eslint_rules/enforce-zod-schema-rule.js
```diff
@@ -0,0 +1,59 @@
...
+          node.callee.type !== 'MemberExpression' ||
+          node.callee.property.type !== 'Identifier'
+        ) {
+          return;
+        }
+
+        const methodName = node.callee.property.name;
+
+        if (methodName === 'nullable') {
+          context.report({
```
Lightning00Blade, line 40: Let's add a comment here explaining that we don't validate that it's called on a `ZodObject` as this should catch arbitrary `.nullable()` calls.

mvanhorn:
Reverted fill_form changes and excluded `src/tools/input.ts` from the lint rule via `ignores` in eslint config. Also addressed @Lightning00Blade's inline comments: renamed rule file, fixed copyright year, added comment about `.nullable()` scope.

eslint.config.mjs
```diff
@@ -134,6 +134,14 @@ export default defineConfig([
       ],
     },
   },
+  {
+    name: 'Tool schema restrictions',
+    files: ['src/tools/**/*.ts'],
+    ignores: ['src/tools/input.ts'],
```
Lightning00Blade: Let's have this ignore inline as those are easier to spot when doing code reviews and/or reading code.

eslint.config.mjs
```diff
@@ -134,6 +134,14 @@ export default defineConfig([
       ],
     },
   },
+  {
+    name: 'Tool schema restrictions',
```
Lightning00Blade: Let's not over specify and just say which files it affects.
```suggestion
    name: 'Tools definitions',
```

OrKoN:
Thanks @mvanhorn !

mvanhorn:
Appreciate the review!

mvanhorn:
Appreciate the fast turnaround on this one too!

Issue #1192, opened 2026-03-16
Memory leak: chrome-devtools-mcp grows ~13 MB/min in --autoConnect mode, causes OOM


dzianisv:
### Environment
- chrome-devtools-mcp@0.19.0 via `npm exec`
- macOS 15, Chrome 146
- Running as MCP server under opencode serve (long-lived)

### Observed behavior

After ~10 hours of idle operation, `chrome-devtools-mcp --autoConnect --channel=stable` consumed **1,661 MB RSS** and contributed to a macOS kernel watchdog panic (swap exhaustion at 8.4 GB total process tree).

Live measurement today after ~9h41m uptime:

```
PID    RSS      VSZ     ELAPSED COMMAND
3106   303248   418351872  09:41:18 node chrome-devtools-mcp --autoConnect
```

303 MB RSS after 9.7 hours = growth rate of ~0.5 MB/min in today's session.

In a previous overnight session, the growth rate was steeper (~13 MB/min), reaching 1.66 GB in ~2 hours when Chrome had more active tabs.

### Evidence from memory monitor

Our opencode memory monitor captured hourly snapshots. At the time of the crash:

| Time | Tree RSS | Children | chrome-devtools-mcp RSS |
|------|----------|----------|------------------------|
| 21:00 | 3,816 MB | 42 | 652 MB |
| 22:00 | 5,612 MB | 55 | 1,126 MB |
| 23:00 | 8,472 MB | 55 | 1,661 MB |
| 07:33 | kernel panic (watchdog timeout, LOW swap space) | | |

### Expected behavior

An MCP server should have stable memory consumption when idle. A long-running process should not grow unboundedly.

### Workaround

We added MCP idle sweep logic in the host (opencode) to periodically close MCP clients that haven't been used within a configurable window. This terminates the leaking process, but it would be better to fix the leak at the source.

### Reproduction

1. Start `chrome-devtools-mcp --autoConnect --channel=stable` as an MCP server
2. Open Chrome with ~10-20 tabs
3. Use the MCP tools occasionally over several hours
4. Monitor RSS with `ps -p <PID> -o rss` periodically

RSS will grow continuously without bound. The growth rate correlates with the number of active Chrome tabs and DevTools protocol events.

### Related

- opencode upstream issue: https://github.com/anomalyco/opencode/issues/16697
- Full forensics: https://github.com/dzianisv/opencode/blob/dev/docs/memory-forensics.md

[...]

PR #1200, opened 2026-03-17
perf(memory): release old navigation request in NetworkCollector

flobo3:
Issue #1192

## Problem
In `--autoConnect` mode, `chrome-devtools-mcp` experiences a severe memory leak (~13 MB/min) when Chrome is actively used. This eventually leads to OOM crashes or kernel panics.

The root cause is in `src/PageCollector.ts`:
1. `NetworkCollector` overrides `splitAfterNavigation` but forgets to call `navigations.splice(this.#maxNavigationSaved)`, causing the array of navigations to grow infinitely.
2. Even within a single navigation (e.g., in long-lived SPA applications), the `navigations[0]` array grows infinitely because there is no limit on the number of items collected per navigation. Since `HTTPRequest` objects are heavy, this quickly exhausts memory.

## Solution
1. Changed `#maxNavigationSaved` to `protected maxNavigationSaved` so subclasses can access it.
2. Added `navigations.splice(this.maxNavigationSaved)` to `NetworkCollector.splitAfterNavigation` to ensure old navigations are properly discarded.
3. Introduced `protected maxItemsPerNavigation = 5000` to limit the number of items stored per navigation. When the limit is reached, the oldest item is removed (`shift()`), preventing unbounded memory growth in SPAs.

Tested locally by running `npm run build` and verifying the fix.

### DIFF
--- a/tests/PageCollector.test.ts
+++ b/tests/PageCollector.test.ts
@@ -303,9 +303,9 @@ describe('NetworkCollector', () => {
       page.emit('framenavigated', mainFrame);
     }
 
-    // We expect 4 arrays in navigations (current + 3 saved)
-    // Each navigation has 1 request, so total should be 4
-    assert.equal(collector.getData(page, true).length, 4);
+    // We expect 3 arrays in navigations (current + 2 saved)
+    // Each navigation has 1 request, so total should be 3
+    assert.equal(collector.getData(page, true).length, 3);
   });
 });
 


wolfib:
Thank you!

Adding the missing `splice(..)` to the `NetworkCollector` LGTM. I'm not so sure about the `maxItemsPerNavigation` change, is it possible that this throws away data we actually want to keep? Here are my suggestions:

- Create a separate PRe for the introduction of `maxItemsPerNavigation`.
- In this PR, add a test to `PageCollector.test.ts` which verifies that `NetworkCollector` doesn't grow beyond `maxNavigationSaved`
- Ensure that `package-lock.json` is not changed (`npm ci` instead of `npm i` might help).

flobo3:
Thanks for the review!

I have updated the PR according to your suggestions:
1. Reverted the `maxItemsPerNavigation` change (I will create a separate PR for it).
2. Added `navigations.splice(3)` to `NetworkCollector` (hardcoded to 3 to match the private `#maxNavigationSaved` in the base class).
3. Added a test in `PageCollector.test.ts` to verify that `NetworkCollector` doesn't grow beyond `maxNavigationSaved`.
4. Reverted `package-lock.json` changes.

Let me know if anything else is needed!

wolfib:
Thanks! Just 1 more nit: could you please go back to `maxNavigationSaved` being `protected` instead of hardcoding it in the derived class?

flobo3:
Done! Changed `#maxNavigationSaved` to `protected maxNavigationSaved` and updated the derived class to use it instead of the hardcoded value.

wolfib:
Cool! Now only the test needs fixing.

src/PageCollector.ts
```diff
@@ -159,7 +159,7 @@ export class PageCollector<T> {
     }
     // Add the latest navigation first
     navigations.unshift([]);
-    navigations.splice(this.#maxNavigationSaved);
+    navigations.splice(this.maxNavigationSaved + 1);
```
Lightning00Blade: This should stay as it is. We want to store the data for the last 3 navigations including the current one.

flobo3:
Done! I've reverted the `+ 1` back to `this.maxNavigationSaved` in both places and updated the test to expect 3 navigations instead of 4. Thanks for the clarification!

PR #1208, opened 2026-03-19
feat: record client name in telemetry.

yulunz:
The changes in the PR intercepts the name of the MCP client right after server initialization, coverts it into one of the existing enum values, and append it with the telemetry request.

Note that the client name is unset if there is no client connected yet. This means the server_start event won't have a client name.

Tested with gemini-cli.

BEGIN_COMMIT_OVERRIDE
fix(telemetry): record client name
END_COMMIT_OVERRIDE

### DIFF
--- a/src/telemetry/ClearcutLogger.ts
+++ b/src/telemetry/ClearcutLogger.ts
@@ -10,7 +10,12 @@ import {logger} from '../logger.js';
 
 import type {LocalState, Persistence} from './persistence.js';
 import {FilePersistence} from './persistence.js';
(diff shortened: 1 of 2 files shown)


Lightning00Blade:
Super nit: this should have been marked as a `fix` (if we want to make a new release for it) or even `chore` as it's not an user facing feature per say, but something for us.

You can use the following to update it after the fact - https://github.com/googleapis/release-please?tab=readme-ov-file#how-can-i-fix-release-notes

PR #1209, opened 2026-03-20
feat: add set_extra_http_headers tool for Network.setExtraHTTPHeaders

### LINKED ISSUE #1175: Feature request: set_extra_http_headers tool for Network.setExtraHTTPHeaders

JayeeHsu:
## Summary
- Add a new MCP tool `set_extra_http_headers` that allows setting extra HTTP headers on all page requests via Puppeteer's `page.setExtraHTTPHeaders()`
- Headers persist across navigations and can be cleared by passing an empty object
- This addresses the need for injecting custom headers (e.g. swim-lane routing, A/B testing, feature flags) on all request types including initial document loads

Closes #1175

## Test plan
- [x] Test setting multiple headers returns correct count and header names
- [x] Test clearing headers with empty object returns clear confirmation
- [x] Test headers are actually included in subsequent requests (end-to-end via echo server)

### DIFF
--- a/src/tools/network.ts
+++ b/src/tools/network.ts
@@ -86,6 +86,34 @@ export const listNetworkRequests = definePageTool({
   },
 });
 
+export const setExtraHttpHeaders = definePageTool({
+  name: 'set_extra_http_headers',
+  description: `Set extra HTTP headers that will be included in every request the page makes. These headers are applied to all resource types including document, script, stylesheet, image, fetch, and XHR requests. The headers persist across navigations until explicitly cleared by calling this tool with an empty headers object.`,
+  annotations: {
+    category: ToolCategory.NETWORK,
+    readOnlyHint: false,
+  },
+  schema: {
+    headers: zod
+      .record(zod.string(), zod.string())
+      .describe(
+        'HTTP headers as key-value pairs to include in every request. Pass an empty object {} to clear previously set headers.',
+      ),
+  },
+  handler: async (request, response) => {
+    const page = request.page;
+    await page.pptrPage.setExtraHTTPHeaders(request.params.headers);
+    const count = Object.keys(request.params.headers).length;
+    if (count === 0) {
+      response.appendResponseLine('Cleared all extra HTTP headers.');
+    } else {
+      response.appendResponseLine(
+        `Set ${count} extra HTTP header(s): ${Object.keys(request.params.headers).join(', ')}`,
+      );
+    }
+  },
+});
+
 export const getNetworkRequest = definePageTool({
   name: 'get_network_request',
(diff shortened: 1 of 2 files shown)


JayeeHsu:
I have signed the CLA.

JayeeHsu:
@wolfib  Could you help review this?

OrKoN (changes requested):
Thanks for the PR! Please align the implementation with the comment in https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1175#issuecomment-4097587153 To also work better with the CLI interface, perhaps we can have a string format for expressing headers.

OrKoN:
Closing in favour of earlier https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1176

Issue #1214, opened 2026-03-21
Bug: Memory leak in autoConnect mode - PageCollector not cleaning up stale entries


Financier-Nuri:
### Bug Type
Memory leak / Resource management bug

### Summary

In `--autoConnect` mode, the `PageCollector` class accumulates stale page entries over time without proper cleanup. Each reconnection creates new CDP sessions but the old page references are never removed from `self._pages`, causing unbounded memory growth (~13 MB/min reported).

### Environment

- chrome-devtools-mcp version: Latest (via npm)
- Chrome DevTools Protocol
- Node.js: 18+
- OS: macOS/Linux

### Steps to reproduce

1. Start chrome-devtools-mcp with `--autoConnect` flag
2. Leave it running for extended period with active browser usage
3. Monitor memory usage: `ps aux | grep chrome-devtools-mcp` or `process.memoryUsage()`
4. Observe continuous memory growth without stabilization

### Expected behavior

Memory usage should stabilize after initial CDP session setup, with only minor fluctuations during normal operation.

### Actual behavior

Memory grows continuously at approximately 13 MB/minute in active browser environments. After ~1 hour, the process may hit OOM limits.

### Root Cause Analysis

In `src/PageCollector.ts`:

```typescript
export class PageCollector {
  private _pages: Map<string, PageInfo> = new Map();
  
  // New page is added but never removed properly
  async addPage(page: PageInfo): Promise<void> {
    // Problem: Only adds, no cleanup of stale entries
    this._pages.set(page.targetId, page);
  }
  
  // When a target is destroyed, the page info remains in the map
  async removePage(targetId: string): Promise<void> {
    // Issue: This is called but the Map keeps growing due to
    // orphaned references from reconnection events
    this._pages.delete(targetId);
  }
}
```

The race condition occurs because:
1. `addPage()` is called when CDP connects to a new target
2. Before `removePage()` is called, a reconnection event may fire
3. New page info is added with the same targetId (from reconnect)
4. The old entry becomes orphaned but still in memory

### Affected Code Location

- File: `src/PageCollector.ts`
- Method: `addPage()`, `removePage()`
- Issue: No cleanup mechanism for orphaned entries during reconnection

### Suggested Fix

```typescript
export class PageCollector {
  private _pages: Map<string, PageInfo> = new Map();
  private _targetIdToConnectionId: Map<string, string> = new Map();
  
[...]

PR #1221, opened 2026-03-22
feat(collector): add maxItemsSavedPerNavigation to prevent unbounded memory growth

gilangjavier:
Issue #1214.

This introduces a configurable limit on the number of items collected per navigation (current navigation slice) in PageCollector. Without this, long-running pages that continuously emit network requests or console messages (e.g., single-page apps, polling) could cause unbounded memory growth. The limit is only applied to the *current* navigation; historic navigations remain capped by maxNavigationSaved.

- Add protected  default  in PageCollector
- Override to 5000 in NetworkCollector and 1000 in ConsoleCollector
- Add  that splices oldest items when limit exceeded\n- Add unit test for pruning behavior
- This complements the existing navigation count limit (maxNavigationSaved) and addresses the memory leak observed in --autoConnect mode (~13 MB/min).

### DIFF
--- a/src/PageCollector.ts
+++ b/src/PageCollector.ts
@@ -65,6 +65,14 @@ export class PageCollector<T> {
   protected maxNavigationSaved = 3;
 
   /**
+   * Max number of collected items to keep for the *current* navigation.
+   *
+   * This is a safety valve to prevent unbounded memory growth in long-running
+   * sessions (e.g. autoConnect mode) on highly active pages.
+   */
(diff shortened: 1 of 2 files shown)


Lightning00Blade:
I think this is fixed by https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2435,
And the Console is tracked at https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2768
I will close this for now.

PR #1237, opened 2026-03-25
docs: provide disclaimer about supported browsers

Lightning00Blade:
We already provide Requirements section that states that Chrome is required. 
This should make it explicit what version we are targeting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ allowing them to inspect, debug, and modify any data in the browser or DevTools.
 Avoid sharing sensitive or personal information that you don't want to share with
 MCP clients.
 
+`chrome-devtools-mcp` officially supports Google Chrome and [Chrome for Testing](https://developer.chrome.com/blog/chrome-for-testing/) only.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,7 +57,7 @@ Collection is disabled if CHROME_DEVTOOLS_MCP_NO_USAGE_STATISTICS or CI env vari
 
 - [Node.js](https://nodejs.org/) v20.19 or a newer [latest maintenance LTS](https://github.com/nodejs/Release#release-schedule) version.
 - [Chrome](https://www.google.com/chrome/) current stable version or newer.
-- [npm](https://www.npmjs.com/).
+- [npm](https://www.npmjs.com/)
```
OrKoN, line 60: dot was removed only on the last list item? Should we move the disclaimer out of the disclaimers and into the point 2 which lists Chrome as the system requirement. I think we also do not actually support or encourage the use of Chrome for Testing for the MCP server? cc @natorion

Lightning00Blade: For the dot question - other places are sentences, but maybe we should remove it there as well.

PR #1239, opened 2026-03-25
chore: Add a new tool for listing in-page tools (gated behind a command line flag)

wolfib:
This adds a `list_in_page_tools` MCP tool. When called, it dispatches a `devtoolstooldiscovery` event on the active page. The page announces its exposing tools by calling the event's `respondWith` method, which causes the exposed tools to be stashed on the page's `window` object. This list of in-page tools is then appended to the `list_in_page_tools` response.

Calling the exposed in-page tools from the MCP server will be handled in a follow-up.

BEGIN_COMMIT_OVERRIDE
chore: Add a new tool for listing in-page tools (gated behind a command line flag)
END_COMMIT_OVERRIDE

### DIFF
--- a/README.md
+++ b/README.md
@@ -475,8 +475,6 @@ If you run into any issues, checkout our [troubleshooting guide](./docs/troubles
   - [`list_console_messages`](docs/tool-reference.md#list_console_messages)
   - [`take_screenshot`](docs/tool-reference.md#take_screenshot)
   - [`take_snapshot`](docs/tool-reference.md#take_snapshot)
-- **In-page tools** (1 tools)
-  - [`list_in_page_tools`](docs/tool-reference.md#list_in_page_tools)
 
 <!-- END AUTO GENERATED TOOLS -->
 
@@ -567,11 +565,6 @@ The Chrome DevTools MCP server supports the following configuration option:
   Set to false to exclude tools related to network.
   - **Type:** boolean
   - **Default:** `true`
-
-- **`--categoryInPageTools`/ `--category-in-page-tools`**
-  Set to true to enable tools exposed by the inspected page itself
-  - **Type:** boolean
-  - **Default:** `false`
 
 - **`--performanceCrux`/ `--performance-crux`**
   Set to false to disable sending URLs from performance traces to CrUX API to get field performance data.
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -37,8 +37,6 @@
   - [`list_console_messages`](#list_console_messages)
   - [`take_screenshot`](#take_screenshot)
   - [`take_snapshot`](#take_snapshot)
-- **[In-page tools](#in-page-tools)** (1 tools)
-  - [`list_in_page_tools`](#list_in_page_tools)
 
 ## Input automation
 
@@ -397,17 +395,5 @@ in the DevTools Elements panel (if any).
 
 - **filePath** (string) _(optional)_: The absolute path, or a path relative to the current working directory, to save the snapshot to instead of attaching it to the response.
 - **verbose** (boolean) _(optional)_: Whether to include all possible information available in the full a11y tree. Default is false.
-
----
-
-## In-page tools
-
-### `list_in_page_tools`
-
-**Description:** Lists all in-page-tools the page exposes for providing runtime information.
-In-page-tools are exposed on the page via the 'window.\_\_dtmcp.executeTool(toolName, params)'
-function where they can be called by '[`evaluate_script`](#evaluate_script)'.
-
-**Parameters:** None
 
 ---


src/tools/inPage.ts
```diff
@@ -0,0 +1,50 @@
...
+import {type JSONSchema7} from '../third_party/index.js';
+
+import {ToolCategory} from './categories.js';
+import {definePageTool} from './ToolDefinition.js';
+
+export interface ToolDefinition {
+  name: string;
+  description: string;
+  inputSchema: JSONSchema7;
+  execute: (input: Record<string, unknown>) => unknown;
```
Lightning00Blade: Given we are using `evaluate`, we can't have non JSON values in this object.
I think we should split it.
We can have ToolGroup accept a generic that by default - without a execute and the in the Window extension add it there.

src/tools/inPage.ts
```diff
@@ -0,0 +1,50 @@
...
+        args: Record<string, unknown>,
+      ) => unknown;
+    };
+  }
+}
+
+export const listInPageTools = definePageTool({
+  name: 'list_in_page_tools',
+  description: `Lists all in-page-tools the page exposes for providing runtime information.
+  In-page-tools are exposed on the page via the 'window.__dtmcp.executeTool(toolName, params)'
```
Lightning00Blade: More of an open question rather then a request for change:
Would it be better to say something like:
`To call the tools call X with Y`.

src/McpResponse.ts
```diff
@@ -40,6 +41,57 @@ interface TraceInsightData {
...
+              throw new Error(`Tool ${toolName} not found`);
+            }
+            return await tool.execute(args);
+          };
+        }
+
+        resolve(toolGroup);
+      };
+      window.dispatchEvent(event);
+      // If the page does not call `event.respondWith`, return instead of timing out
```
Lightning00Blade: I was a little bit confused, but then I re-read that `dispatchEvent` is sync so maybe updating the comment here to reflect that would be nice for whenever someone else reads the code.

tests/tools/inPage.test.ts
```diff
@@ -0,0 +1,91 @@
...
+            ],
+          };
+
+          await page.pptrPage.evaluate(() => {
+            window.addEventListener('devtoolstooldiscovery', () => {
+              // No-op
+            });
+          });
+
+          const evaluateStub = sinon.stub(page.pptrPage, 'evaluate');
```
Lightning00Blade: Can't we just do 
```ts
page.evaluate(() => {
window.__dtmcp.toolGroup = {}
})
```
That way we verify serialization though CDP does not cause issue.

tests/tools/inPage.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      await withMcpContext(
+        async (response, context) => {
+          const page = await context.newPage();
+          await page.pptrPage.evaluate(() => {
+            window.addEventListener('devtoolstooldiscovery', () => {
+              // No-op
+            });
+          });
+
+          const evaluateStub = sinon.stub(page.pptrPage, 'evaluate');
```
Lightning00Blade: Same here.

tests/index.test.ts
```diff
@@ -72,47 +72,52 @@ describe('e2e', () => {
...
           }
         }
-      }
-      definedNames.sort();
-      assert.deepStrictEqual(exposedNames, definedNames);
-    });
+        definedNames.sort();
+        assert.deepStrictEqual(exposedNames, definedNames);
+      },
+      ['--category-in-page-tools'],
```
Lightning00Blade, line 127: We should provide Conditions and add a new test for this case.

tests/McpResponse.test.ts
```diff
@@ -1013,3 +1016,79 @@ describe('lighthouse', () => {
     });
   });
 });
+
+describe('in-page tools', () => {
```
Lightning00Blade:
```suggestion
describe('inPage tools', () => {
```
Let's use the filename.

src/McpResponse.ts
```diff
@@ -357,6 +416,12 @@ export class McpResponse implements Response {
     if (this.#listExtensions) {
       extensions = context.listExtensions();
     }
+
+    let inPageTools: ToolGroup | undefined;
+    if (this.#listInPageTools) {
+      inPageTools = await getToolGroup(context.getSelectedMcpPage());
```
Lightning00Blade, line 424: Can you double check how this works with `experimentalPageIdRouting`?
If it does not we should leave a TODO here for it.

Lightning00Blade (commented):
Over all looks good with some comments.

src/McpResponse.ts
wolfib: Verified manually that it works in combination with `experimentalPageIdRouting`

wolfib:
Addressed the comments. PTAL again.

PR #1250, opened 2026-03-27
chore: add function to sanitize params for tool calls

yulunz:
This adds functions to sanitize the tool call parameters. They are not called as of now since we don't have server side changes landed yet to support these.


nattallius:
@yulunz  can you resolve all format checks?

yulunz:
> @yulunz can you resolve all format checks?

hi, of course, should have marked this as a draft. I was thinking like in the CL world :)

PR #1252, opened 2026-03-27
chore: add update notification to both binaries

mathiasbynens:
Both `chrome-devtools` and `chrome-devtools-mcp` now log a notification when a newer version is detected to be available.

Issue: https://crbug.com/496757598

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "@types/filesystem": "^0.0.36",
     "@types/node": "^25.0.0",
     "@types/sinon": "^21.0.0",
+    "@types/update-notifier": "^6.0.8",
     "@types/yargs": "^17.0.33",
     "@typescript-eslint/eslint-plugin": "^8.43.0",
     "@typescript-eslint/parser": "^8.43.0",
@@ -80,5 +81,8 @@
   },
   "engines": {
     "node": "^20.19.0 || ^22.12.0 || >=23"
+  },
+  "dependencies": {
+    "update-notifier": "^7.3.1"
   }
 }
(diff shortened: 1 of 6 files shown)


mathiasbynens:
> Two things:
> 
> * Should we wrap it in a helper to reuse same code?
> * Should we use `shouldNotifyInNpmScript ` as we use `npx`?

Both done. PTAL. Also looping in @OrKoN.

I don’t like that this adds a dependency. cc @natorion

src/bin/chrome-devtools.ts
```diff
@@ -19,11 +19,14 @@ import {
 import {isDaemonRunning, serializeArgs} from '../daemon/utils.js';
 import {logDisclaimers} from '../index.js';
 import {hideBin, yargs, type CallToolResult} from '../third_party/index.js';
+import {notifyUpdate} from '../utils/update-notifier.js';
 import {VERSION} from '../version.js';
 
 import {commands} from './chrome-devtools-cli-options.js';
 import {cliOptions, parseArguments} from './chrome-devtools-mcp-cli-options.js';
 
+notifyUpdate();
```
OrKoN: can we customize the message to indicate that the daemon has to be re-started? does it output to stdout or stderr?

mathiasbynens: It outputs to stdout. Done; PTAL

package.json
```diff
@@ -80,5 +81,8 @@
   },
   "engines": {
     "node": "^20.19.0 || ^22.12.0 || >=23"
+  },
+  "dependencies": {
+    "update-notifier": "^7.3.1"
```
OrKoN, line 86: this needs to be a dev-dependency, re-exported in src/third_party/index.ts so that it is bundled

mathiasbynens: Capturing offline discussion: alternatively, we can just request `https://registry.npmjs.org/chrome-devtools-mcp/latest` (failing silently in case of any errors) and compare the version number. If the request succeeds and the version number is different, we log the message. I vastly prefer this approach as it avoids transitive dependencies and complications.

mathiasbynens: Alternate approach: https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1783/changes

mathiasbynens:
Closing in favor of #1783.

PR #1253, opened 2026-03-27
chore: remove specific chrome executable usage on extensions tests

nroscino:
This change removes the download and usage of the version 146.0.7680.31. 
The version was previously needed to get the triggerAction working in extensions. This is no longer needed since the current stable now has the CDP command available.

### DIFF
--- a/scripts/test.mjs
+++ b/scripts/test.mjs
@@ -92,9 +92,6 @@ async function runTests(attempt) {
   });
 }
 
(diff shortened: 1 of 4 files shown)


nattallius (approved):

nroscino:
Closed with https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1911

PR #1257, opened 2026-03-28
chore: add a script to generate tool_call_metrics.json for telemetry.

yulunz:
This adds a script that generates a json file that summarizes all tool calls and arguments for each.

- The arguments run through the blocklist filtering so arguments containing high entropy ids are filtered out (e.g. "uid", "msgid" etc).
- It uses existing functions from clearcut logger module to transform the arg name and value (i.e. take the length of the string, take the size of the array, and rename the string to be "string_length", and array "array_count" etc).
- These functions from the clearcut logger module will be later used to sanitize the params as we start to log them.

This doesn't include the append only / deprecation logic just yet (i.e. it's doesn't handle the case when new tools are added / removed, or arguments of the existing tools are modified). This will be added in following PRs.

The parent PR is #1250.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "typecheck": "tsc --noEmit",
     "format": "eslint --cache --fix . && prettier --write --cache .",
     "check-format": "eslint --cache . && prettier --check --cache .;",
-    "gen": "npm run build && npm run docs:generate && npm run cli:generate && npm run format",
+    "gen": "npm run build && npm run docs:generate && npm run cli:generate && npm run update-tool-call-metrics && npm run format",
     "docs:generate": "node --experimental-strip-types scripts/generate-docs.ts",
     "start": "npm run build && node build/src/index.js",
     "start-debug": "DEBUG=mcp:* DEBUG_COLORS=false npm run build && node build/src/index.js",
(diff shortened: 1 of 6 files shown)


tests/telemetry/toolMetricsUtils.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+import {
+  generateToolMetrics,
+  validateEnumHomogeneity,
+} from '../../src/telemetry/toolMetricsUtils.js';
+import {zod} from '../../src/third_party/index.js';
+import {ToolCategory} from '../../src/tools/categories.js';
+import type {ToolDefinition} from '../../src/tools/ToolDefinition.js';
+
+test('validateEnumHomogeneity should return the primitive type of a homogeneous enum', () => {
```
OrKoN: we use `describe/it` interfaces in the rest of the tests, let's keep using them for consistency.

package.json
```diff
@@ -27,6 +27,7 @@
     "prepare": "node --experimental-strip-types scripts/prepare.ts",
     "verify-server-json-version": "node --experimental-strip-types scripts/verify-server-json-version.ts",
     "update-lighthouse": "node --experimental-strip-types scripts/update-lighthouse.ts",
+    "update-tool-call-metrics": "node --experimental-strip-types scripts/update_tool_call_metrics.ts",
```
OrKoN, line 30: let's add it to the "gen" command ensuring the data is kept up to date.

tests/telemetry/toolMetricsUtils.test.ts
yulunz: yeah, done.

package.json
yulunz: done.

yulunz:
I am roughly still having this like I had it initially in order not to make it too complicated. Follow-up CLs will make this look like what we have discussed.

PR #1258, opened 2026-03-28
fix(cli): avoid defaulting to isolated when userDataDir is provided

pidofme:
## Summary

   Fix `chrome-devtools start` so it no longer implicitly enables `isolated` when `--userDataDir` is provided.

   Previously, the CLI wrapper always defaulted `isolated` to `true` for `start`, which caused `userDataDir` and `isolated` to conflict even when the user only specified `--userDataDir`. This made it impossible to start the CLI daemon against a persistent browser profile.

   ## Changes

   - Update `chrome-devtools start` default handling in `src/bin/chrome-devtools.ts`
     - only default `isolated=true` when `userDataDir` is not set
   - Clarify the `isolated` CLI description to document the conditional default
   - Update `docs/cli.md` to reflect that:
     - `headless` is enabled by default
     - `isolated` is enabled by default unless `--userDataDir` is provided
   - Fix a small error message typo

   ## Why

   This matches the intended semantics of the flags:

   - `--isolated` means use a temporary user data dir
   - `--userDataDir` means use a persistent, explicit user data dir

   If the user passes `--userDataDir`, the CLI should not also implicitly enable `isolated`.

   ## Testing

   - Ran:
     - `npm test -- tests/cli.test.ts`
     - `npm test -- tests/e2e/chrome-devtools.test.ts`

   Added an e2e regression test in `tests/e2e/chrome-devtools.test.ts` to verify that:

   - `chrome-devtools start --userDataDir <temp dir>` succeeds
   - the CLI no longer fails with `Arguments userDataDir and isolated are mutually exclusive`
   - the daemon starts successfully when `userDataDir` is provided

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -17,7 +17,7 @@ The CLI acts as a client to a background `chrome-devtools-mcp` daemon (uses Unix
 
 - **Automatic Start**: The first time you call a tool (e.g., `list_pages`), the CLI automatically starts the MCP server and the browser in the background if they aren't already running.
 - **Persistence**: The same background instance is reused for subsequent commands, preserving the browser state (open pages, cookies, etc.).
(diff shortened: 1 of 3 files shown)


OrKoN:
Thanks!

PR #1259, opened 2026-03-30
fix: support multi-file uploads in upload_file

### LINKED ISSUE #1217: Support multi-file uploads for existing-session mode
### Summary

When using the `existing-session` browser mode, file uploads currently only support a single file at a time.

### Error message

```text
existing-session file uploads currently support one file at a time.

### Reproduction

Attach MCP to an existing Chrome session
Navigate to a page with input[type="file"][multiple]
Confirm the input supports multiple files
Attempt to upload more than one file in a single operation
Observe the error:
existing-session file uploads currently support one file at a time.


### Expectation

If a page contains a file input with the multiple attribute, the MCP upload API should support uploading multiple files in a single call when attached to an existing browser session.

For example, this should work for:

input[type="file"][multiple]
workflows where the site expects all selected files to be set on the same input at once
cases where sequential single-file uploads overwrite the previous selection instead of appending

Why this matters
Many modern sites support multi-file uploads through a hidden input[type="file"] element with multiple=true. In these cases, automations often need to set all files at once.

Single-file-only behavior is not sufficient because:

the browser automation agent cannot interact with the native OS file picker to manually multi-select files
repeated single-file uploads may replace the previous file selection
some upload components require all files to be provided in one operation

### MCP configuration

_No response_

0.20.3

### Chrome version

_No response_

### Coding agent version

_No response_

### Model version

_No response_

### Chat log

_No response_

### Node version

_No response_

### Operating system

None

### Extra checklist

### PR DESCRIPTION
**Pull request by zerone0x**
## Summary
- allow upload_file to accept multiple file paths via filePaths (keeps filePath for single uploads)
- pass arrays through to Puppeteer uploadFile / fileChooser accept
- document new argument in CLI tool definitions

## Testing
- not run (no test suite for this change)

Fixes #1217

### DIFF
--- a/src/tools/input.ts
+++ b/src/tools/input.ts
@@ -361,17 +361,30 @@ export const uploadFile = definePageTool({
       .describe(
         'The uid of the file input element or an element that will open file chooser on the page from the page content snapshot',
       ),
-    filePath: zod.string().describe('The local path of the file to upload'),
+    filePath: zod
+      .string()
+      .describe('The local path of a file to upload. Use filePaths for multiple files.')
+      .optional(),
+    filePaths: zod
+      .array(zod.string())
+      .describe('One or more local file paths to upload in a single operation.')
+      .optional(),
     includeSnapshot: includeSnapshotSchema,
   },
   handler: async (request, response) => {
-    const {uid, filePath} = request.params;
+    const {uid} = request.params;
+    const filePaths =
+      request.params.filePaths ??
+      (request.params.filePath ? [request.params.filePath] : []);
+    if (!filePaths.length) {
+      throw new Error('Provide filePath or filePaths to upload.');
+    }
     const handle = (await request.page.getElementByUid(
       uid,
     )) as ElementHandle<HTMLInputElement>;
     try {
       try {
-        await handle.uploadFile(filePath);
+        await handle.uploadFile(...filePaths);
       } catch {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/tools/input.ts [hunk 1]
```diff
@@ -361,17 +361,30 @@ export const uploadFile = definePageTool({
       .describe(
         'The uid of the file input element or an element that will open file chooser on the page from the page content snapshot',
       ),
-    filePath: zod.string().describe('The local path of the file to upload'),
+    filePath: zod
```
Lightning00Blade, line 364: I think we should try to leave it single argument and have it separated by some separator like other commands.

Lightning00Blade (changes requested):
Please revert any changes unrelated to the description of the PR (like `protocolTimeout`).

src/tools/input.ts
OrKoN: yes, let's support comma separated values in filePath to be compatible with the CLI.

zerone0x:
updated upload_file to accept comma-separated values in filePath (e.g. --filePath 'a.txt,b.txt'), plus updated arg descriptions. pushed in 0e3862c.

Lightning00Blade:
You will need to fix the CLA issue before we can merge.

OrKoN:
Closing as stale.

Issue #1156, opened 2026-03-10
Network.enable timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.


yanglijun05:
### Description of the bug

Network.enable timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.

### Expectation

enable MCP

### MCP configuration

  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": [
        "chrome-devtools-mcp@latest",
        "--autoConnect",
        "--usageStatistics=false",
        "--performanceCrux=false"
      ]
    }
  }

### Chrome DevTools MCP version

0.19.0

### Node version

v20.19.2

### Operating system

macOS

### Extra checklist

- [ ] I want to provide a PR to fix this bug

zyzyzyryxy:
Hello, thank you for your time reporting the issue.

To properly triage it, we need more details on circumstances you encountered this issue:
- when the error happened? Was it during chrome-devtools-mcp connecting to Chrome?
- How many tabs did you have open in the Chrome instance you were trying to connect to?
- Was this your local Chrome or a remote configuration?
- was the system under a heavy load at the moment?
- Does this error happen frequently in your setup?

One solution that comes to mind would be to add a flag to customize `protocolTimeout`, but that would be a feature request and we need answers to the above to plan it correctly.

grexzen:
This error is happening all the time now. It was working at some point, but today every call shows that same error.

Chrome is local, no heavy load anything, something happened that now all calls are blocked.

zyzyzyryxy:
I have no issues connecting to Chrome using gemini-cli and setup specified in the description. If it happens consistently for you, please provide your Chrome version and share logs from the MCP server by following instructions in https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/docs/troubleshooting.md#debugging

Issue #657, opened 2025-12-09
Network.enable timed out


caijinrui:
### Description of the bug

Running Chrome in --remote-debugging-port mode results in a timeout exception after a period of time (within one day).

Environment
Chrome DevTools MCP version: Latest
Chrome version: Playwright’s built-in chrome-headless-shell (Playwright v1.57.0) 
— Browser Launch Arguments
const CHROME_ARGS = [
  '--headless',
  '--allow-pre-commit-input',
  '--disable-background-networking',
  '--disable-background-timer-throttling',
  '--disable-backgrounding-occluded-windows',
  '--disable-breakpad',
  '--disable-client-side-phishing-detection',
  '--disable-component-extensions-with-background-pages',
  '--disable-crash-reporter',
  '--disable-default-apps',
  '--disable-dev-shm-usage',
  '--disable-hang-monitor',
  '--disable-infobars',
  '--disable-ipc-flooding-protection',
  '--disable-popup-blocking',
  '--disable-prompt-on-repost',
  '--disable-renderer-backgrounding',
  '--disable-search-engine-choice-screen',
  '--disable-sync',
  '--enable-automation',
  '--export-tagged-pdf',
  '--force-color-profile=srgb',
  '--generate-pdf-document-outline',
  '--metrics-recording-only',
  '--no-first-run',
  '--password-store=basic',
  '--use-mock-keychain',
  '--disable-features=Translate,AcceptCHFrame,MediaRouter,OptimizationHints,RenderDocument,ProcessPerSiteUpToMainFrameThreshold,IsolateSandboxedIframes',
  '--enable-features=PdfOopif',
  '--disable-extensions',
  '--hide-crash-restore-bubble',
  '--remote-debugging-port=9222'
];

### Chrome DevTools MCP version

latest

### Chrome version

Google Chrome for Testing 143.0.7499.4

### Operating system

None

### Extra checklist

- [ ] I want to provide a PR to fix this bug

OrKoN:
Thanks for reporting but the MCP is tested with Google Chrome only and might not be compatible with other browsers such as Playwright’s built-in chrome-headless-shell. Can you reproduce the issue with regular Chrome? Also, please provide exact steps to reproduce the issue with Google Chrome.

sjy3:
@caijinrui so how to solve this problem?

PR #1761, opened 2026-03-30
fix: kill stale MCP processes on reconnect to same endpoint

### LINKED ISSUE #1763: multiple MCP processes on same debug port cause Network.enable timeout
## Problem

When MCP clients reconnect (e.g. running `/mcp` in Claude Code, or restarting the client), each reconnect spawns a new `chrome-devtools-mcp` process without killing the previous one. Multiple CDP clients on the same debug port fight over protocol sessions, causing `Network.enable timed out` errors.

## Root cause

There's no mechanism to prevent or clean up duplicate connections to the same browser endpoint. Running `lsof -i :9222` after a few reconnects shows 3-4 node processes all connected to port 9222.

The existing in-process guard in `browser.ts` (`if (browser?.connected)`) only prevents reconnections within the same process. It does nothing for separate processes.

## Reproduction

1. Configure chrome-devtools-mcp with `--browserUrl=http://127.0.0.1:9222`
2. Start an MCP session (connect from Claude Code, Gemini CLI, etc.)
3. Reconnect the MCP session (e.g. `/mcp` in Claude Code)
4. Repeat 2-3 times
5. Run `lsof -i :9222 | grep node` and observe multiple processes
6. Any tool call now fails with `Network.enable timed out`

## Fix

Draft PR: #1761

Adds endpoint-based PID lock files. On startup, the server checks if another instance is already connected to the same endpoint, kills it, waits for it to die, then acquires the lock.

Related: #1156, #657

yuvrajangadsingh:
## Problem

When MCP clients reconnect (e.g. running `/mcp` in Claude Code), each reconnect spawns a new `chrome-devtools-mcp` process without killing the previous one. Multiple CDP clients on the same debug port cause `Network.enable timed out` errors because sessions conflict.

Reproducing is easy: run `/mcp` three times in a row and check `lsof -i :9222`. You'll see 3+ node processes all connected to the same debug port.

## Fix

Adds endpoint-based PID lock files. On startup, the server:

1. Checks if another instance is already connected to the same endpoint
2. Sends SIGTERM (with SIGKILL fallback after 1s)
3. Waits for the old process to actually exit
4. Acquires the lock with its own PID (atomic via `openSync` with `wx` flag)
5. Releases the lock on exit (SIGINT/SIGTERM/exit handlers call `process.exit()`)

Lock files are keyed by normalized endpoint URL (not just port) so different hosts on the same port don't collide. Both `--browserUrl` and `--wsEndpoint` connections are covered.

## Changes

- `src/browser.ts`: `acquireEndpointLock()` and `releaseEndpointLock()` functions
- `src/bin/chrome-devtools-mcp-main.ts`: early lock acquisition on startup, cleanup on exit

## Tested

Verified locally on macOS with Arc browser (port 9222):
- Instance 1 acquires lock, Instance 2 kills Instance 1 and takes over
- Lock file correctly transitions between PIDs
- Old process actually dies (SIGTERM + SIGKILL fallback)
- Lock released on clean exit

Fixes #1763
Related: #1156, #657

### DIFF
--- a/src/bin/chrome-devtools-mcp-main.ts
+++ b/src/bin/chrome-devtools-mcp-main.ts
@@ -8,6 +8,7 @@ import '../polyfill.js';
 
 import process from 'node:process';
 
+import {acquireEndpointLock, releaseEndpointLock} from '../browser.js';
 import {createMcpServer, logDisclaimers} from '../index.js';
 import {logger, saveLogsToFile} from '../logger.js';
 import {computeFlagUsage} from '../telemetry/flagUtils.js';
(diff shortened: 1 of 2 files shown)


yuvrajangadsingh:
closing this. the root cause is frozen/discarded tabs at the Chromium level, not stale MCP processes. there's already a Chromium fix in review (https://chromium-review.git.corp.google.com/c/chromium/src/+/7557410) that will resolve it once it lands in browsers. the lock file approach here was solving a secondary issue.

PR #1766, opened 2026-03-30
chore: Add `execute_in_page_tool` tool

wolfib:
This allows the MCP server to call the in-page tools provided by the inspected page.

Handling of (non-serializable) DOM elements as tool parameters or tool output will be added in follow-ups.

### DIFF
--- a/src/tools/ToolDefinition.ts
+++ b/src/tools/ToolDefinition.ts
@@ -198,7 +198,6 @@ export type Context = Readonly<{
   triggerExtensionAction(id: string): Promise<void>;
   listExtensions(): InstalledExtension[];
   getExtension(id: string): InstalledExtension | undefined;
-  setInPageTools(toolGroup?: ToolGroup<InPageToolDefinition>): void;
   getInPageTools(): ToolGroup<InPageToolDefinition> | undefined;
   getSelectedMcpPage(): McpPage;
   getExtensionServiceWorkers(): ExtensionServiceWorker[];
--- a/src/tools/inPage.ts
+++ b/src/tools/inPage.ts
@@ -66,14 +66,27 @@ export const executeInPageTool = definePageTool({
   schema: {
     toolName: zod.string().describe('The name of the tool to execute'),
     params: zod
-      .record(zod.string(), zod.unknown())
+      .string()
       .optional()
-      .describe('The parameters to pass to the tool'),
+      .describe('The JSON-stringified parameters to pass to the tool'),
   },
   handler: async (request, response, context) => {
     const page = context.getSelectedMcpPage();
     const toolName = request.params.toolName;
-    const params = request.params.params ?? {};
+    let params: Record<string, unknown> = {};
+    if (request.params.params) {
+      try {
+        const parsed = JSON.parse(request.params.params);
+        if (typeof parsed === 'object' && parsed !== null) {
+          params = parsed;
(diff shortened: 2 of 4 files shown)


src/tools/inPage.ts
```diff
@@ -37,9 +37,13 @@ declare global {
...
 export const listInPageTools = definePageTool({
   name: 'list_in_page_tools',
-  description: `Lists all in-page-tools the page exposes for providing runtime information.
-  To call 'list_in_page_tools', call 'evaluate_script' with
-  'window.__dtmcp.executeTool("list_in_page_tools", {})'.`,
+  description: `Lists all in-page tools the page exposes for providing runtime information.
+  In-page tools can be called via the 'execute_in_page_tool()' MCP tool.
+  Alternatively, in-page tools can be executed by calling 'evaluate_script' and adding the
+  following command to the script:
+  'window.__dtmcp.executeTool(toolName, params)'
```
OrKoN, line 44: tbh I am not sure about this (in addition I am not sure it can pass params in evaluate script?). The model will likely have troubles composing scripts correctly.

src/tools/inPage.ts
```diff
@@ -50,3 +54,55 @@ export const listInPageTools = definePageTool({
...
+  name: 'execute_in_page_tool',
+  description: `Executes a tool exposed by the page.`,
+  annotations: {
+    category: ToolCategory.IN_PAGE,
+    readOnlyHint: false,
+    conditions: ['inPageTools'],
+  },
+  schema: {
+    toolName: zod.string().describe('The name of the tool to execute'),
+    params: zod
```
OrKoN, line 68: to make it interoperable with the CLI, should accept JSON stringified params as a string?

src/tools/ToolDefinition.ts
```diff
@@ -194,6 +198,8 @@ export type Context = Readonly<{
   triggerExtensionAction(id: string): Promise<void>;
   listExtensions(): InstalledExtension[];
   getExtension(id: string): InstalledExtension | undefined;
+  setInPageTools(toolGroup?: ToolGroup<InPageToolDefinition>): void;
```
OrKoN: is this used by tools? if not, it should not be here.

wolfib: removed.

src/tools/inPage.ts
```diff
@@ -50,3 +54,68 @@ export const listInPageTools = definePageTool({
...
+      .string()
+      .optional()
+      .describe('The JSON-stringified parameters to pass to the tool'),
+  },
+  handler: async (request, response, context) => {
+    const page = context.getSelectedMcpPage();
```
OrKoN, line 74: forgot to catch this: with experimental pageId routing this should request.page and not `context.getSelectedMcpPage()`

src/McpContext.ts
```diff
@@ -101,6 +102,7 @@ export class McpContext implements Context {
   #screenRecorderData: {recorder: ScreenRecorder; filePath: string} | null =
     null;
 
+  #inPageTools?: ToolGroup<ToolDefinition>;
```
OrKoN, line 105: we should move this to McpPage instead of McpContext to support pageId routing (currently experimental).

OrKoN: please check if other new tools have the same problem

PR #1777, opened 2026-03-31
feat: make pageId required for page-scoped tools by default

bcfmtolgahan:
Removes `--experimental-page-id-routing` making it the default behavior. To go back to the previous behavior, pass `--pageIdRouting=false` when starting the server.

### DIFF
--- a/src/bin/chrome-devtools-mcp-cli-options.ts
+++ b/src/bin/chrome-devtools-mcp-cli-options.ts
@@ -147,10 +147,11 @@ export const cliOptions = {
     type: 'boolean',
     description: `If enabled, ignores errors relative to self-signed and expired certificates. Use with caution.`,
   },
-  experimentalPageIdRouting: {
+  pageIdRouting: {
     type: 'boolean',
     describe:
-      'Whether to expose pageId on page-scoped tools and route requests by page ID (useful for concurrent agent sessions).',
+      'Expose pageId on page-scoped tools and route requests by page ID (useful for concurrent agent sessions). Use --no-page-id-routing to disable.',
+    default: true,
   },
   experimentalDevtools: {
     type: 'boolean',
--- a/README.md
+++ b/README.md
@@ -855,8 +855,13 @@ Most MCP clients start one Chrome DevTools MCP server per conversation.
 By default, the server runs with `--pageIdRouting` enabled, making `pageId` a
 required parameter on page-scoped tools (such as `click`, `fill`, `navigate_page`,
 `take_snapshot`, etc.) so multiple agents or subagents sharing a server instance can
-route tool calls directly to the specific tab they are working with. To disable
-this behavior and default to the currently selected page, pass
+route tool calls directly to the specific tab they are working with.
+
+For `evaluate_script`, `pageId` is required by default for targeting pages, but
+becomes optional when `--categoryExtensions` is enabled so that `serviceWorkerId`
+can be specified instead to evaluate inside an extension background service worker.
+
+To disable this behavior and default to the currently selected page, pass
 `--no-page-id-routing`.
 
 ```json
--- a/README.md
+++ b/README.md
@@ -663,10 +663,10 @@ The Chrome DevTools MCP server supports the following configuration option:
   - **Type:** boolean
   - **Default:** `false`
 
-- **`--experimentalPageIdRouting`/ `--experimental-page-id-routing`**
-  Whether to expose pageId on page-scoped tools and route requests by page ID (useful for concurrent agent sessions).
+- **`--pageIdRouting`/ `--page-id-routing`**
+  Expose pageId on page-scoped tools and route requests by page ID (useful for concurrent agent sessions). Use --no-page-id-routing to disable.
   - **Type:** boolean
-  - **Default:** `false`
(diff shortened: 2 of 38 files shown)


src/bin/chrome-devtools-mcp-cli-options.ts
```diff
@@ -150,7 +150,7 @@ export const cliOptions = {
   experimentalPageIdRouting: {
     type: 'boolean',
     describe:
-      'Whether to expose pageId on page-scoped tools and route requests by page ID.',
+      '(Deprecated, now always enabled) pageId is always exposed on page-scoped tools.',
     hidden: true,
```
OrKoN: so we have options:

1) remove the flag altogether
2) remove hidden: true an experimental prefix.

If we keep the flag, I think we can decide to keep it off by default or keep an option to turn off pageId routing to save tokens. 

cc @natorion wdyt?

natorion: The overhead is only the resident token cost of having pageId in help, right? The parameter per-se is optional.

In CLI scenarios I expect such a resident token cost to be miniscule compared to whatever is returned via take_snapshot.

OrKoN: It's optional but it is included on each llm iteration even if no heavy output like take_snapshot is used.

OrKoN:
TODO: this cannot land before the fixes to the wait helpers are made to correctly operate on the specified page.

OrKoN:
TODO: evaluate that the model uses pageId correctly via eval scenarios in https://github.com/ChromeDevTools/chrome-devtools-mcp/tree/main/scripts/eval_scenarios

src/bin/chrome-devtools-mcp-cli-options.ts
bcfmtolgahan: PR #1780 addresses the wait helpers, waitForEventsAfterAction is now on McpPage and each tool uses the correct page's throttling settings.

On the flag: happy to go either way. If keeping it as an opt-out for token savings makes sense, I can rewire index.ts to respect it as a disable toggle and rename it (e.g. --noPageIdRouting). If the consensus is that the overhead is acceptable, I'll remove it entirely. What's your preference?

OrKoN: We need some time to evaluate if the overhead is acceptable. For now, let's keep the flag in case some users are not happy with the overhead or in case the pageId routing does not work as expected. There is no need to rename to `--noPageIdRouting` as yargs automatically handles this for boolean args. Right now you can already use `--noExperimentalPageRouting`, so I would say let's keep the flag, drop the experimental prefix, remove hidden: true and set default to true. If we find the token overhead to be too much, we can update the arg to default to false. Or users can use `--no-page-id-routing` or `--page-id-routing=false`.

bcfmtolgahan: Done. Dropped the experimental prefix, removed hidden: true, set default: true. Users can opt out with --no-page-id-routing.

OrKoN:
@bcfmtolgahan would you be up to rebasing this PR?

bcfmtolgahan:
@OrKoN  rebased onto latest main. Adapted to the new 'ToolHandler.ts', location for schema injection and routing. Squashed to a single commit since the changes got intertwined during the adaptation, happy to split if preferred.

OrKoN:
FYI we made pageId required when the page id routing is enabled because it makes the MCP server perform better with the agents. But it also causes a regression for the CLI interface where the required pageId becomes a mandatory positional argument for all commands and the skill is not updated. I think we would to disable pageId routing for the CLI for now. cc @samiyac

OrKoN:
also the tests fail because the pageId is now a required argument.

bcfmtolgahan:
@OrKoN Rebased onto latest main and prototyped the CLI fix you mentioned: gated pageId injection on !viaCli (daemon + generate-cli.ts already pass --viaCli), so CLI commands drop the mandatory pageId positional while the MCP server keeps it. Also fixed the failing e2e tests. Suite is green.

Happy to push this if it matches what you had in mind or if you'd rather own the CLI-disable (or prefer making pageId optional instead), just let me know and I'll hold off.

samiyac:
Hi @bcfmtolgahan 
@OrKoN  and I had a discussion about the regression for the CLI interface and I submitted #2142 as a fix for it. We are re-running our evaluations to see if this fixed the regression. I will also be editing the SKILL file for CLI once #1777 is in.

bcfmtolgahan:
Understood, thanks @samiyac  If anything's needed on my end (rebasing onto latest main, etc.), just let me know

thomasbachem:
On 1.7.0, enabling `--experimentalPageIdRouting` makes `pageId` required, so any call without one fails:

```
MCP error -32602: Input validation error: Invalid arguments for tool evaluate_script: Required at pageId
```

Intentional per the thread above, but the flag description – *"expose pageId on page-scoped tools and route requests by page ID"* – reads as additive, so I enabled it expecting untargeted calls to fall back to the selected page. Since this PR makes it the default, a line in the docs or a migration note would save others the detour.

OrKoN (approved):

PR #1244, opened 2026-03-26
feat: make pageId routing always-on and fix waitForEventsAfterAction page isolation

bcfmtolgahan:
## Summary

  - Remove the `--experimental-page-id-routing` gate so `pageId` is always exposed on page-scoped tools, preventing race conditions in multi-agent workflows where concurrent `select_page` calls cause tools to operate on the wrong page
  - Fix `waitForEventsAfterAction` to accept an explicit `page` parameter so it uses the correct page's emulation settings (CPU throttling, network conditions) instead of the global selected page
  - Update all page-scoped tool handlers (`input`, `pages`, `script`) to pass the explicit page to `waitForEventsAfterAction`

  ## Changes

  - `src/index.ts`: Remove `experimentalPageIdRouting` gate from schema injection and page resolution
  - `src/McpContext.ts`: Add optional `page` parameter to `waitForEventsAfterAction`
  - `src/tools/ToolDefinition.ts`: Update `Context` interface and `pageId` description
  - `src/tools/input.ts`: Pass `{page: request.page}` to all 8 `waitForEventsAfterAction` call sites
  - `src/tools/pages.ts`: Pass explicit page in `navigatePage` and `newPage`; update `select_page` description for multi-agent guidance
  - `src/tools/script.ts`: Remove duplicate `experimentalPageIdRouting` gate, always include `pageIdSchema`, pass page to `performEvaluation`
  - `src/bin/chrome-devtools-mcp-cli-options.ts`: Deprecate `--experimental-page-id-routing` flag
  - `docs/tool-reference.md`, `src/bin/cliDefinitions.ts`: Regenerated via `npm run gen`

  ## Test plan

  - [x] E2E tests for pageId routing (screenshot, evaluate_script, navigate_page targeting correct page via pageId)
  - [x] E2E tests for race condition simulation (interleaved select_page calls don't affect pageId-routed operations)
  - [x] E2E tests for waitForEventsAfterAction page isolation (CPU throttling settings don't leak across pages)
  - [x] All 544 existing tests pass with zero regressions

PR #1780, opened 2026-04-01
refactor: move waitForEventsAfterAction to McpPage

bcfmtolgahan:
Splits out from #1244 per review feedback.

'waitForEventsAfterAction' previously lived in 'McpContext' and always used the selected page's CPU/network throttling settings. With pageId routing, a tool can target a different page than the selected one, meaning wrong throttling multipliers were applied.

Moving the method to 'McpPage' fixes this: each tool now calls 'page.waitForEventsAfterAction(...)' and gets the correct page's emulation settings.

'getNetworkMultiplierFromString' is extracted to 'WaitForHelper.ts' to avoid a circular import (McpContext → McpPage already exists).

Unblocks #1777.

### DIFF
--- a/src/tools/script.ts
+++ b/src/tools/script.ts
@@ -77,9 +77,11 @@ Example with arguments: \`(el) => {
         }
 
         const worker = await getWebWorker(context, serviceWorkerId);
-        await context.getSelectedMcpPage().waitForEventsAfterAction(async () => {
-          await performEvaluation(worker, fnString, [], response);
-        });
+        await context
+          .getSelectedMcpPage()
+          .waitForEventsAfterAction(async () => {
+            await performEvaluation(worker, fnString, [], response);
+          });
         return;
       }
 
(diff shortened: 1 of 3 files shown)


src/tools/script.ts
```diff
@@ -77,7 +77,7 @@ Example with arguments: \`(el) => {
         }
 
         const worker = await getWebWorker(context, serviceWorkerId);
-        await performEvaluation(worker, fnString, [], response, context);
+        await performEvaluation(worker, fnString, [], response, context.getSelectedMcpPage());
```
OrKoN: I think there was a pre-existing issue here. Service worker evaluation should use the page based waitForEventsAfterAction helper. We should extract the waitForEventsAfterAction call out of performEvaluation I believe. cc @nroscino

bcfmtolgahan: Addressed your feedback. Extracted waitForEventsAfterAction out of performEvaluation. Service worker path now uses context.getSelectedMcpPage().waitForEventsAfterAction(), normal page path uses mcpPage.waitForEventsAfterAction(). Also fixed the same anti-pattern in the newly added executeInPageTool.

PR #1783, opened 2026-04-01
feat: add update notification to both binaries

mathiasbynens:
Both `chrome-devtools` and `chrome-devtools-mcp` now log a notification when a newer version is detected to be available.

This detection is implemented as follows:

1. Read the latest version from a local 24-hour cache (`~/.cache/chrome-devtools-mcp/latest.json`).
2. If the cache is stale or missing, spawn a detached background process to fetch the latest version from the npm registry and update the cache file.

Issue: https://crbug.com/496757598

### DIFF
--- a/src/utils/check-for-updates.ts
+++ b/src/utils/check-for-updates.ts
@@ -16,10 +16,18 @@ import {VERSION} from '../version.js';
  * Notifies the user if an update is available.
  * @param message The message to display in the update notification.
  */
+let isChecking = false;
+
+/** @internal Reset flag for tests only. */
+export function resetUpdateCheckFlagForTesting() {
+  isChecking = false;
+}
+
 export async function checkForUpdates(message: string) {
-  if (process.env['CHROME_DEVTOOLS_MCP_NO_UPDATE_CHECKS']) {
+  if (isChecking || process.env['CHROME_DEVTOOLS_MCP_NO_UPDATE_CHECKS']) {
     return;
   }
+  isChecking = true;
 
   const cachePath = path.join(
     os.homedir(),
@@ -47,6 +55,20 @@ export async function checkForUpdates(message: string) {
   const now = Date.now();
   if (stats && now - stats.mtimeMs < 24 * 60 * 60 * 1000) {
     return;
+  }
+
+  // Update mtime immediately to prevent multiple subprocesses.
+  try {
+    const parentDir = path.dirname(cachePath);
+    await fs.mkdir(parentDir, {recursive: true});
+    const nowTime = new Date();
+    if (stats) {
+      await fs.utimes(cachePath, nowTime, nowTime);
+    } else {
+      await fs.writeFile(cachePath, JSON.stringify({version: VERSION}));
+    }
+  } catch {
+    // Ignore errors.
   }
 
   // In a separate process, check the latest available version number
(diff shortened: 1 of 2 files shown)


src/utils/check-for-updates.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  const now = Date.now();
+  if (cache && now - cache.timestamp < 24 * 60 * 60 * 1000) {
+    return;
+  }
+
+  // In a separate process, check the latest available version number
+  // and update the local snapshot accordingly.
+  const scriptPath = path.join(import.meta.dirname, 'check-latest-version.js');
```
OrKoN: I think the file is there when the server is bundled, but let's try to move this file to the bin folder to make it explicit that this file (check-latest-version.js) is used a separate executable?

src/utils/check-for-updates.ts
```diff
@@ -0,0 +1,63 @@
...
+ */
+export async function checkForUpdates(message: string) {
+  const cachePath = path.join(
+    os.homedir(),
+    '.cache',
+    'chrome-devtools-mcp',
+    'latest.json',
+  );
+
+  let cache: {version: string; timestamp: number} | undefined;
```
OrKoN: could we use the file modified date instead? or are there any drawbacks to that?

src/utils/check-for-updates.ts
mathiasbynens: Good call. Done.

src/utils/check-for-updates.ts
```diff
@@ -0,0 +1,63 @@
...
+import path from 'node:path';
+import process from 'node:process';
+
+import {VERSION} from '../version.js';
+
+/**
+ * Notifies the user if an update is available.
+ * @param message The message to display in the update notification.
+ */
+export async function checkForUpdates(message: string) {
```
OrKoN, line 26: let's add an env variable to suppress the checks. I imagine some folks might have custom registries and mirrors and might not be able to update right away or access the npm registry. WDYT?

src/utils/check-for-updates.ts
mathiasbynens: No strong preference here. It felt simpler to just read the file contents rather than also check the file metadata. I’ll give this a spin.

mathiasbynens: Pushed as a separate commit so you can see the diff: c7a8ae4a07622469c3c57985fead537e21185367 Note that I stuck to JSON as opposed to saving the raw version as a text file, in case we want to add more properties in the future.

OrKoN: Just thought of the following edge case, what if writing to file takes a long time but we have many CLI invocation in a short amount of time, would we create multiple (potentially too many) concurrent subprocesses to update the version?

src/utils/check-for-updates.ts
mathiasbynens: f394814c28fdf528d61fe9556a5107d7d64796a4

src/utils/check-for-updates.ts
mathiasbynens: We could try to work around that by updating the file’s mtime before spawning the subprocess that performs the fetch + file update: f53ab369334d0ead1284149c40f9ca793c33e6ed It’s still possible that the mtime update succeeds but writing the file contents takes a long time or fails, however.

OrKoN (approved):
LGTM thanks! changed to `feat:`

Issue #978, opened 2026-02-17
Connection drops when opening multiple browser pages - no graceful degradation


ljch2018:
### Description of the bug

When using chrome-devtools-mcp for multi-page operations, opening too many browser tabs (6+) causes the WebSocket connection to drop with error:

MCP error -32000: Connection closed

## Root Cause Analysis

After investigation into the source code (v0.17.1), I identified three contributing factors:

### 1. CDP Session Multiplexing Overhead
Each browser page creates a separate CDP session via Target.attachToTarget. With 6+ pages, the event volume can overwhelm the single WebSocket channel.

### 2. No Connection Health Monitoring  
The WebSocket transport has no heartbeat/keepalive mechanism. When Chrome becomes unresponsive, the connection silently dies.

### 3. No Auto-Recovery
In browser.js, the code only checks connected state but provides no reconnection logic.

### Chrome DevTools MCP version

0.17.1

### Chrome version

Latest stable

### Coding agent version

Claude Code

### Model version

Claude Opus 4.5

### Node version

v22.x

### Operating system

macOS

### Extra checklist

- [x] I want to provide a PR to fix this bug

OrKoN:
How do you launch the MCP server and which Chrome version do you launch and with what args? by default, the MCP server is using pipes and not WebSockets.

OrKoN:
I am not able to reproduce the behavior with 6+ tabs, works fine for me.

OrKoN:
Closed as not reproducible. Please provide [logs](https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/docs/troubleshooting.md#debugging) or more detailed steps to reproduce and we can take another look.

Issue #870, opened 2026-02-03
list_pages/new_page time out with ProtocolError: Network.enable timed out when using --autoConnect (Chrome 144, OpenCode)


alexaaaaaander:
## What I'm doing
- MCP client: OpenCode (local MCP server)
- chrome-devtools-mcp started via:
  - `npx -y chrome-devtools-mcp@latest --autoConnect --channel=stable --logFile <path>`
- Chrome setup:
  - Chrome 144 stable
  - Enabled in Chrome UI: `chrome://inspect/#remote-debugging` (checkbox "Allow remote debugging for this browser instance")
  - Accepted the "Allow remote debugging?" permission prompt

## Expected
- `list_pages` returns open tabs/pages.
- `new_page` can open a URL.

## Actual
- MCP client reports request timeout on `list_pages`.
- Server log shows `list_pages` and `new_page` failing with CDP/Puppeteer protocol timeouts, especially `Network.enable timed out`.

## Evidence from server log
(edited to remove URLs and local username)

```text
2026-02-03T00:13:13.093Z puppeteer:error [
  ProtocolError: Runtime.runIfWaitingForDebugger timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.
  ...
]
2026-02-03T00:13:13.095Z mcp:log list_pages error: ProtocolError: Network.enable timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.
    at NetworkManager.addClient (.../chrome-devtools-mcp/.../index.js:49803:24)
    at FrameManager.initialize (.../chrome-devtools-mcp/.../index.js:50406:38)
    at #initialize (.../chrome-devtools-mcp/.../index.js:51924:36)
    at CdpPage._create (.../chrome-devtools-mcp/.../index.js:51738:31)

2026-02-03T00:13:13.097Z mcp:log new_page request: { "url": "https://example.com", "background": true, "timeout": 30000 }
2026-02-03T00:13:13.097Z mcp:log new_page error: ProtocolError: Network.enable timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.
    at NetworkManager.addClient (.../chrome-devtools-mcp/.../index.js:49803:24)
    at FrameManager.initialize (.../chrome-devtools-mcp/.../index.js:50406:38)
    at #initialize (.../chrome-devtools-mcp/.../index.js:51924:36)
[...]

PR #1804, opened 2026-04-02
feat: add --skipUnresponsiveTabs flag for connecting to browsers with many tabs

jaidhyani:
`browser.pages()` hangs forever if any tab is unresponsive (discarded, sleeping, dead localhost). This is common when connecting to a daily-driver browser with 20+ tabs via `--browserUrl`.

Root cause: `puppeteer.connect()` has no `protocolTimeout`, and `browser.pages()` tries to attach to every target. One bad tab blocks everything.

This adds `--skipUnresponsiveTabs` (default: false). When on:
- Sets `protocolTimeout: 15s` on connect
- If `browser.pages()` times out (10s), falls back to per-target enumeration with 5s individual timeouts
- Skips unresponsive tabs and logs them

Flag off = no behavior change.

Hit this on Brave 145 (28 tabs, 5 discarded). Brave is more aggressive about discarding than Chrome but the fix isn't browser-specific.

Related: puppeteer/puppeteer#5633, puppeteer/puppeteer#14708, #870, #978

### DIFF
--- a/README.md
+++ b/README.md
@@ -560,6 +560,11 @@ The Chrome DevTools MCP server supports the following configuration option:
   If enabled, ignores errors relative to self-signed and expired certificates. Use with caution.
   - **Type:** boolean
 
+- **`--skipUnresponsiveTabs`/ `--skip-unresponsive-tabs`**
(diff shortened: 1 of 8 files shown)


OrKoN:
I believe this issue was fixed in Chrome. Also, I am not sure this fixed the underlying issue and the PR has conflicts so closing as stale.

PR #1808, opened 2026-04-03
fix: remove stray )} characters from redirect chain output

firat-bcfm:
The redirect chain formatter in `converNetworkRequestDetailedToStringDetailed` had a stray `)}` at the end of the template literal, causing every redirect chain entry to be suffixed with those characters in tool output.


nroscino:
Hi @firat-bcfm!
Before proceeding we need you to sign the CLA. 
Also, can you please cover this with a test so that next time something similar happens we will catch it with a failing test?

firat-bcfm:
Okey

PR #1823, opened 2026-04-07
feat(screenshot): add CLI options to cap screenshot size at the source

### LINKED ISSUE #879: Claude errors if the image dimensions exceed 8000 pixels
## Problem

When using `take_screenshot` on large viewports (e.g., testing responsive design on wide monitors or using `fullPage: true` on long pages), the tool fails with:

```
API Error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.7.content.2.image.source.base64.data: At least one of the image dimensions exceed max allowed size: 8000 pixels"}}
```

This happens frequently when testing responsive layouts, making the tool unusable for this common use case.

## Expected behavior

Screenshots should be automatically resized to fit within Claude API limits (max 8000px per dimension) before being sent, preserving aspect ratio.

## Reproduction

1. Open a page in a browser with a large viewport or long scrollable content
2. Call `take_screenshot` (especially with `fullPage: true`)
3. Tool fails with the above error

## Suggested fix

Resize images before base64 encoding when any dimension exceeds 8000px, preserving aspect ratio. This is a common pattern in similar tools.

## Environment

- MCP: chrome-devtools-mcp (latest via npx)
- Client: Claude Code
- OS: macOS

antoinekm:
## Summary

Adds **opt-in** CLI flags so operators can cap the size of screenshots returned by `take_screenshot` before they are embedded in the MCP response. Refs #879.

The flags address two related symptoms reported when MCP clients display screenshots inline:

1. **Per-image dimension limit**: hosted LLM APIs commonly reject images exceeding per-image dimension constraints (typical caps are in the 2000-8000 px range, sometimes scaling down further when many images are in the same request). This is the exact error reported in #879.
2. **Cumulative request size**: after many captures, the cumulative base64 payload eventually pushes a request over the per-call body size limit imposed by the LLM API.

Both can be mitigated at the source by reducing format/quality and downscaling the capture.

## New flags (all opt-in)

- `--screenshot-format <jpeg|png|webp>`: override the default format used by `take_screenshot` when the caller does not specify one
- `--screenshot-quality <0-100>`: override the default JPEG/WebP quality. Ignored for PNG
- `--screenshot-max-width <px>`: downscale screenshots wider than this before they are returned
- `--screenshot-max-height <px>`: downscale screenshots taller than this. Combines with `--screenshot-max-width`; the smaller scale wins so both bounds are respected while preserving aspect ratio

For the exact error in #879, the recipe is `--screenshot-max-width=8000 --screenshot-max-height=8000` (or a smaller value such as `2000` if many images may end up in the same request, depending on the operator's chosen API).

## Implementation

- Resizing leverages Puppeteer's `clip.scale` (CDP `Page.captureScreenshot`), so **no new dependencies**.
- Source dimensions per capture mode:
  - viewport: `page.viewport()`
  - full page: `document.documentElement.scrollWidth/scrollHeight` via `page.evaluate()`
  - element (`uid`): `elementHandle.boundingBox()`
- For element and full-page captures with a downscale clip, the call routes through `page.screenshot({clip})` so the scale parameter applies. `captureBeyondViewport` is left to Puppeteer's default (`true` when a clip is set), preserving correct behavior for elements below the fold and full-page captures.
- ~150 lines of source code, ~200 lines of new tests.

## Backwards compatibility

**Fully opt-in**: when no flags are set, `take_screenshot` returns the exact same bytes as before. No behavioral change for existing users.

## Design alignment

- Aligned with the **"Reference over Value"** principle in `docs/design-principles.md`: the existing 2 MB threshold still routes oversized screenshots to a temporary file. This change only reduces the size of the **inline base64 fallback path**, which the principles document calls out as an acceptable exception when MCP clients display images natively.
- The MCP server **hardcodes no LLM-specific size limits**. Operators pick the values that match their client/model combination. This keeps the maintenance surface here minimal as model limits evolve, and is intended as a **complement to, not a replacement for**, fixes in the MCP client itself.

## Addressing concerns raised in #879

> "It's not feasible for us to maintain this. Limits will change when models change." (@natorion)

The flags are pure parameters; nothing about the upstream LLM is encoded in the server. When a vendor raises (or lowers) a limit, no code change is needed here, only the operator's CLI args change.

> "`filePath` / `page_resize` already work as a workaround." (@OrKoN)

`filePath` is great when the call site knows it's about to take a huge screenshot, but as you noted earlier in the thread, an oversized image already in the request history keeps causing failures even on subsequent calls. `page_resize` works but mutates the page being debugged. The resize in this PR happens **between Puppeteer and the MCP response**, so the inspected page is untouched and the failure mode is prevented at the source.

> "Should be fixed client side."

Agreed, this PR is intended as a complement, not a substitute. A client-side fix (e.g. compaction evicts/downsamples old images) handles the cumulative case for *any* MCP. A server-side cap handles the per-call dimension limit for users who hit it before compaction can kick in. The two address overlapping but distinct failure modes.

Happy to drop or rework any of this if the maintainers prefer a different shape, for example making the threshold automatic from a single `--max-image-bytes` knob, or rejecting the PR entirely in favor of waiting for a client-side fix. Just wanted to put a concrete option on the table.

## Tests

Added 6 new tests:

- `honors screenshotFormat default from CLI args`
- `keeps "png" as default format when no CLI override is set`
- `downscales viewport screenshot when screenshotMaxWidth is set`
- `downscales using the smaller scale when both max-width and max-height are set`
- `does not resize when source is smaller than the max bounds`
- `downscales full page screenshot when screenshotMaxWidth is set`

All 627 tests in the suite pass. `npm run typecheck` and `npm run check-format` are clean.

## Notes for reviewers

- The dimensions compared against `--screenshot-max-width/height` are **CSS pixels** (`page.viewport()`), not raw bitmap pixels. With `deviceScaleFactor > 1` (HiDPI emulation) the actual bitmap may still be larger. Happy to clarify this in the option description if preferred.
- For element captures with a downscale clip, the call routes through `page.screenshot({clip})` instead of `element.screenshot()`. Same-frame elements are correct (boundingBox returns main-frame coords). I have **not** exercised this path against cross-origin iframe elements; let me know if you'd like a fallback there.
- The PR is currently in **Draft** state pending CLA verification and any feedback on the framing above.

Refs #879
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/879

### DIFF
--- a/src/tools/screenshot.ts
+++ b/src/tools/screenshot.ts
@@ -5,107 +5,232 @@
  */
 
 import {zod} from '../third_party/index.js';
-import type {ElementHandle, Page} from '../third_party/index.js';
+import type {
+  BoundingBox,
+  ElementHandle,
+  Page,
+  ScreenshotClip,
+} from '../third_party/index.js';
 
 import {ToolCategory} from './categories.js';
 import {definePageTool} from './ToolDefinition.js';
 
-export const screenshot = definePageTool({
-  name: 'take_screenshot',
-  description: `Take a screenshot of the page or element.`,
-  annotations: {
-    category: ToolCategory.DEBUGGING,
-    // Not read-only due to filePath param.
-    readOnlyHint: false,
-  },
-  schema: {
-    format: zod
-      .enum(['png', 'jpeg', 'webp'])
-      .default('png')
-      .describe('Type of format to save the screenshot as. Default is "png"'),
-    quality: zod
-      .number()
-      .min(0)
-      .max(100)
-      .optional()
-      .describe(
-        'Compression quality for JPEG and WebP formats (0-100). Higher values mean better quality but larger file sizes. Ignored for PNG format.',
+type ScreenshotFormat = 'png' | 'jpeg' | 'webp';
+
+async function getSourceBox(
+  page: Page,
+  element: ElementHandle | undefined,
+  fullPage: boolean,
+): Promise<BoundingBox | undefined> {
+  if (element) {
+    const box = await element.boundingBox();
+    return box ?? undefined;
+  }
+  if (fullPage) {
+    const dims = await page.evaluate(() => ({
+      width: Math.max(
+        document.documentElement.scrollWidth,
+        document.body?.scrollWidth ?? 0,
       ),
-    uid: zod
-      .string()
-      .optional()
-      .describe(
-        'The uid of an element on the page from the page content snapshot. If omitted, takes a page screenshot.',
+      height: Math.max(
+        document.documentElement.scrollHeight,
+        document.body?.scrollHeight ?? 0,
       ),
-    fullPage: zod
-      .boolean()
-      .optional()
-      .describe(
-        'If set to true takes a screenshot of the full page instead of the currently visible viewport. Incompatible with uid.',
-      ),
-    filePath: zod
-      .string()
-      .optional()
-      .describe(
-        'The absolute path, or a path relative to the current working directory, to save the screenshot to instead of attaching it to the response.',
-      ),
-  },
-  blockedByDialog: true,
-  verifyFilesSchema: ['filePath'],
-  handler: async (request, response, context) => {
-    if (request.params.uid && request.params.fullPage) {
-      throw new Error('Providing both "uid" and "fullPage" is not allowed.');
+    }));
+    if (dims.width <= 0 || dims.height <= 0) {
+      return undefined;
     }
+    return {x: 0, y: 0, width: dims.width, height: dims.height};
+  }
+  const viewport = page.viewport();
+  if (!viewport) {
+    return undefined;
+  }
+  return {x: 0, y: 0, width: viewport.width, height: viewport.height};
+}
 
-    let pageOrHandle: Page | ElementHandle;
-    if (request.params.uid) {
-      pageOrHandle = await request.page.getElementByUid(request.params.uid);
-    } else {
-      pageOrHandle = request.page.pptrPage;
-    }
+function computeDownscaleClip(
+  box: BoundingBox,
+  maxWidth: number | undefined,
+  maxHeight: number | undefined,
+): ScreenshotClip | undefined {
+  const widthScale =
+    maxWidth !== undefined ? Math.min(1, maxWidth / box.width) : 1;
+  const heightScale =
+    maxHeight !== undefined ? Math.min(1, maxHeight / box.height) : 1;
+  const scale = Math.min(widthScale, heightScale);
+  if (scale >= 1) {
+    return undefined;
+  }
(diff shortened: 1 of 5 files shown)


OrKoN:
@antoinekm could you please rebase the PR?

src/tools/screenshot.ts
```diff
@@ -5,101 +5,236 @@
...
-        'The uid of an element on the page from the page content snapshot. If omitted, takes a page screenshot.',
-      ),
-    fullPage: zod
-      .boolean()
-      .optional()
-      .describe(
-        'If set to true takes a screenshot of the full page instead of the currently visible viewport. Incompatible with uid.',
+type ScreenshotFormat = 'png' | 'jpeg' | 'webp';
+
+function isScreenshotFormat(value: unknown): value is ScreenshotFormat {
```
OrKoN: why do we need isScreenshotFormat? MCP schema validates the format.

src/tools/screenshot.ts
```diff
@@ -5,101 +5,236 @@
...
+  return value === 'png' || value === 'jpeg' || value === 'webp';
+}
+
+function isPositiveFiniteNumber(value: number | undefined): value is number {
+  return value !== undefined && Number.isFinite(value) && value > 0;
```
OrKoN: let's do number validation when parsing CLI args with yargs? I think here we should expect an integer number or undefined so we do not need to check again if the CLI args were parse.

antoinekm:
Rebased on `main` (now `8c76d72`). Conflicts with the recent `verifyFilesSchema` refactor (#2152) and the new `saveFile(extension)` signature were resolved. Also addressed both review comments below in the same commit so the PR stays at a single conventional commit. PTAL 🙏

src/tools/screenshot.ts
antoinekm: Good catch, removed. The yargs `choices: ['jpeg', 'png', 'webp'] as const` already narrows `args.screenshotFormat` to `'jpeg' | 'png' | 'webp' | undefined`, so the runtime guard was dead code. Now simply `const defaultFormat: ScreenshotFormat = screenshotFormat ?? 'png';`.

src/tools/screenshot.ts
antoinekm: Done. Moved the validation into `coerce` on each numeric CLI option in `chrome-devtools-mcp-cli-options.ts`:

- `screenshotQuality`: integer in `[0, 100]`
- `screenshotMaxWidth` / `screenshotMaxHeight`: positive integer

Invalid values now throw a clear error at parse time (e.g. `Invalid screenshotMaxWidth -5. Expected a positive integer.`), and the tool consumes the args directly with no re-validation.

OrKoN (approved):
LGTM thanks!

PR #1839, opened 2026-04-09
feat: auto handle dialogs during script evaluation 

Lightning00Blade:
Provide a argument to LLM to handle dialogs that come up during code execution.

### DIFF
--- a/src/tools/script.ts
+++ b/src/tools/script.ts
@@ -81,7 +81,7 @@ Example with arguments: \`(el) => {
           .getSelectedMcpPage()
           .waitForEventsAfterAction(async () => {
             await performEvaluation(worker, fnString, [], response);
-          });
+          }, {dialog: 'accept'});
         return;
       }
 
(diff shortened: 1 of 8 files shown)


src/tools/script.ts
```diff
@@ -77,11 +77,12 @@ Example with arguments: \`(el) => {
         }
 
         const worker = await getWebWorker(context, serviceWorkerId);
-        await context
-          .getSelectedMcpPage()
-          .waitForEventsAfterAction(async () => {
+        await context.getSelectedMcpPage().waitForEventsAfterAction(
```
OrKoN, line 87: I wonder if we should expose it on the tool params similar to 

```
handleBeforeUnload: zod
      .enum(['accept', 'decline'])
      .optional()
      .describe(
        'Whether to auto accept or beforeunload dialogs triggered by this navigation. Default is accept.',
      ),
```

or at least update the tool description and indicate that all dialogs caused by the script execution will be auto-accepted to avoid unexpected acceptance of a dialog. WDYT?

ubutatu (approved):

ubutatu:
> - [ ] - - **
**

ubutatu (commented):
Hh

OrKoN (approved):

PR #1841, opened 2026-04-09
fix: skip frozen/discarded targets in page enumeration

### LINKED ISSUE #1230: Frozen/discarded background tabs can make page enumeration fail with Network.enable timeout
### Description of the bug

`chrome-devtools-mcp` can fail to initialize if the connected Chrome profile contains background tabs that have been frozen or discarded by Chrome's memory-saving behavior.

In my repro, browser-level CDP remains healthy (`Browser.getVersion`, `Target.getTargets`, `Target.setAutoAttach` all succeed), but page enumeration fails because Puppeteer times out on `Network.enable` while constructing some `Page` sessions.

The important detail is that the affected targets recover immediately after `Target.activateTarget`, which suggests the root cause is not a broken browser connection but a page target that is suspended/discarded and temporarily unable to respond to page-session CDP commands.

This looks related to #1156, but I am filing this separately because I have a more specific reproduction and a narrower suspected fix direction.

### Reproduction

1. Start Chrome on macOS and enable remote debugging.
2. Open many tabs in one long-lived profile.
3. Leave some tabs in the background long enough for Chrome to freeze/discard them.
4. Connect `chrome-devtools-mcp` to that running browser profile.
5. Call any MCP tool that eventually needs page enumeration / `browser.pages()` (for example `list_pages`, `take_snapshot`, or some navigation-related flows).
6. Observe a timeout like:

```text
Network.enable timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.
```

Additional signal from direct CDP probing:

- Browser-level commands still work.
- Attaching to the affected target succeeds.
- `Network.enable` and `Runtime.enable` can time out for that target while it is in the background.
- After calling `Target.activateTarget` for the same target, `Network.enable` succeeds.

### Expectation

If one page target is frozen/discarded and cannot complete page-session initialization, `chrome-devtools-mcp` should skip that target instead of failing the entire page enumeration / MCP initialization flow.

A defensive behavior like this would make the tool much more robust when used against real long-lived browsing profiles.

### MCP configuration

```json
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": [
        "chrome-devtools-mcp@latest",
        "--autoConnect"
      ]
    }
  }
}
```

I also reproduced the same underlying failure while connecting through Puppeteer to the browser WebSocket endpoint and calling `browser.pages()`.

0.20.3

### Chrome version

146.0.7680.153

### Coding agent version

Codex CLI 0.116.0

### Model version

GPT-5.4

### Chat log

Not included because it contains local browsing context. The relevant protocol behavior is summarized above.

### Node version

23.7.0

### Operating system

macOS

### Extra checklist

20syldev:
## Problem

When Chrome freezes or discards background tabs (memory-saving behavior), `browser.pages()` can time out on `Network.enable` for those targets and abort the entire page enumeration — even though the rest of the browser and all active tabs are healthy.

```
Network.enable timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.
```

This causes any MCP tool that needs page enumeration (`list_pages`, `take_snapshot`, navigation flows, etc.) to fail completely when the connected Chrome profile has been open long enough for background tabs to be frozen/discarded.

## Root cause

`browser.pages()` iterates every page-type target and calls `target.page()` for each, which creates a Puppeteer CDP session and sends initialization commands including `Network.enable`. Frozen/discarded targets cannot respond to page-session CDP commands, so they time out — and the error propagates out of `browser.pages()`, killing the entire list.

## Fix

Replace the single `browser.pages()` call in `#getAllPages()` with a per-target iteration over `browser.targets()`. Each call to `target.page()` is wrapped in its own `try/catch` so that one unresponsive background tab is logged and skipped instead of aborting the whole enumeration.

- `page` type targets are always included (existing behavior)
- `background_page` / `webview` types are included when `experimentalIncludeAllPages` is set (preserves existing opt-in behavior)
- The `allTargets` snapshot is reused for the extension-target loop below, so there is no extra `browser.targets()` call

Fixes #1230

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -564,11 +564,47 @@ export class McpContext implements Context {
     isolatedContextNames: Map<Page, string>;
   }> {
     const defaultCtx = this.browser.defaultBrowserContext();
-    const allPages = await this.browser.pages(
-      this.#options.experimentalIncludeAllPages,
-    );
 
+    // Enumerate targets individually instead of calling browser.pages() so
+    // that a single frozen/discarded background tab that times out on
+    // Network.enable cannot abort the entire page enumeration (see #1230).
     const allTargets = this.browser.targets();
+    const pageTargets = allTargets.filter(target => {
+      const type = target.type();
+      if (type === 'page') return true;
+      if (this.#options.experimentalIncludeAllPages) {
+        return type === 'background_page' || type === 'webview';
+      }
+      return false;
+    });
+    const pageResults = await Promise.all(
+      pageTargets.map(async target => {
+        try {
+          const page = await Promise.race([
+            target.page(),
+            new Promise<null>(resolve =>
+              setTimeout(() => resolve(null), DEFAULT_TIMEOUT),
+            ),
+          ]);
+          if (!page) {
+            this.logger(
+              'Timed out attaching to target at',
+              target.url(),
+              '— likely frozen or discarded',
+            );
+          }
+          return page;
+        } catch (err) {
+          this.logger(
+            'Skipping frozen/discarded target at',
+            target.url(),
+            err,
+          );
+          return null;
+        }
+      }),
+    );
+    const allPages = pageResults.filter((p): p is Page => p !== null);
     const extensionTargets = allTargets.filter(target => {
       return (
         target.url().startsWith('chrome-extension://') &&
@@ -578,7 +614,12 @@ export class McpContext implements Context {
 
     for (const target of extensionTargets) {
       // Right now target.page() returns null for popup and side panel pages.
-      let page = await target.page();
+      let page = await Promise.race([
+        target.page(),
+        new Promise<null>(resolve =>
+          setTimeout(() => resolve(null), DEFAULT_TIMEOUT),
+        ),
+      ]);
       if (!page) {
         // We need to cache pages instances for targets because target.asPage()
         // returns a new page instance every time.
--- a/tests/McpContext.test.ts
+++ b/tests/McpContext.test.ts
@@ -9,9 +9,12 @@ import path from 'node:path';
 import {afterEach, describe, it} from 'node:test';
 import {pathToFileURL} from 'node:url';
 
+import logger from 'debug';
+import puppeteer, {Locator} from 'puppeteer';
 import type {Target} from 'puppeteer-core';
 import sinon from 'sinon';
 
+import {McpContext} from '../src/McpContext.js';
 import {NetworkFormatter} from '../src/formatters/NetworkFormatter.js';
 import {TextSnapshot} from '../src/TextSnapshot.js';
 import type {HTTPResponse} from '../src/third_party/index.js';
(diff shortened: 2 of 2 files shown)


src/McpContext.ts
```diff
@@ -562,11 +562,34 @@ export class McpContext implements Context {
...
+      if (type === 'page') return true;
+      if (this.#options.experimentalIncludeAllPages) {
+        return type === 'background_page' || type === 'webview';
+      }
+      return false;
+    });
+    const pageResults = await Promise.all(
+      pageTargets.map(async target => {
+        try {
+          return await target.page();
```
OrKoN: Thanks for the PR. I believe this will take 180 sec to time out which is too long to be practical. Do you have an automated test to demonstrate that the fix works?

OrKoN: In addition, I think it will time out during connect/launch within Puppeteer and not at this stage.

OrKoN (changes requested):
cc @nroscino

src/McpContext.ts
20syldev: Thanks for the review! Each target.page() call is now wrapped in a Promise.race with a 5s timeout so frozen targets are skipped instead of blocking for 180s. 
I also added a test that injects a fake target whose page() never resolves and asserts that createPagesSnapshot() still completes quickly and enumerates the healthy pages.

OrKoN (changes requested):
Please add a test demonstrating that it works with actual frozen tabs in the browser and not only in the mock. As I mentioned on https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1230#issuecomment-4221939980 I have troubles reproducing the issue. Do you have additional details for the manual reproduction?

20syldev:
> Please add a test demonstrating that it works with actual frozen tabs in the browser and not only in the mock. As I mentioned on [#1230 (comment)](https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1230#issuecomment-4221939980) I have troubles reproducing the issue. Do you have additional details for the manual reproduction?

### How to reproduce

The issue requires a **dead renderer** — `Page.setWebLifecycleState("frozen")` via CDP is not enough because the renderer still responds to CDP commands. The actual scenario is a tab whose renderer process was killed (discarded by Chrome's memory saver or crashed).

**Steps:**
1. Launch Chrome with remote debugging:
   ```bash
   google-chrome --remote-debugging-port=9222 --user-data-dir=/tmp/chrome-debug
   ```
2. Open a few tabs, then navigate one to `chrome://crash` (kills the renderer)
3. Run this script:
   ```js
   import { connect } from 'puppeteer';

   const browser = await connect({ browserURL: 'http://localhost:9222' });
   console.log('Connected, calling browser.pages()...');
   const pages = await browser.pages(); // hangs here for 180s
   console.log(pages.length);
   ```
4. `browser.pages()` → **hangs indefinitely** (180s protocolTimeout)

This only happens with `puppeteer.connect()`, not `puppeteer.launch()`, because launch auto-attaches to all targets and caches `Page` instances. On connect, `target.page()` has to initialize from scratch, sending `Network.enable` to a dead renderer that never responds.

### Added test

The new integration test (`should enumerate pages when a tab has a crashed renderer`) reproduces exactly this:
- Launches Chrome, creates an active page and a crashed page (`chrome://crash`)
- Disconnects and reconnects via WebSocket
- Creates an `McpContext` (which calls `#getAllPages()` internally)
- Asserts it completes in under 20s and finds the active pages

```bash
$ NODE_TEST_REPORTER=spec npm test -- tests/McpContext.test.ts 2>&1 | grep -A2 "crashed renderer"
```
```bash
✔ should enumerate pages when a tab has a crashed renderer (10459.212304ms)
```

~10s = two `#getAllPages()` calls during init, each hitting the 5s per-target timeout on the crashed tab. Without the fix, this would be 360s+ (2 × 180s protocolTimeout).

src/McpContext.ts
```diff
@@ -562,11 +562,47 @@ export class McpContext implements Context {
...
 
+    // Enumerate targets individually instead of calling browser.pages() so
+    // that a single frozen/discarded background tab that times out on
+    // Network.enable cannot abort the entire page enumeration (see #1230).
     const allTargets = this.browser.targets();
+    const pageTargets = allTargets.filter(target => {
+      const type = target.type();
+      if (type === 'page') return true;
+      if (this.#options.experimentalIncludeAllPages) {
+        return type === 'background_page' || type === 'webview';
```
barretlee, line 574: Is the enumeration here sufficient? type: What are the categories in total?

20syldev: The CDP target types are: `page`, `background_page`, `service_worker`, `webview`, `worker`, `shared_worker`, `browser`, and `other`.

Here only `page` is enumerated by default (same as what `browser.pages()` did before), and `background_page` + `webview` are added when `experimentalIncludeAllPages` is on — that's the pre-existing opt-in. The rest (`service_worker`, `worker`, etc.) aren't navigable pages so `target.page()` wouldn't return anything useful for them.

OrKoN:
I do not think crashed renderer makes the target frozen or unresponsive. https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1230 was already fixed in Chrome.

PR #1845, opened 2026-04-10
feat(webmcp): Add experimental tool to list WebMCP tools page exposes

beaufortfrancois:
image

```
npx @modelcontextprotocol/inspector node build/src/bin/chrome-devtools-mcp.js --channel=canary --experimentalWebmcp --chrome-arg='--enable-features=WebMCPTesting,DevToolsWebMCPSupport'
```

image

cc @OrKoN


src/tools/ToolDefinition.ts
```diff
@@ -199,6 +201,7 @@ export type Context = Readonly<{
   getExtensionServiceWorkerId(
     extensionServiceWorker: ExtensionServiceWorker,
   ): string | undefined;
+  getWebMcpTools(page: ContextPage): WebMcpTool[];
```
OrKoN: we do not need to add it if the method is not used by tools/*

src/PageCollector.ts
```diff
@@ -412,3 +414,59 @@ export class NetworkCollector extends PageCollector<HTTPRequest> {
...
+        this.#page.emit('webmcpToolAdded', tool);
+      }
+    };
+  }
+
+  async subscribe() {
+    this.#session.on('WebMCP.toolsAdded', this.#onToolsAdded);
+    try {
+      // @ts-expect-error WebMCP is an experimental domain
+      await this.#session.send('WebMCP.enable');
```
OrKoN: just to confirm: we do not need to use CDP and we do not need a collector, let's just use the Puppeteer API.

src/tools/ToolDefinition.ts
beaufortfrancois: Gotcha

src/PageCollector.ts
beaufortfrancois: I've removed CDP and collector in latest commit.

beaufortfrancois:
I believe https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1845 is now ready for review. I was able to make tests pass locally by using next version of puppeteer.

src/tools/pages.ts
```diff
@@ -28,6 +28,7 @@ export const listPages = defineTool(args => {
     handler: async (_request, response) => {
       response.setIncludePages(true);
       response.setListInPageTools();
+      response.setListWebMcpTools();
```
OrKoN, line 31: let's also call setListWebMcpTools in the select_page tool

beaufortfrancois: Shall we also do this in `navigate_page` tool then?

tests/tools/webmcp.test.ts
```diff
@@ -0,0 +1,86 @@
...
+        );
+
+        await listWebMcpTools.handler(
+          {params: {}, page: context.getSelectedMcpPage()},
+          response,
+          context,
+        );
+
+        const formattedResponse = await response.handle('test', context);
+        const textContent = getTextContent(formattedResponse.content[0]);
```
OrKoN: let's use text snapshots and assert the entire response and not only that it matches a pattern. In general, I think in webmcp.test.ts we should only test if the McpResponse has correct flags set and then let's test formatting with snapshots in tests/McpResponse.test.ts

src/tools/pages.ts
OrKoN: yes, I think this would be good too

src/McpResponse.ts
```diff
@@ -232,6 +235,12 @@ export class McpResponse implements Response {
     }
   }
 
+  setListWebMcpTools(): void {
+    if (this.#args.experimentalWebmcp) {
```
OrKoN: perhaps let's check the this.#args.experimentalWebmcp when we are fetching webmcp tools so that we can test that tools correctly configure the webmcp behavior irrespective of the arg value.

tests/tools/webmcp.test.ts
beaufortfrancois: 1. Which flags should be set in the McpResponse?
2. Snapshots in tests/McpResponse.test.ts would include id like frame _id and _loaderId which are unique every time. How do I get around this?

src/tools/pages.ts
beaufortfrancois: Done

src/McpResponse.ts
beaufortfrancois: Do you mean removing this `if (this.#args.experimentalWebmcp) {` in `setListWebMcpTools()` and moving it to `format()` instead?

tests/tools/webmcp.test.ts
OrKoN: 1. the `setListWebMcpTools` flag.
2. we should not return the frame or frame id in text or json responses. I guess we could return frame.url() but I would also just omit the whole frame data as the MCP server has no concept of frames right now so it would be useless to clients.

src/McpResponse.ts
OrKoN: yes, but moving it to `handle()` instead where the decision to fetch the tools is made

tests/tools/webmcp.test.ts
beaufortfrancois: 1. The `setListWebMcpTools` is now set unconditionally. What/how can I test it properly?
2. Done

src/McpResponse.ts
beaufortfrancois: done

tests/tools/webmcp.test.ts
OrKoN: We can expose a getter and verify that it reports the correct value for all tools that set it.

beaufortfrancois: Done

src/McpResponse.ts
```diff
@@ -488,6 +499,12 @@ export class McpResponse implements Response {
       page.inPageTools = inPageTools;
     }
 
+    let webmcpTools: WebMCPTool[] | undefined;
+    if (this.#listWebMcpTools && this.#args.experimentalWebmcp) {
+      const page = this.#page ?? context.getSelectedMcpPage();
+      webmcpTools = context.getWebMcpTools(page);
```
OrKoN: let's move getWebMcpTools getter to McpPage so that this becomes `webmcpTools page.getWebMcpTools();`?

OrKoN (commented):
I think this is the last comment

src/McpResponse.ts
beaufortfrancois: Done

OrKoN (approved):

tests/McpResponse.test.ts
```diff
@@ -1455,3 +1455,130 @@ describe('replaceHtmlElementsWithUids', () => {
     }
   });
 });
+
+describe('webmcp', () => {
```
OrKoN, line 1459: we do not have expectation files to skip tests here, so perhaps we can `.skip` it here

OrKoN (approved):

tests/McpResponse.test.ts
beaufortfrancois: Since listing tools is available in Chrome Stable, we don't need `.skip` as discussed offline. https://chromiumdash.appspot.com/commit/18ae39a0c878c36745294b9980f9f4fc6c6b47f8

PR #1848, opened 2026-04-10
feat: support DevTools header redactions as an option

OrKoN:
This PR adds a CLI flag to enable redacting network headers in the same way they are redacted in DevTools. Note that sometimes it might prevent the agent from properly analysing network issues. Pass `--redact-headers=false` to revert to the previous behavior.

### DIFF
--- a/src/formatters/NetworkFormatter.ts
+++ b/src/formatters/NetworkFormatter.ts
@@ -6,7 +6,11 @@
 
 import {isUtf8} from 'node:buffer';
 
-import type {HTTPRequest, HTTPResponse} from '../third_party/index.js';
+import {
+  DevTools,
+  type HTTPRequest,
+  type HTTPResponse,
+} from '../third_party/index.js';
 
 const BODY_CONTEXT_SIZE_LIMIT = 10000;
 
@@ -21,6 +25,7 @@ export interface NetworkFormatterOptions {
     data: Uint8Array<ArrayBufferLike>,
     filename: string,
   ) => Promise<{filename: string}>;
+  redactNetworkHeaders: boolean;
 }
 
 interface NetworkRequestConcise {
@@ -150,6 +155,20 @@ export class NetworkFormatter {
     };
   }
 
+  #redactNetworkHeaders(
+    headers: Record<string, string>,
+  ): Record<string, string> {
+    const headersList = Object.entries(headers).map(item => {
+      return {name: item[0], value: item[1]};
+    });
+    const redacted =
+      DevTools.NetworkRequestFormatter.sanitizeHeaders(headersList);
+    return redacted.reduce<Record<string, string>>((acc, item) => {
+      acc[item.name] = item.value;
+      return acc;
+    }, {});
+  }
+
   toJSONDetailed(): NetworkRequestDetailed {
     const redirectChain = this.#request.redirectChain();
     const formattedRedirectChain = redirectChain.reverse().map(request => {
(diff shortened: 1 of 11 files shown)


SandyMtz9 (approved):

OrKoN:
@natorion I currently think it makes sense to add this as an option aligning with DevTools. PTAL

src/formatters/NetworkFormatter.ts
```diff
@@ -159,16 +178,21 @@ export class NetworkFormatter {
...
     return {
       ...this.toJSON(),
-      requestHeaders: this.#request.headers(),
+      requestHeaders: this.#options.redactNetworkHeaders
+        ? this.#redactNetworkHeaders(this.#request.headers())
+        : this.#request.headers(),
       requestBody: this.#requestBody,
       requestBodyFilePath: this.#requestBodyFilePath,
-      responseHeaders: this.#request.response()?.headers(),
+      responseHeaders: this.#options.redactNetworkHeaders
```
Lightning00Blade: This may be fine but just wanted to point out that here when we redact when there are no response we create an empty object, which may have different behaviour when we show it as text.

tests/utils.ts
```diff
@@ -170,9 +171,11 @@ export function getMockRequest(
       return options.resourceType ?? 'document';
     },
     headers(): Record<string, string> {
-      return {
-        'content-size': '10',
-      };
+      return (
+        options.headers ?? {
```
Lightning00Blade, line 175: Let's spread them after the options.

tests/McpContext.test.js.snapshot
```diff
@@ -6,8 +6,9 @@ exports[`McpContext > should include detailed network request in structured cont
     "url": "http://example.com/detail",
     "status": "pending",
     "requestHeaders": {
-      "content-size": "10"
-    }
+      "content-size": "<redacted>"
+    },
+    "responseHeaders": {}
```
Lightning00Blade: This is what I was talking in the other comment.

Lightning00Blade (approved):

natorion:
Is this enabled or disabled by default? I am not sure we want to have redaction enabled by default - that sounds a like a box of surprises.

If enabled we should also clearly indicate/state how it can turned off via the settings. Let's make srue that agents can self fix the problem.

tests/utils.ts
OrKoN: did you mean the following?

```
options.headers ? {...options.headers} : {
          'content-size': '10',
        }
      );
```

OrKoN:
> Is this enabled or disabled by default? I am not sure we want to have redaction enabled by default - that sounds a like a box of surprises.
> 
> If enabled we should also clearly indicate/state how it can turned off via the settings. Let's make srue that agents can self fix the problem.

Changed to disabled by default.

tests/utils.ts
```diff
@@ -170,9 +171,11 @@ export function getMockRequest(
...
+          'content-size': '10',
+        }
+      );
```
Lightning00Blade, line 178: More like this, as there is always some headers.
```suggestion
      return (
        {
          'content-size': '10',
          ...options.headers
        }
      );
```

tests/utils.ts
Lightning00Blade: Left another comment as I couldn't select the correct lines.

tests/utils.ts
OrKoN: I think I would rather expect the provided headers to replace the default ones as otherwise it is difficult to unset default headers if needed

Lightning00Blade: One could just pass `undefined` to unset any header value.

Lightning00Blade: This is non-blocking so feel free to merge as is.

natorion (approved):

PR #1851, opened 2026-04-11
feat: add Claude plugin browser URL config

### LINKED ISSUE #1232: Plugin mode: support --browserUrl via pluginConfigs
## Problem

When using `chrome-devtools-mcp` as a **Claude Code plugin** (`chrome-devtools-mcp@claude-plugins-official`), the MCP server is launched as:

```
npx chrome-devtools-mcp@latest
```

There is no way to pass `--browserUrl http://127.0.0.1:9222` (or other CLI flags) through the plugin configuration system. This means the plugin cannot connect to an existing Chrome instance with remote debugging enabled.

### Current workaround

Disable the plugin and configure a standalone MCP server in `.mcp.json`:

```json
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": ["chrome-devtools-mcp@latest", "--browserUrl", "http://127.0.0.1:9222"]
    }
  }
}
```

This works for tools, but **loses the plugin's bundled skills** (a11y-debugging, debug-optimize-lcp, chrome-devtools, troubleshooting).

## Proposed solution

Support `--browserUrl` (and potentially other CLI flags like `--autoConnect`, `--channel`) as user-configurable options through Claude Code's `pluginConfigs` mechanism in `settings.json`:

```json
{
  "pluginConfigs": {
    "chrome-devtools-mcp@claude-plugins-official": {
      "mcpServers": {
        "chrome-devtools": {
          "browserUrl": "http://127.0.0.1:9222"
        }
      }
    }
  }
}
```

This would allow users to connect to an existing Chrome instance while retaining the plugin's skills.

## Environment

- OS: Windows 11
- Claude Code plugin: `chrome-devtools-mcp@claude-plugins-official`
- chrome-devtools-mcp version: 0.19.0
- Chrome remote debugging enabled on `127.0.0.1:9222`

officialasishkumar:
## Summary

- add a Claude plugin `browserUrl` user config option for connecting to a running debuggable Chrome instance
- pass the option through to `chrome-devtools-mcp` with `--browser-url`, while a blank value keeps the default automatic Chrome launch behavior
- document the Claude Code `pluginConfigs` settings shape for the plugin option

Fixes #1232

## Testing

- parsed `.claude-plugin/plugin.json` with `JSON.parse`
- `npm run check-format`
- `npm run gen`


OrKoN:
Thanks for the PR, we should implement the suggestion from https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1981#discussion_r3165809246 to generate all configuration options.

OrKoN (changes requested):
We should update the generator to generate the userConfig options as suggested in https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1981#discussion_r3165809246

officialasishkumar:
Updated this to generate plugin userConfig from the CLI options and wire the generated values through env defaults, so browserUrl and future options are not maintained by hand.

.claude-plugin/plugin.json
```diff
@@ -7,7 +7,219 @@
       "command": "npx",
       "args": [
         "chrome-devtools-mcp@latest"
-      ]
+      ],
+      "env": {
```
OrKoN: I think we should keep using arguments for the configuration and do not introduce environment variables. I think the initial version of this PR used arguments.

OrKoN (changes requested):

.claude-plugin/plugin.json
OrKoN: I think I actually saw this in another PR https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1981

officialasishkumar: Updated this in 8b3a6e2: generated plugin user config now stays on CLI args, and the env-default path was removed.

.github/plugin/plugin.json
```diff
@@ -6,8 +6,253 @@
     "chrome-devtools": {
       "command": "npx",
       "args": [
-        "chrome-devtools-mcp@latest"
+        "chrome-devtools-mcp@latest",
```
OrKoN, line 9: I do not seem to be able to make sure of the plugin configuration in VSCode / Copilot. Is it supported by github plugins?

OrKoN:
Thanks for the PR. I will close it as stale as we are working on an alternative configuration solution that would not require host-specific support.

PR #1852, opened 2026-04-11
fix: ignore unmapped PerformanceIssue events

### LINKED ISSUE #1850: Unhandled Audits PerformanceIssue spams stderr: 'No handler registered for issue code PerformanceIssue'
## Summary

`chrome-devtools-mcp@0.21.0` logs this warning repeatedly to stderr when Chrome emits an Audits issue with code `PerformanceIssue`:

```text
No handler registered for issue code PerformanceIssue
```

In long-running agent sessions this can flood the terminal.

## What I found

In the bundled DevTools frontend code, `PerformanceIssue` is present in the registered `Audits.InspectorIssueCode` enum, but there is no corresponding entry in `issueCodeHandlers`.

### Evidence from the installed 0.21.0 bundle

The generated bundle includes:

```js
inspectorBackend.registerEnum("Audits.InspectorIssueCode", {
  ...,
  PerformanceIssue: "PerformanceIssue",
  ...
});
```

But `issueCodeHandlers` does not include a `PerformanceIssue` handler, and the fallback path is:

```js
function createIssuesFromProtocolIssue(issuesModel, inspectorIssue) {
    const handler = issueCodeHandlers.get(inspectorIssue.code);
    if (handler) {
        return handler(issuesModel, inspectorIssue);
    }
    console.warn(`No handler registered for issue code ${inspectorIssue.code}`);
    return [];
}
```

## Repro shape

1. Start `chrome-devtools-mcp@0.21.0`
2. Attach it to a live Chrome session
3. Open / interact with a page that causes Chrome DevTools to emit `Audits.issueAdded` with code `PerformanceIssue`
4. Observe repeated stderr warnings:

```text
No handler registered for issue code PerformanceIssue
```

## Expected

One of:

- add a proper `PerformanceIssue` handler, or
- intentionally ignore / downgrade `PerformanceIssue` without warning spam, if this issue type is not meant to be surfaced yet.

## Actual

The server warns on every occurrence, which is noisy enough to overwhelm terminal-based agent workflows.

## Environment

- `chrome-devtools-mcp`: 0.21.0
- Chrome: current stable on macOS
- Invocation path: via `npm exec chrome-devtools-mcp@latest ...`

## Local workaround

I locally patched the generated bundle to suppress the warning only for `PerformanceIssue`, and that eliminated the spam without affecting other unknown issue-code warnings.

officialasishkumar:
## Summary

- ignore Audits `PerformanceIssue` events before passing them to the DevTools IssuesManager mapper, which currently has no handler for that issue code
- preserve the existing mapper/logging path for other issue codes
- add regression coverage that `PerformanceIssue` is ignored without collecting an issue or writing a console warning

Fixes #1850

## Testing

- `npm run build`
- `npm run test:no-build -- tests/PageCollector.test.ts`
- `npm run check-format`

### DIFF
--- a/src/PageCollector.ts
+++ b/src/PageCollector.ts
@@ -342,6 +342,11 @@ class PageEventSubscriber {
 
   #onIssueAdded = (inspectorIssue: Issue) => {
     try {
+      // DevTools currently defines this protocol issue code but has no
+      // IssuesManager handler for it, so calling into the mapper only warns.
+      if (String(inspectorIssue.code) === 'PerformanceIssue') {
+        return;
+      }
       const issue = DevTools.createIssuesFromProtocolIssue(
         null,
         // @ts-expect-error Protocol types diverge.
--- a/tests/PageCollector.test.ts
+++ b/tests/PageCollector.test.ts
@@ -358,6 +358,35 @@ describe('ConsoleCollector', () => {
     assert.equal(data.length, 2);
   });
 
+  it('silently ignores unmapped PerformanceIssue events', async () => {
+    const browser = getMockBrowser();
+    const page = (await browser.pages())[0];
+    const warnStub = sinon.stub(console, 'warn');
+
+    const collector = new ConsoleCollector(browser, collect => {
+      return {
+        devtoolsAggregatedIssue: issue => {
+          collect(issue);
+        },
+      } as ListenerMap;
+    });
+    await collector.init([page]);
+
+    const performanceIssue = {
+      code: 'PerformanceIssue',
+      details: {
+        performanceIssueDetails: {
+          performanceIssueType: 'DocumentCookie',
+        },
+      },
+    } as unknown as Protocol.Audits.InspectorIssue;
+
+    page.emit('issue', performanceIssue);
+
+    assert.equal(collector.getData(page).length, 0);
+    sinon.assert.notCalled(warnStub);
+  });
+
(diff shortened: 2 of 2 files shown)


src/PageCollector.ts
```diff
@@ -342,6 +342,11 @@ class PageEventSubscriber {
   #onIssueAdded = (data: Protocol.Audits.IssueAddedEvent) => {
     try {
       const inspectorIssue = data.issue;
+      // DevTools currently defines this protocol issue code but has no
+      // IssuesManager handler for it, so calling into the mapper only warns.
+      if (String(inspectorIssue.code) === 'PerformanceIssue') {
```
Lightning00Blade, line 347: Note I have landed a change upstream in DevTools - https://crrev.com/c/7761323
to get the supported list so we don't have to hard code this bit.

Lightning00Blade (approved):
Thanks for the PR,

I think we can merge after the rebase and update to tests.
And remove the the hard coded bit after we update the version of devtools.

src/PageCollector.ts
officialasishkumar: Rebased onto latest `main` and updated the regression test for the current page issue event path. Kept the guard isolated until the DevTools bundle with the supported-list change is pulled in here.

Lightning00Blade (approved):

PR #1853, opened 2026-04-12
feat: report new URL after actions that trigger navigation

### LINKED ISSUE #243: Actions that result in a navigation should report the new page URL in the response
We have the waiting logic to detect navigations and wait till they finish but we never report the navigated page URL back to the client. We should do it so that there is no need to re-query the pages list.

rogeroliveira84:
## Summary

- Input tools (`click`, `fill`, `press_key`, `hover`, `drag`, `type_text`, `fill_form`, `click_at`) and `evaluate_script` now append a `Page navigated to <url>.` line to the response when the action triggers a cross-document navigation.
- `WaitForHelper.waitForEventsAfterAction` returns `{navigated: boolean}` instead of `void`, surfacing the navigation signal that was already being detected internally.
- No change to `navigate_page` or `new_page` since they already report the URL explicitly.

Fixes #243

## Why

Today, if a `click` causes a page navigation, the response says *"Successfully clicked on the element"* with no indication that the page URL changed. The agent has to make an extra `list_pages` call to discover where it landed. This saves that round-trip for every navigation-triggering action.

## Design

The existing `waitForNavigationStarted` in `WaitForHelper` already knows whether a cross-document navigation started. We propagate that signal as `{navigated: boolean}` through the return value of `waitForEventsAfterAction` → `McpPage` → `ContextPage` interface, and let each handler append the URL line when `navigated` is true.

Same-document (history API) navigations remain filtered out by the existing `waitForNavigationStarted` logic, matching current behavior. Click-opens-new-tab is a separate concern (#367).

## Test plan

- [x] New test: click on a link that causes navigation → response includes `Page navigated to <url>.`
- [x] New test: click on a button that doesn't navigate → no navigation line in response
- [x] Full test suite (563 tests) passes
- [x] TypeScript typecheck clean
- [x] ESLint + Prettier clean


src/WaitForHelper.ts
```diff
@@ -159,9 +161,19 @@ export class WaitForHelper {
     } finally {
       this.#abortController.abort();
     }
+
+    return {navigated};
```
Lightning00Blade, line 165: I think we can simplify this can just provide the URL here if it was changed.
```suggestion
    return {
    ...(navigated ? {url: this.#page.url()} : {})
    };
```

src/WaitForHelper.ts
```diff
@@ -159,9 +161,19 @@ export class WaitForHelper {
...
   }
 }
 
+export interface WaitForEventsResult {
+  /**
+   * Whether a cross-document navigation started and finished during the
+   * action. Same-document (history API) navigations are not reported.
+   */
+  navigated: boolean;
```
Lightning00Blade, line 174: With the above change this will become 
```suggestion
  navigatedToUrl?: string;
```

src/WaitForHelper.ts
```diff
@@ -159,9 +161,19 @@ export class WaitForHelper {
...
     }
+
+    return {navigated};
   }
 }
 
+export interface WaitForEventsResult {
+  /**
+   * Whether a cross-document navigation started and finished during the
+   * action. Same-document (history API) navigations are not reported.
```
Lightning00Blade, line 172: This is wrong we use Puppeteer waitForNavigation and that does support same-document navigation 
https://pptr.dev/api/puppeteer.page.waitfornavigation

src/tools/input.ts
```diff
@@ -42,6 +43,16 @@ function handleActionError(error: unknown, uid: string) {
   );
 }
 
+function appendNavigationIfAny(
```
Lightning00Blade, line 46: Let's rename it to `appendWaitForResult` to match the interface we provide.
We can also move this into the WaitForHelper file as we can use in other places as well.

src/tools/script.ts
```diff
@@ -77,11 +77,15 @@ Example with arguments: \`(el) => {
...
-        await context
-          .getSelectedMcpPage()
-          .waitForEventsAfterAction(async () => {
-            await performEvaluation(worker, fnString, [], response);
-          });
+        const selectedPage = context.getSelectedMcpPage();
+        const result = await selectedPage.waitForEventsAfterAction(async () => {
+          await performEvaluation(worker, fnString, [], response);
+        });
+        if (result.navigated) {
```
Lightning00Blade, line 84: This should use the helper.

src/tools/script.ts
```diff
@@ -101,9 +105,12 @@ Example with arguments: \`(el) => {
 
         const evaluatable = await getPageOrFrame(page, frames);
 
-        await mcpPage.waitForEventsAfterAction(async () => {
+        const result = await mcpPage.waitForEventsAfterAction(async () => {
           await performEvaluation(evaluatable, fnString, args, response);
         });
+        if (result.navigated) {
```
Lightning00Blade, line 111: Same here.

Lightning00Blade (commented):
Thanks for the PR, 
there are some comments that would need to be address before we can merge.

src/WaitForHelper.ts
Lightning00Blade: Actually let's call it `navigatedToUrl` to be more explicit.

masamaru0513:
Hi @rogeroliveira84, thanks for the original PR and the groundwork here! I noticed the review feedback from @Lightning00Blade hadn't been addressed yet, so I went ahead and opened #1954 incorporating those suggestions (`navigatedToUrl` naming, shared helper in `WaitForHelper.ts`, etc.).

rogeroliveira84:
Hi @Lightning00Blade I have on going work for that was planing to push later on, but once I do you can also check comment on it and see how it goes... cheers!

rogeroliveira84:
Hi @Lightning00Blade I pushed changes, could review pls. Thanks.

OrKoN:
@rogeroliveira84 could you please rebase?

rogeroliveira84:
> @rogeroliveira84 could you please rebase?

Sure, it is rebased now.

OrKoN:
Actually, it looks like there is another PR about this feature https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1954

rogeroliveira84:
> Actually, it looks like there is another PR about this feature #1954

is this one going ahead?

Lightning00Blade (approved):
LGTM

rogeroliveira84:
> LGTM

Can approve for merging?

PR #1855, opened 2026-04-12
fix: show extension popup pages in connected browser mode

edxeth:
Connected-browser mode was still filtering out chrome-extension:// page targets, so wallet popups never showed up in list_pages even though Chrome exposed them over remote debugging.

This keeps extension page/service worker visibility in connected sessions, but still leaves extension management tools disabled unless chrome-devtools-mcp launched the browser itself.

Also adds a regression test for a popup opened over remote debugging.


Lightning00Blade:
Hey @edxeth,

I think this fall under the same changes we are planning for extension mentioned [here](https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1171#issuecomment-4177586810).
So I will close this PR, but please stay tuned!

PR #1861, opened 2026-04-14
chore: some build and test fixes

Lightning00Blade:
Under some cases one could not re-build correctly due to the directory being remove.
And we can save some CPU cycles in test to disable the update checker.


OrKoN (approved):
Thanks!

PR #1867, opened 2026-04-15
feat: ensure extensions for file outputs

### LINKED ISSUE #1864: Enforce a file extension for files saved by the MCP server
By ensuring that file extensions are set on files written by the MCP server, it makes it less likely that other types of files are overwritten by the MCP server. To be used on conjunction with https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1860

OrKoN:
This PR ensures the extensions for the file outputs of different types minimizing the chance of misuse. The input filePath, thus, might be modified but it should not be an issue for clients as the final output path is returned to the clients in the response.

Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1864

### DIFF
--- a/src/utils/files.ts
+++ b/src/utils/files.ts
@@ -24,3 +24,11 @@ export async function saveTemporaryFile(
     throw new Error('Could not save a file', {cause: err});
   }
 }
+
+export function ensureExtension(filepath: string, extension: string): string {
+  const normalizedExtension = extension.startsWith('.')
+    ? extension
+    : `.${extension}`;
+  const ext = path.extname(filepath);
+  return filepath.slice(0, filepath.length - ext.length) + normalizedExtension;
+}
--- a/src/utils/files.ts
+++ b/src/utils/files.ts
@@ -25,10 +25,10 @@ export async function saveTemporaryFile(
   }
 }
 
-export function ensureExtension(filepath: string, extension: string): string {
-  const normalizedExtension = extension.startsWith('.')
-    ? extension
-    : `.${extension}`;
+export function ensureExtension(
+  filepath: string,
+  extension: `.${string}`,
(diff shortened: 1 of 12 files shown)


src/utils/files.ts
```diff
@@ -24,3 +24,11 @@ export async function saveTemporaryFile(
     throw new Error('Could not save a file', {cause: err});
   }
 }
+
+export function ensureExtension(filepath: string, extension: string): string {
```
Lightning00Blade: Not a blocker, more of an FYI - we could maybe type the function like:
```suggestion
export function ensureExtension(filepath: string, extension: `.${string)`: string {
```
To remove the need for normalization and rely on TypeChecks.

tests/utils/files.test.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
Lightning00Blade:
```suggestion
 * Copyright 2026 Google LLC
```

Lightning00Blade (commented):
LGTM with comments 

- Did you miss the memory snapshot place?
- Feature work: Should we ban the use of fs/path in the repo and create our wrappers to manage this in one place?

OrKoN:
> Did you miss the memory snapshot place?

Yep, fixed!

> Feature work: Should we ban the use of fs/path in the repo and create our wrappers to manage this in one place?

Note sure, I think there could legitimate uses of fs/path but if we can lint all places where we read/write tools for tools (excluding daemons, cli, metrics). I think it would be great. Yeah maybe we can just ban it in tools/. With Puppeteer APIs it is not so straightforward.

Lightning00Blade (approved):

PR #1874, opened 2026-04-16
chore: implement memory snapshot information

Lightning00Blade:
WIP of exposing a tool for debugging memory issue directly with the MCP


src/formatters/HeapSnapshotFormatter.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
OrKoN:
```suggestion
 * Copyright 2026 Google LLC
```

src/tools/memory.ts
```diff
@@ -34,3 +34,37 @@ export const takeMemorySnapshot = definePageTool({
     );
   },
 });
+
+export const exploreMemorySnapshot = defineTool({
+  name: 'explored_memory_snapshot',
```
OrKoN:
```suggestion
  name: 'load_memory_snapshot',
```

src/tools/memory.ts
```diff
@@ -34,3 +34,37 @@ export const takeMemorySnapshot = definePageTool({
...
+  description: 'Explose ',
```
OrKoN:
```suggestion
  description: 'Loads a memory heapsnapshot and returns snapshot summary',
```

src/tools/memory.ts
```diff
@@ -34,3 +34,37 @@ export const takeMemorySnapshot = definePageTool({
...
+    pageSize: zod.number().optional().describe('Page size for pagination.'),
+    pageIdx: zod.number().optional().describe('Page index for pagination.'),
+  },
+  handler: async (request, response, context) => {
+    const snapshot = await context.getHeapSnapshotProxy(
+      request.params.filePath,
+    );
+    const stats = await snapshot.getStatistics();
+
+    response.appendResponseLine(
```
OrKoN: we should probably also set it via the response object

src/tools/memory.ts
```diff
@@ -34,3 +34,37 @@ export const takeMemorySnapshot = definePageTool({
...
+  name: 'explored_memory_snapshot',
+  description: 'Explose ',
+  annotations: {
+    category: ToolCategory.PERFORMANCE,
+    readOnlyHint: true,
+  },
+  schema: {
+    filePath: zod.string().describe('A path to a .heapsnapshot file to read.'),
+    pageSize: zod.number().optional().describe('Page size for pagination.'),
+    pageIdx: zod.number().optional().describe('Page index for pagination.'),
```
OrKoN: I think let's do the aggregates output in a separate tool call.

src/McpContext.ts
```diff
@@ -913,4 +914,45 @@ export class McpContext implements Context {
   getExtension(id: string): InstalledExtension | undefined {
     return this.#extensionRegistry.getById(id);
   }
+
+  async getHeapSnapshotProxy(
+    heapsnapshotPath: string,
+  ): Promise<DevTools.HeapSnapshotModel.HeapSnapshotProxy.HeapSnapshotProxy> {
+    const workerProxy =
```
OrKoN: I am thinking let's keep it in a map by snapshot file path? let's wait for an explicit command from the client to dispose (can be implement in the future PRs).

src/tools/memory.ts
Lightning00Blade: I guess here the comment is for the Statistics.

src/McpContext.ts
Lightning00Blade: Actually I am currently implementing that. I will do the dispose sepraterly.

src/formatters/HeapSnapshotFormatter.ts
```diff
@@ -0,0 +1,56 @@
+/**
```
OrKoN, line 1: leaving a comment to not forget: let's add a hidden experimental feature flag for new tools and land as `chore: ` to allow easier/concurrent work on tools after the infrastructure for loading/unloading snapshots is landed.

OrKoN (approved):

PR #1877, opened 2026-04-16
chore: in-page tool output: stash DOM elements and preprocess for serialization

wolfib:
Improvements for handling in-page tool responses. In order to successfully pass an in-page tool response from the page context to the MCP server, the response needs to be serializable. The code walks the response object an performs the following changes:

- DOM elements are stashed onto the window object and replaced with an ID. On the MCP server side this ID is used to map back to the corresponding UID in the page snapshot generated from the accessibility tree.
- Circular references are replaced with a string.
- Class instances (which can be complex or non-serializable) are replaced with a string.
- Functions are replaced with a string.

If the in-page tool response contains DOM elements which are not part of the page snapshot, a new snapshot is created add the missing elements are added explicitly.


src/tools/inPage.ts
```diff
@@ -141,14 +143,155 @@ export const executeInPageTool = definePageTool({
...
+
+    const elementHandles: ElementHandle[] = [];
+    for (let i = 0; i < (result.stashed ?? 0); i++) {
+      const elementHandle = await request.page.pptrPage.evaluateHandle(
+        index => {
+          return window.__dtmcp?.stashedElements?.[index] ?? null;
+        },
+        i,
+      );
+      elementHandles.push(elementHandle as ElementHandle);
```
OrKoN, line 230: do we need the cast? it looks like it sometimes might return null or a jshandle?

src/tools/inPage.ts
```diff
@@ -141,14 +143,155 @@ export const executeInPageTool = definePageTool({
         }
         const toolResult = await window.__dtmcp.executeTool(name, args);
 
+        const stashDOMElement = (el: Element) => {
```
OrKoN, line 146: Can we move this to McpPage so that tools do not need so many new methods on the ToolDefinition interfaces?

src/McpContext.ts
```diff
@@ -180,14 +184,14 @@ export class McpContext implements Context {
   }
 
   resolveCdpElementId(
-    page: McpPage,
+    page: ContextPage,
```
OrKoN, line 187: why do we need to change to ContextPage? I think McpPage is right here.

src/tools/inPage.ts
```diff
@@ -141,14 +143,155 @@ export const executeInPageTool = definePageTool({
...
+      if (!backendNodeId) {
+        logger(`No backendNodeId for stashed DOM element with index ${index}`);
+        return {uid: `stashed-${index}`};
+      }
+      let cdpElementId = context.resolveCdpElementId(
+        request.page,
+        backendNodeId,
+      );
+      if (!cdpElementId) {
+        await context.createTextSnapshot(
```
OrKoN, line 246: do we create a new snapshot repeatedly? can we resolve ids and then add them to the snapshot at once?

src/McpContext.ts
wolfib: In `executeInPageCall` we want to replace the IDs referring to stashed elements with UIDs matching the page snapshot. If there is an ID where the UID lookup is unsuccessful, a new snapshot is created and the missing elements are added.
The order is
1) lookup ID -> unsuccessful
2) generate snapshot and add missing elements
3) lookup ID again -> successful

In order to achieve this order, I want to create a snasphot in the middle of the tool execution. So far, snapshot creation required an `McpPage`, and tools only have access to a `ContextPage`.

Option 1 (what I did): refactor snapshot generation to accept a `ContextPage` instead of an `McpPage` 

Option 2: allow `executeInPageTool` to get the `McpPage` (from the response), such that it can be passed as param for snapshot generation

Option 3: Try to move snapshot generation to `ContextPage`

WDYT?

OrKoN: `McpPage implements ContextPage` so I think you do not need to use the interface types in the implementation classes. I think you you move the whole logic to the McpPage per other comment there should be no issue here.

src/tools/inPage.ts
wolfib: If there are multiple `elementHandles`, all of them are passed to `createTextSnapshot`, which adds all of them to the snapshot. This means the next lookup should be successful and `createTextSnapshot` is not called again.

However, `response.includeSnapshot()` causes a 2nd snapshot to be created when assembling the response. If this is too expensive, I could add a parameter which tells `McpResponse` to re-use the existing snapshot.

src/tools/inPage.ts
wolfib: Done.

src/tools/inPage.ts
wolfib: Moved a bunch of logic to McpPage.

src/McpContext.ts
wolfib: Done, moved logic to McpPage, which means I don't need to change the parameter types in McpContext.

src/McpPage.ts
```diff
@@ -131,6 +136,186 @@ export class McpPage implements ContextPage {
...
+          result: processToolResult(toolResult),
+          stashed: window.__dtmcp?.stashedElements?.length ?? 0,
+        };
+      },
+      toolName,
+      params,
+      ...handles,
+    );
+
+    const elementHandles: ElementHandle[] = [];
```
OrKoN, line 255: do we dispose handles somewhere?

src/TextSnapshot.ts
```diff
@@ -0,0 +1,316 @@
...
+        : attachTarget.children
+          ? attachTarget.children.length
+          : 0;
+    };
+
+    if (extraHandles.length) {
+      page.extraHandles = extraHandles;
+    }
+    for (const handle of page.extraHandles) {
+      const extraNode = await createExtraNode(handle);
```
OrKoN, line 301: this loop is still mixing fetching and updates but I think it can be looked at in a follow up

OrKoN (approved):

src/tools/inPage.ts
wolfib: Refactored to move the snapshot creation out of the loop. This makes it clear that it's not called multiple times.

Issue #1797, opened 2026-04-02
lighthouse_audit always fails: Network.emulateNetworkConditions timed out (even with increased protocolTimeout)


CressZZ:
### Description of the bug

 ## Description                                                                                                          
   
  `lighthouse_audit` tool always fails with `Network.emulateNetworkConditions timed out` error, regardless of mode        
  (`navigation`/`snapshot`) or device (`mobile`/`desktop`).                                                             
                                                                                                                        
  Other tools like `performance_start_trace`, `navigate_page`, `list_pages` work perfectly fine on the same Chrome instance.
                                                                                                                          
  ## Environment                                                                                                        
                                                                                                                        
  - **chrome-devtools-mcp version**: 0.21.0 (latest, installed via `npx chrome-devtools-mcp@latest`)
  - **Chrome version**: 146.0.7680.165
  - **OS**: macOS (Darwin 25.3.0)                                                                                         
   
  ## Steps to Reproduce                                                                                                   
                                                                                                                        
  1. Start chrome-devtools-mcp (v0.21.0)                                                                                
  2. Navigate to any page (e.g., `https://web.dev`)
  3. Call `lighthouse_audit` with any combination of parameters — all fail                                                
   
  ## Error                                                                                                                
                                                                                                                        
  Network.emulateNetworkConditions timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.
                                                                                                                          
  ## Investigation                                                                                                      
                                                                                                                        
  ### It's NOT a timeout issue

  The Lighthouse bundle (`lighthouse-devtools-mcp-bundle.js:57060`) has:

  ```js
  var DEFAULT_PROTOCOL_TIMEOUT = 3e4; // 30 seconds
 ```
                                                                                                                          
  Manually patched to 18e4 (180s), restarted MCP — still fails. Chrome never responds to the command.                     
                                                                                                                          
  Other CDP commands work fine                                                                                            
                                                                                                                        
  - performance_start_trace → works                                                                                     
  - navigate_page → works
  - Lighthouse CLI (npx lighthouse) on the same machine → works                                                           
   
  Root Cause Hypothesis                                                                                                   
                                                                                                                        
  Lighthouse's internal CDP session sends Network.emulateNetworkConditions but Chrome never responds. Possibly:           
                                                                                                                        
  1. CDP session conflict — MCP's existing session holds the Network domain, blocking Lighthouse's session                
[...]

PR #1891, opened 2026-04-17
fix: skip unnecessary network reset during emulation restore

dhananjay6561:
solves #1797 

## Summary
Fixes a Lighthouse audit failure where `Network.emulateNetworkConditions` could hang during emulation restore.

## Root cause
`lighthouse_audit` always runs `restoreEmulation()` in a `finally` block.
During restore, `McpContext.emulate()` always called `page.emulateNetworkConditions(null)` when `networkConditions` was not set.
On some Chrome/CDP combinations, that reset call can time out even though no network emulation was active.

## What changed
- Added an internal `skipNetworkReset` option to `McpContext.emulate()`.
- Updated `restoreEmulation()` to skip the network reset when the page had no saved `networkConditions`.
- Kept all other emulation restore behavior unchanged.

## Validation
- `npm run test:no-build -- tests/McpContext.test.ts`
- `npm run test:no-build -- tests/tools/lighthouse.test.ts`

## Risk
Low. The change only avoids one unnecessary network-reset call when there was nothing to reset.


dhananjay6561:
@Lightning00Blade hi!
kindly review the pr!

dhananjay6561:
@zyzyzyryxy kindly review!

zyzyzyryxy:
Hello, thank you for your contribution! As described in the bug report comment, this PR doesn't fully fix the issue, just moves the goalpost. With your change I see timeout on `Network.setUserAgentOverride` instead of `Network.emulateNetworkConditions`

dhananjay6561:
@zyzyzyryxy my prev change avoided the timeout on Network.emulateNetworkConditions, but it still left other restore-time reset calls in place, which is why the failure shifted to Network.setUserAgentOverride.

updated restore logic to skip all unspecified reset calls when there’s nothing to restore, instead of only skipping the network reset. That prevents the timeout from just moving to another CDP command. also added a regression test to verify restore does not call setUserAgent in that no-emulation case.

zyzyzyryxy:
@dhananjay6561 
> updated restore logic to skip all unspecified reset calls when there’s nothing to restore [...]

Sorry, I don't see this update, only merges from main. Please double check you did push correctly.

wolfib:
#1797 has been fixed

PR #1899, opened 2026-04-18
fix(cli): correct WebP MIME type check in handleResponse ('webp' → 'image/webp')

### LINKED ISSUE #1898: Bug: WebP image MIME type check uses 'webp' instead of 'image/webp' in handleResponse
## Bug Report

**File:** `src/daemon/client.ts`

### Description

In the `handleResponse` function, the MIME type case for WebP images is incorrectly written as `'webp'` instead of `'image/webp'`. This means WebP images served with the correct MIME type (`image/webp`) will **never** match the case, fall through to the default `.png` extension, and be saved with the wrong extension.

### Affected Code (lines ~145-149)

```ts
switch (mimeType) {
  case 'image/jpg':
  case 'image/jpeg':
    extension = '.jpeg';
    break;
  case 'webp':          // ❌ Should be 'image/webp'
    extension = '.webp';
    break;
}
```

### Expected Behavior

WebP images should be saved with the `.webp` extension.

### Actual Behavior

WebP images are saved with the default `.png` extension because `'webp'` never matches the `mimeType` string `'image/webp'`.

### Fix

Change `case 'webp':` to `case 'image/webp':`

Signed-off-by: cocoon <54054995+kuishou68@users.noreply.github.com>

kuishou68:
## Summary

Fixes a bug in `src/daemon/client.ts` where the MIME type case for WebP images was `'webp'` instead of the correct `'image/webp'`.

## Problem

In `handleResponse`, the switch statement for determining file extension used the incorrect string `'webp'` as the MIME type. Since browsers and servers send `image/webp` as the MIME type, the case never matched, and all WebP images were saved with a `.png` extension instead of `.webp`.

```ts
// Before (buggy)
case 'webp':
  extension = '.webp';
  break;

// After (fixed)
case 'image/webp':
  extension = '.webp';
  break;
```

## Testing

WebP screenshots/images returned by the tool will now correctly receive the `.webp` extension.

Closes #1898

Signed-off-by: cocoon <54054995+kuishou68@users.noreply.github.com>

### DIFF
--- a/src/daemon/client.ts
+++ b/src/daemon/client.ts
@@ -172,7 +172,7 @@ export async function handleResponse(
         case 'image/jpeg':
           extension = '.jpeg';
           break;
-        case 'webp':
+        case 'image/webp':
           extension = '.webp';
           break;
       }


OrKoN (approved):
Thanks!

PR #1901, opened 2026-04-18
test: save WebP responses with the right extension

### LINKED ISSUE #1898: Bug: WebP image MIME type check uses 'webp' instead of 'image/webp' in handleResponse

MukundaKatta:
## Summary
- fix the WebP MIME type match in daemon response handling
- save  responses with a  suffix instead of falling back to 
- add a regression test for WebP image handling

## Testing
- npm run build
- node --test --test-name-pattern="parsing" build/tests/daemon/client.test.js

Fixes #1898

### DIFF
--- a/tests/daemon/client.test.ts
+++ b/tests/daemon/client.test.ts
@@ -107,5 +107,20 @@ describe('daemon client', () => {
       const response = await handleResponse(unsupportedContentResponse, 'md');
       assert.ok(response.includes('.png'));
     });
+
+    it('uses the webp extension for WebP images', async () => {
+      const webpContentResponse = {
+        content: [
+          {
+            type: 'image' as const,
+            data: 'base64data',
+            mimeType: 'image/webp',
+          },
+        ],
+        structuredContent: {},
+      };
+      const response = await handleResponse(webpContentResponse, 'md');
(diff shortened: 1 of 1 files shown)


OrKoN (approved):
thanks!

PR #1903, opened 2026-04-18
fix: avoid duplicate plugin MCP registration

### LINKED ISSUE #1854: Claude Code /doctor reports duplicate MCP server when plugin is loaded via marketplace
## Summary

When `chrome-devtools-mcp` is installed as a Claude Code plugin via marketplace, `/doctor` reports:

```
Plugin Errors
└ 1 plugin error(s) detected:
  └ plugin:chrome-devtools-mcp:chrome-devtools [chrome-devtools-mcp]: MCP server "chrome-devtools" skipped — same command/URL as already-configured "chrome-devtools"
```

## Root Cause

The marketplace structure has `source: "./"` in `marketplace.json`, which means the marketplace root directory **is** the plugin directory. This causes the same `plugin.json` (containing `mcpServers.chrome-devtools`) to be loaded twice:

1. First load from `~/.claude/plugins/marketplaces/chrome-devtools-plugins/.claude-plugin/plugin.json`
2. Second load from the cached copy at `~/.claude/plugins/cache/chrome-devtools-plugins/chrome-devtools-mcp/latest/.claude-plugin/plugin.json`

Both contain the identical `mcpServers` block:

```json
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": ["chrome-devtools-mcp@latest"]
    }
  }
}
```

The second registration is skipped as a duplicate, producing the warning.

## Impact

- Functional: **None** — the MCP server loads correctly on the first registration
- UX: Persistent error in `/doctor` output that cannot be resolved by the user

## Possible Fix

- Remove `mcpServers` from `plugin.json` and keep it only in `.mcp.json` (or vice versa), ensuring only one source of truth
- Or adjust the caching/loading logic so `source: "./"` doesn't result in double-loading

## Environment

- Claude Code v2.1.104
- macOS (Darwin 25.3.0)
- Plugin version: latest

MukundaKatta:
Summary:
- remove the duplicate chrome-devtools MCP server declaration from .claude-plugin/plugin.json
- keep .mcp.json as the single source of truth for plugin MCP registration
- avoid Claude Code /doctor warnings when the marketplace plugin is installed

Testing:
- node -e "JSON.parse(require('fs').readFileSync('.claude-plugin/plugin.json','utf8'))"

Fixes #1854

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,11 +1,5 @@
 {
   "name": "chrome-devtools-mcp",
   "version": "0.21.0",
-  "description": "Reliable automation, in-depth debugging, and performance analysis in Chrome using Chrome DevTools and Puppeteer",
-  "mcpServers": {
-    "chrome-devtools": {
-      "command": "npx",
-      "args": ["chrome-devtools-mcp@latest"]
-    }
-  }
+  "description": "Reliable automation, in-depth debugging, and performance analysis in Chrome using Chrome DevTools and Puppeteer"
 }


OrKoN:
We should delete .mcp.json instead and keep the configuration within individual plugins to support things like https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1851

Issue #1184, opened 2026-03-15
CLI  or chrome-devtools-cli support --autoConnect ?


chaunsin:
### Is your feature request related to a problem? Please describe.

Hello, I would like to know if the chrome-devtools-cli skills support --autoConnect mode is available, similar to the mcp?

### Describe the solution you'd like

no

### Describe alternatives you've considered

no

wolfib:
Yes, as far as I understand it should be possible to use the --autoConnect option with the experimental CLI as well. Please report back with repro instructions if it doesn't work for you.

chaunsin:
> Yes, as far as I understand it should be possible to use the --autoConnect option with the experimental CLI as well. Please report back with repro instructions if it doesn't work for you.

I didn't find the content of the --autoConnect parameter. My CLI version is `0.20.0`.

```
chaunsin@MacBook-Pro % chrome-devtools start --help
chrome-devtools start

Start or restart chrome-devtools-mcp

Options:
      --version                 Show version number                                                            [boolean]
      --help                    Show help                                                                      [boolean]
  -u, --browserUrl              Connect to a running, debuggable Chrome instance (e.g. `http://127.0.0.1:9222`). For
                                more details see: https://github.com/ChromeDevTools/chrome-devtools-mcp#connecting-to-a-
                                running-chrome-instance.                                                        [string]
  -w, --wsEndpoint              WebSocket endpoint to connect to a running Chrome instance (e.g.,
                                ws://127.0.0.1:9222/devtools/browser/<id>). Alternative to --browserUrl.        [string]
      --wsHeaders               Custom headers for WebSocket connection in JSON format (e.g., '{"Authorization":"Bearer
                                token"}'). Only works with --wsEndpoint.                                        [string]
      --headless                Whether to run in headless (no UI) mode.                       [boolean] [default: true]
  -e, --executablePath          Path to custom Chrome executable.                                               [string]
      --isolated                If specified, creates a temporary user-data-dir that is automatically cleaned up after
                                the browser is closed. Defaults to false.                                      [boolean]
      --userDataDir             Path to the user data directory for Chrome. Default is
                                $HOME/.cache/chrome-devtools-mcp/chrome-profile$CHANNEL_SUFFIX_IF_NON_STABLE    [string]
      --channel                 Specify a different Chrome channel that should be used. The default is the stable
                                channel version.                   [string] [choices: "stable", "canary", "beta", "dev"]
[...]

PR #1859, opened 2026-04-14
fix(cli): allow --autoConnect in chrome-devtools start

zerone0x:
## Summary
- allow chrome-devtools start to accept --autoConnect and forward it to the daemon
- document the CLI example for auto-connect

## Testing
- not run (not requested)

Fixes #1184

PR #1905, opened 2026-04-18
docs: clarify CLI autoConnect support

MukundaKatta:
Summary:
- document that chrome-devtools start intentionally does not support --autoConnect
- point users to --browserUrl as the supported way to attach the CLI daemon to an existing Chrome instance

Testing:
- reviewed docs diff locally

Relates to #1184


Lightning00Blade:
Thanks for the PR but we are going to enable support once https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1859 lands, so this PR will be obsolete.

PR #1909, opened 2026-04-20
chore(memory): expose a tool for getting the inital data.

Lightning00Blade:
This refactors the code to extract the Id logic from the PageCollector and provide it in the heapsnapshot.
We need to use an UID as we need the internal ClassKey to query the heapsnapshot, but that is a strange string (usually looking like `,ClassName`) which may get the LLM confused as we use comma separated output.

### DIFF
--- a/src/HeapSnapshotManager.ts
+++ b/src/HeapSnapshotManager.ts
@@ -23,6 +23,7 @@ export class HeapSnapshotManager {
     {
       snapshot: DevTools.HeapSnapshotModel.HeapSnapshotProxy.HeapSnapshotProxy;
       worker: DevTools.HeapSnapshotModel.HeapSnapshotProxy.HeapSnapshotWorkerProxy;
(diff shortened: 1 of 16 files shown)


src/HeapSnapshotManager.ts
```diff
@@ -28,20 +40,35 @@ export class HeapSnapshotManager {
     }
 
     const {snapshot, worker} = await this.#loadSnapshot(absolutePath);
-    this.#snapshots.set(absolutePath, {snapshot, worker});
+    this.#snapshots.set(absolutePath, {
+      snapshot,
+      worker,
+      uidToClassKey: new Map<number, string>(),
+      classKeyToUid: new Map<string, number>(),
```
Lightning00Blade, line 47: In the next PR where we query the Nodes we need the map back from uid to classKey.

OrKoN (approved):

PR #1911, opened 2026-04-20
refactor: use puppeteer Extension API

nroscino:
This PR refactor the extension usages using the new Puppeteer API.

Requires puppeteer release before merging


OrKoN:
should be a `refactor: ` unless you also want to enable the feature and update the docs in this PR?

nroscino:
> should be a `refactor: ` unless you also want to enable the feature and update the docs in this PR?

Makes sense. I changed it to refactor. I think it's better to enable it after the logs implementation

OrKoN (approved):

PR #1915, opened 2026-04-20
feat: implement extension service worker logs

nroscino:
This PR introduces the possiblity to collect console messages from service workers.

### DIFF
--- a/src/ServiceWorkerCollector.ts
+++ b/src/ServiceWorkerCollector.ts
@@ -1,0 +1,227 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {UncaughtError} from './PageCollector.js';
+import type {
+  ConsoleMessage,
+  WebWorker,
+  Target,
+  CDPSession,
+  Protocol,
+  Browser,
+} from './third_party/index.js';
+import type {ExtensionServiceWorker} from './types.js';
+import type {WithSymbolId} from './utils/id.js';
+import {createIdGenerator, stableIdSymbol} from './utils/id.js';
+
+const CHROME_EXTENSION_PREFIX = 'chrome-extension://';
+
+export class ServiceWorkerSubscriber {
+  #target: Target;
+  #callback: (item: ConsoleMessage | UncaughtError) => void;
+  #session?: CDPSession;
+  #worker?: WebWorker;
+
+  constructor(
+    target: Target,
+    callback: (item: ConsoleMessage | UncaughtError) => void,
+  ) {
+    this.#target = target;
+    this.#callback = callback;
+  }
+
+  async subscribe() {
+    this.#session = await this.#target.createCDPSession();
+    await this.#session.send('Runtime.enable');
+    this.#session.on('Runtime.exceptionThrown', this.#onExceptionThrown);
+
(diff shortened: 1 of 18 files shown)


src/ServiceWorkerCollector.ts
```diff
@@ -0,0 +1,202 @@
...
+    if (this.#worker) {
+      this.#worker.on('console', this.#onConsole);
+    }
+  }
+
+  async unsubscribe() {
+    if (this.#worker) {
+      this.#worker.off('console', this.#onConsole);
+    }
+    if (this.#session) {
```
Lightning00Blade: I think we need to close the session in all cases.

tests/tools/console.test.ts
```diff
@@ -13,22 +14,140 @@ import type {ParsedArguments} from '../../src/bin/chrome-devtools-mcp-cli-option
...
+        if (!sw) {
+          assert.fail('Service worker not found in context list');
+        }
+
+        const response2 = new McpResponse({} as ParsedArguments);
+
+        await context.triggerExtensionAction(extensionId);
+        const worker = await swTarget.worker();
+
+        // Puppeteer's waitForFunction injects a utility script that requires MutationObserver.
```
Lightning00Blade: Do you think we should fix that upstream in Puppeteer?

tests/tools/console.test.ts
```diff
@@ -13,22 +14,140 @@ import type {ParsedArguments} from '../../src/bin/chrome-devtools-mcp-cli-option
...
+        assert.ok(extensionId, 'Extension ID should be returned');
+
+        const swTarget = await context.browser.waitForTarget(
+          t => t.type() === 'service_worker' && t.url().includes(extensionId),
+        );
+
+        const swList = await context.createExtensionServiceWorkersSnapshot();
+        const sw = swList.find(s => s.target === swTarget);
+        if (!sw) {
+          assert.fail('Service worker not found in context list');
```
Lightning00Blade: Nit: Use `assert( X, `

tests/tools/console.test.ts
```diff
@@ -13,22 +14,140 @@ import type {ParsedArguments} from '../../src/bin/chrome-devtools-mcp-cli-option
...
+            () => {
+              clearTimeout(timeout);
+              resolve(undefined);
+            },
+          );
+        });
+
+        await worker?.evaluate(
+          `
+            console.log('Service Worker starting...');
```
Lightning00Blade: I am curious we we should move this in the extension code just to verify that we are not missing things, which I suspect we might.

Lightning00Blade (approved):
Overall looks good, I will try to play around to check if we can extract the duplciation from the other Collector logic.

tests/tools/console.test.ts
nroscino: Yes, I was also thinking that we should probably have `worker.on("error",...)` instead of relying on `Runtime.exceptionThrown`. 
I can do both in puppeteer and open a follow up to remove these here, WDYT?

Lightning00Blade: That sounds good. I think expose it like `console` as the Page impl of `page.on("error"` does not provide full information, we should fix that as well and use the same mechism for both removing some other code in the repo as well.

Lightning00Blade (approved):
LGTM

PR #1922, opened 2026-04-21
feat: support Chrome extensions debugging

### LINKED ISSUE #96: Feature Request: Support for browser extension development and debugging
## Goals

The primary goals of these new tools is to enable Chrome Extension debugging for AI Assistants:

*   Manage the extension lifecycle (install, reload, …)
*   Ability to inspect and debug extension components, such as service workers, side panel, and popups.
*   Simulate extension user interactions (e.g. click the extension action, …)

## Example prompts

*   <code>Install my extension and open https://example.com. Find out why the reading time is not injected into the page below the first heading.</code>
*   <code>Why does my extension not show a summary of https://example.com in the sidepanel when clicking the extension action?</code>
*   <code>When selecting the first heading on https://example.com, opening my extension popup and clicking "translate text" the translation is not displayed in the popup. Why?</code>

## Use cases

1. Load an extension while in development
2. Debug extensions via console logs for extension pages (side panel, tabs, popup, offscreen document) and extension context (content scripts, service worker)
3. Inspect extension state by evaluating scripts in extension pages and contexts
4. Test an extension by triggering an extension action 
5. Test an extension by simulating user input on extension pages

## Features

The following tools should be added to the Chrome DevTools MCP server:

### Extension Management

*   `list_extensions`: Lists all installed extensions, including their name, ID, version, and enabled status.
*   `install_extension`: Installs an extension.
    *   Existing CDP method: Extensions.loadUnpacked
*   `uninstall_extension`: Uninstalls an enabled extension.
    *   Existing CDP method: Extensions.uninstall
*   `reload_extension`: Reloads an extension to apply any changes.

### Extension Debugging

*   Update `list_pages` to return a list of [active extension contexts](https://developer.chrome.com/docs/extensions/reference/api/runtime#method-getContexts) (sidepanel, popup, extension page, service worker, offscreen document). See also [Support extensions execution contexts · Issue #1215 · puppeteer/puppeteer](https://github.com/puppeteer/puppeteer/issues/1215) 
    *   Existing CDP event: Runtime.executionContextCreated
*   Update `select_page` to also support extensions contexts
*   Update `evaluate_script` to also support script execution in an extension context
*   Update `list_console_messages` to also support console logs from an extension context

### Extension Interaction

*   `click_extension_icon`: Simulates a click on an extension's browser action or page action icon.
    *   Note: Once the extension's popup is open, CDP already exposes the popup as a target, so we can inspect it just like a page.
*   `click_extension_context_menu`: Simulates a click on an extension’s context menu entry.
*   All [input automation tools](https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/docs/tool-reference.md#input-automation) need to be able to work in an extension context (sidepanel, popup, extension tab page)

## Constraints

*   Token usage should not increase when extension debugging is not enabled.

### LINKED ISSUE #1173: Feature request: support extension page and service-worker visibility when attaching to existing Chrome
## Context

This issue was originally filed as a bug report because the attached-browser path does not pass `categoryExtensions` through the same way as the launched-browser path. After maintainer feedback, I understand that the current behavior is intentional: full extension debugging is currently supported only over piped connection, and enabling the full extension category over WebSocket-based attach flows would expose tools that are known to fail.

I am keeping that original context here so the existing thread and linked PR discussion still make sense.

## Feature request

Support a narrower extension-debugging capability when attaching to an already-running Chrome instance via:

- `--browserUrl`
- `--wsEndpoint`
- `--autoConnect`

Specifically, the request is to expose extension pages and extension service workers for observability in attach mode, without implying that the full extension-management toolset is supported there.

## Requested scope

When connected to an existing Chrome instance, allow `list_pages` to include:

- extension pages
- extension service workers

This would support workflows where the MCP server attaches to a real Chrome profile that already has authenticated state, tab layout, and active extension contexts, and the user needs visibility into those extension targets.

## Non-goals

This request is not asking to enable the full extension-management toolset over WebSocket-based attach flows. In particular, tools such as these may remain unavailable or explicitly unsupported in connected-browser mode:

- `install_extension`
- `uninstall_extension`
- `reload_extension`
- `list_extensions`
- `trigger_extension_action`

## Why this narrower feature would still help

Even without full extension-management support, page/service-worker visibility in attach mode would still be valuable for:

- attaching to a real browser profile with existing auth/session state
- inspecting extension contexts in that same running browser
- debugging workflows that depend on extension pages or extension service workers being visible to the MCP server

## Existing prototype

There is a prototype PR here: https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1171

That PR currently assumes the attached-browser behavior should match launch-mode behavior more directly. If the narrower feature direction described above is acceptable, the implementation can be reworked around that supported subset instead.

nroscino:
Adds ability to install, uninstall and reload extensions, trigger extension actions and inspect extension pages, service workers and content scripts. Specify `--categoryExtensions` to enable.

Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1173
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/96

### DIFF
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -36,6 +36,12 @@
   - [`list_console_messages`](#list_console_messages)
   - [`take_screenshot`](#take_screenshot)
   - [`take_snapshot`](#take_snapshot)
+- **[Extensions](#extensions)** (5 tools)
+  - [`install_extension`](#install_extension)
+  - [`list_extensions`](#list_extensions)
+  - [`reload_extension`](#reload_extension)
+  - [`trigger_extension_action`](#trigger_extension_action)
+  - [`uninstall_extension`](#uninstall_extension)
 - **[Memory](#memory)** (1 tools)
   - [`take_memory_snapshot`](#take_memory_snapshot)
 
@@ -387,6 +393,56 @@ in the DevTools Elements panel (if any).
 
 - **filePath** (string) _(optional)_: The absolute path, or a path relative to the current working directory, to save the snapshot to instead of attaching it to the response.
 - **verbose** (boolean) _(optional)_: Whether to include all possible information available in the full a11y tree. Default is false.
+
+---
+
+## Extensions
+
+### `install_extension`
+
+**Description:** Installs a Chrome extension from the given path.
+
+**Parameters:**
+
+- **path** (string) **(required)**: Absolute path to the unpacked extension folder.
+
+---
+
+### `list_extensions`
+
+**Description:** Lists all the Chrome extensions installed in the browser. This includes their name, ID, version, and enabled status.
+
+**Parameters:** None
+
+---
+
+### `reload_extension`
+
+**Description:** Reloads an unpacked Chrome extension by its ID.
+
+**Parameters:**
+
+- **id** (string) **(required)**: ID of the extension to reload.
+
+---
+
+### `trigger_extension_action`
+
+**Description:** Triggers the default action of an extension by its ID.
+
+**Parameters:**
+
+- **id** (string) **(required)**: ID of the extension to trigger the action for.
+
+---
+
+### `uninstall_extension`
+
+**Description:** Uninstalls a Chrome extension by its ID.
+
+**Parameters:**
+
+- **id** (string) **(required)**: ID of the extension to uninstall.
 
 ---
 
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -398,6 +398,8 @@ in the DevTools Elements panel (if any).
 
 ## Extensions
 
+> NOTE: Extensions are not active by default. Use the '--category-extensions' flag
+
 ### `install_extension`
 
 **Description:** Installs a Chrome extension from the given path.
--- a/skills/troubleshooting/SKILL.md
+++ b/skills/troubleshooting/SKILL.md
@@ -47,6 +47,13 @@ If the server starts successfully but only a limited subset of tools (like `list
 
 All tools in `chrome-devtools-mcp` are annotated with `readOnlyHint: true` (for safe, non-modifying tools) or `readOnlyHint: false` (for tools that modify browser state, like `emulate`, `click`, `navigate_page`). To access the full suite of tools, the user must disable read-only mode in their MCP client (e.g., by exiting "Plan Mode" in Gemini CLI or adjusting their client's tool safety settings).
 
+#### Symptom: Extension tools are missing or extensions fail to load
+
+If the tools related to extensions (like `install_extension`) are not available, or if the extensions you load are not functioning:
+
+1. **Check for the `--categoryExtensions` flag**: Ensure this flag is passed in the MCP server configuration to enable the extension category tools.
+2. **Make sure the browser was not started with `--autoConnect`**: When using `--autoConnect`, Chrome is not able to load extensions.
+
 #### Other Common Errors
 
 Identify other error messages from the failed tool call or the MCP initialization logs:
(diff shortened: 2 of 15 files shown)


src/bin/chrome-devtools-mcp-cli-options.ts
```diff
@@ -222,10 +222,10 @@ export const cliOptions = {
   },
   categoryExtensions: {
     type: 'boolean',
-    hidden: true,
+    hidden: false,
```
OrKoN, line 225:
```suggestion
    default: false,
```

docs/tool-reference.md
```diff
@@ -36,6 +36,12 @@
   - [`list_console_messages`](#list_console_messages)
   - [`take_screenshot`](#take_screenshot)
   - [`take_snapshot`](#take_snapshot)
+- **[Extensions](#extensions)** (5 tools)
```
OrKoN, line 39: I think we did not have a category that is disabled by default before, I think generally non-default tools did not show up in the reference before.

OrKoN: I think we need to add a note here that these are only available if `--category-extensions` is enabled

skills/troubleshooting/SKILL.md
```diff
@@ -47,6 +47,13 @@ If the server starts successfully but only a limited subset of tools (like `list
 
 All tools in `chrome-devtools-mcp` are annotated with `readOnlyHint: true` (for safe, non-modifying tools) or `readOnlyHint: false` (for tools that modify browser state, like `emulate`, `click`, `navigate_page`). To access the full suite of tools, the user must disable read-only mode in their MCP client (e.g., by exiting "Plan Mode" in Gemini CLI or adjusting their client's tool safety settings).
 
+#### Symptom: Extension tools are missing or extensions fail to load
+
+If the tools related to extensions (like `install_extension`) are not available, or if the extensions you load are not functioning:
+
+1. **Check for the `--categoryExtensions` flag**: Ensure this flag is passed in the MCP server configuration to enable the extension category tools.
+2. **Make sure the browser was not started with `--autoConnect`**: When using `--autoConnect`, Chrome is not able to load extensions.
```
mathiasbynens: IIUC, `--autoConnect` is a flag of the MCP server, not the browser, right? In that case, how about something like this:


```suggestion
2. **Make sure the MCP server configuration does not include `--autoConnect`**: when using `--autoConnect`, Chrome is not able to load extensions.
```

src/bin/chrome-devtools-mcp-cli-options.ts
```diff
@@ -222,10 +222,10 @@ export const cliOptions = {
...
     conflicts: ['browserUrl', 'autoConnect', 'wsEndpoint'],
```
OrKoN: we should remove conflicts here to allow using it with newer versions

skills/troubleshooting/SKILL.md
nroscino: Looks good!

skills/troubleshooting/SKILL.md
```diff
@@ -47,6 +47,13 @@ If the server starts successfully but only a limited subset of tools (like `list
 
 All tools in `chrome-devtools-mcp` are annotated with `readOnlyHint: true` (for safe, non-modifying tools) or `readOnlyHint: false` (for tools that modify browser state, like `emulate`, `click`, `navigate_page`). To access the full suite of tools, the user must disable read-only mode in their MCP client (e.g., by exiting "Plan Mode" in Gemini CLI or adjusting their client's tool safety settings).
 
+#### Symptom: Extension tools are missing or extensions fail to load
+
+If the tools related to extensions (like `install_extension`) are not available, or if the extensions you load are not functioning:
+
+1. **Check for the `--categoryExtensions` flag**: Ensure this flag is passed in the MCP server configuration to enable the extension category tools.
+2. **Make sure the MCP server configuration does not include `--autoConnect`**: when using `--autoConnect`, Chrome is not able to load extensions.
```
OrKoN:
```suggestion
2. **Make sure the MCP server is configured to launch Chrome instead of connecting to an existing instance. Chrome before 149 is not able to load extensions when connecting to existing instance (`--auto-connect`, `--browserUrl`).
```

OrKoN (approved):

mathiasbynens (approved):

PR #1926, opened 2026-04-21
test: refactor tests to reduce duplication

nroscino:


tests/index.test.ts
```diff
@@ -245,3 +160,56 @@ describe('e2e', () => {
...
+        if (tool.annotations?.conditions) {
+          continue;
+        }
+        if (
+          tool.annotations?.category &&
+          filterOutCategories.includes(tool.annotations?.category)
+        ) {
+          continue;
+        }
+        definedNames.push(tool.name);
```
OrKoN: can we deduplicate here too?

tests/index.test.ts
```diff
@@ -245,3 +161,55 @@ describe('e2e', () => {
...
+    }
+  }
+  return definedNames;
+}
+
+function toolShouldBeSkipped(
+  tool: ToolDefinition,
+  filteredOutCategories: ToolCategory[],
+) {
+  if (tool === null || typeof tool !== 'object' || !('name' in tool)) {
```
OrKoN: do we need these checks?

nroscino: right now in the tools folder we are not exporting only tools (for example the "category.ts"  file). If we remove this check completely then tests will fail when `const fileTools = await import(`../src/tools/${file}`);` didn't actually get a tool definition file.

OrKoN: I see, I guess the type is wrong then.

OrKoN (approved):

tests/index.test.ts
nroscino: Yeah, it kinda is. I can put `unknown` and assume the type is correct only after that first if, is it ok for you?

OrKoN: if we can filter out based on this before we call toolShouldBeSkipped to ensure the right type in the toolShouldBeSkipped function, I think it would be nice but not required.

nroscino: Done!

PR #1929, opened 2026-04-21
docs(cli): clarify how to attach to an existing Chrome instance

MukundaKatta:
## Summary
- clarify that `chrome-devtools start` does not accept `--autoConnect`
- point users to `--browserUrl` when they want to attach to an already running Chrome instance

## Why
The CLI docs already describe supported daemon startup arguments, but it is easy to assume that `--autoConnect` is available here as well. Calling out the intended `--browserUrl` workflow makes the attachment path clearer for CLI users.


MukundaKatta:
Withdrawing — low-value drive-by docs change. Apologies for the noise.

PR #1930, opened 2026-04-21
fix(daemon): treat WebP screenshots as image/webp

MukundaKatta:
## Summary
- return the correct WebP MIME type from the daemon client
- add regression coverage for the client screenshot handling path

## Testing
- unable to run the targeted Jest slice in this workspace because this clone does not currently have a local Jest binary installed
- added focused coverage in tests/daemon/client.test.ts


OrKoN:
Thanks! I think this test has already landed in previous PRs.

MukundaKatta:
Thanks @OrKoN — appreciate the heads-up.

PR #1931, opened 2026-04-22
feat: agentic browsing in lighthouse

lusayaa:
pending https://github.com/GoogleChrome/lighthouse/pull/16980


src/third_party/LIGHTHOUSE_MCP_BUNDLE_THIRD_PARTY_NOTICES
```diff
@@ -1529,29 +1529,6 @@ URL: zeit/ms
 Version: 2.1.2
 License: MIT
 
-The MIT License (MIT)
```
OrKoN, line 1532: strange that the license is removed but the dependency is still there? perhaps zeit/ms moved the license file?

OrKoN: `npm run bundle && npm run test:no-build -- --test-update-snapshots` would allow rebasing this snapshot.

lusayaa: looks like they renamed it to `license.md`

src/tools/lighthouse.ts
```diff
@@ -42,19 +43,31 @@ export const lighthouseAudit = definePageTool({
       .string()
       .optional()
       .describe('Directory for reports. If omitted, uses temporary files.'),
+    categories: zod
```
OrKoN: let's remove categories

matthiasrohmer:
A drive-by nit: this seems to fail when chrome://flags/#enable-webmcp-testing is not `Enabled`. Could that be? With Antigravity I get `Protocol error (WebMCP.enable)` if use a out of the box Chrome instance. Might be good to handle this error case, as I assume most people don't have it enabled by default and I am not too sure when it will be.

OrKoN:
Interesting, the test tests/tools/lighthouse.test.ts seems to pass but I guess Lighthouse might be sending WebMCP.enable conditionally.

lusayaa:
Ah I suppose that was kind of on purpose to surface an error with the backend. Does the whole lighthouse run fail for you or just the specific WebMCP audits? @matthiasrohmer

matthiasrohmer:
Yep the whole run failed, and the agent just stopped. :)

OrKoN:
I am not able to reproduce the error but I do not get a new category in the results (Chrome Canary, launched via `build/src/bin/chrome-devtools-mcp.js --channel=canary --chrome-arg='--enable-features=WebMCPTesting,DevToolsWebMCPSupport'`):

Screenshot 2026-04-27 at 14 16 52

OrKoN:
> I am not able to reproduce the error but I do not get a new category in the results (Chrome Canary, launched via `build/src/bin/chrome-devtools-mcp.js --channel=canary --chrome-arg='--enable-features=WebMCPTesting,DevToolsWebMCPSupport'`):
> 
> Screenshot 2026-04-27 at 14 16 52

I had a stale build somehow.

OrKoN:
Discussed offline: this PR waits for a follow-up Lighthouse update

OrKoN (approved):

PR #1934, opened 2026-04-22
feat: support webm format in screencast

OrKoN:
Modifies the experimental screencast command to support `.webm` in addition to `.mp4`. Renamed `path` to `filePath` to align with other tools.

### DIFF
--- a/src/tools/screencast.ts
+++ b/src/tools/screencast.ts
@@ -9,7 +9,7 @@ import os from 'node:os';
 import path from 'node:path';
 
 import {zod} from '../third_party/index.js';
-import type {ScreenRecorder} from '../third_party/index.js';
+import type {ScreenRecorder, VideoFormat} from '../third_party/index.js';
 import {ensureExtension} from '../utils/files.js';
 
 import {ToolCategory} from './categories.js';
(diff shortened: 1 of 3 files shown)


src/tools/screencast.ts
```diff
@@ -46,16 +47,30 @@ export const startScreencast = definePageTool({
       return;
     }
 
-    const filePath = request.params.path ?? (await generateTempFilePath());
-    const resolvedPath = ensureExtension(path.resolve(filePath), '.mp4');
+    const filePath = request.params.filePath ?? (await generateTempFilePath());
+    let enforcedExtension = '.mp4' as `.${string}`;
```
wolfib, line 51: As long as `enforcedExtension` has a hardcoded value, the following loop does not really achieve anything.

OrKoN: Yeah it is a typo on line 55, should be `filePath.endsWith`

src/tools/screencast.ts
```diff
@@ -46,16 +47,30 @@ export const startScreencast = definePageTool({
...
+        enforcedExtension = supportedExtension;
+        format = supportedExtension.substring(1) as VideoFormat;
+        break;
+      }
+    }
+
+    const resolvedPath = ensureExtension(
+      path.resolve(filePath),
+      enforcedExtension,
+    ) as `${string}.webm`;
```
wolfib, line 65: webp or mp4 instead of webm?

OrKoN: should be webm here, the typo is in the supported formats.

wolfib: But 'mp4' is allowed too, isn't it? 
```
as `${string}.${'mp4'|'webm'}`
```

OrKoN: yes but it does not matter and I do not want to repeat all possible types here

OrKoN: the problem here is that Puppeteer requires overly specific literal types.

wolfib: Got it, thanks!

wolfib (approved):
LGTM

PR #1025, opened 2026-02-23
Group identical console messages in list_console_messages

vamgan:
Fixes #904\n\n- Collapse consecutive identical console messages in list_console_messages output and annotate with "[N times]".\n- Preserve the first msgid and expose grouped msgids/count in structuredContent.\n- Add a unit test for grouping behavior.

PR #963, opened 2026-02-14
feat(console): group consecutive identical console messagesfeat(console): group consecutive identical console messages

SucRunBug:
## Summary

- Group consecutive identical messages in `list_console_messages`.
- Add `[N times]` suffix in text output for collapsed duplicates.
- Add `occurrences` to structured console message output when duplicates are collapsed.
- Add test coverage for grouped duplicate console messages.

Closes #904.

## Changes

- `/Users/yanmai/Downloads/MCP 测试/chrome-devtools-mcp/src/McpResponse.ts`
  - Added grouping/formatting logic for consecutive identical console messages.
- `/Users/yanmai/Downloads/MCP 测试/chrome-devtools-mcp/tests/tools/console.test.ts`
  - Added `groups consecutive identical messages` test.

## Validation

- `npm run typecheck`
- `npx eslint src/McpResponse.ts tests/tools/console.test.ts`
- `PUPPETEER_EXECUTABLE_PATH='/Applications/Google Chrome.app/Contents/MacOS/Google Chrome' npm run test:no-build -- --test-name-pattern='groups consecutive identical messages' tests/tools/console.test.ts`
- `PUPPETEER_EXECUTABLE_PATH='/Applications/Google Chrome.app/Contents/MacOS/Google Chrome' npm run test:no-build -- --test-name-pattern='adds console messages when the setting is true' tests/McpResponse.test.ts`

PR #1939, opened 2026-04-22
feat: group identical consecutive console messages in list_console_messages

### LINKED ISSUE #904: Identical console messages should be grouped together in the list_console_messages tool

It is easy to log a message from a hot loop. This results in tons of identical messages.

Similar to Chrome DevTools, subsequent identical messages should be collapsed and only shown once. A new `[20 times]` prefix/suffix can be added to signify how many of these identical console messages there are.

Being careful about logging xD

### Additional context

_No response_

masamaru0513:
## Summary

Group consecutive identical console messages in `list_console_messages`, similar to Chrome DevTools' console grouping behavior.

Fixes #904

## Changes

- Introduce `GroupedConsoleFormatter` subclass that extends `ConsoleFormatter` and overrides `toString()` / `toJSON()` for count-aware formatting
- Add `ConsoleFormatter.groupConsecutive()` static method that groups consecutive messages with the same type, text, and argument count
- Apply grouping **before pagination** so grouped counts are accurate and page sizes reflect the collapsed view
- Add unit tests for grouping logic, string formatting, and JSON output

## Key design decisions

- **`GroupedConsoleFormatter` subclass**: Keeps the existing formatter interface clean — no new methods added to `ConsoleFormatter`. `ConsoleFormatter` and `GroupedConsoleFormatter` are interchangeable via the same interface.
- **Grouping before pagination** (not at format time): This was the feedback on #963 and #1025 — grouping at format time breaks pagination counts. This implementation groups in `McpResponse` before calling `paginate()`.
- **No `lastId`**: Since grouped messages are truly identical, only the first message's ID is needed.
- **`argCount` matching**: Prevents false grouping of messages with the same text but different argument counts.

## Output example

```
msgid=1 [log] hello world (1 args) [5 times]
```

## Testing

- Unit tests in `tests/formatters/ConsoleFormatterGrouping.test.ts`
- Manual verification: identical messages (×5), mixed pattern (A,A,B,A,A → A×2, B×1, A×2)

### DIFF
--- a/src/formatters/ConsoleFormatter.ts
+++ b/src/formatters/ConsoleFormatter.ts
@@ -13,6 +13,8 @@ import {UncaughtError} from '../PageCollector.js';
 import * as DevTools from '../third_party/index.js';
 import type {ConsoleMessage} from '../third_party/index.js';
 
+import type {IssueFormatter} from './IssueFormatter.js';
+
 export interface ConsoleFormatterOptions {
   fetchDetailedData?: boolean;
   id: number;
@@ -32,6 +34,7 @@ interface ConsoleMessageConcise {
   text: string;
   argsCount: number;
   id: number;
+  count?: number;
 }
 
 interface ConsoleMessageDetailed extends ConsoleMessageConcise {
@@ -54,7 +57,7 @@ export class ConsoleFormatter {
 
   readonly isIgnored: IgnoreCheck;
 
-  private constructor(params: {
+  protected constructor(params: {
     id: number;
     type: string;
     text: string;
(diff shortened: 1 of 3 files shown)


src/formatters/ConsoleFormatter.ts
```diff
@@ -175,6 +179,18 @@ export class ConsoleFormatter {
     return convertConsoleMessageConciseToString(this.toJSON());
   }
 
+  // The short format with a repeat count.
+  toStringGrouped(count: number, lastId?: number): string {
```
OrKoN: we should not define new methods on the formatter interfaces. WDYT instead to create a GroupedConsoleFormatter subclass that holds a group of messages. If the messages are truly identical we do not need to know the lastId and we can just use the ID of the first message. We can then use ConsoleFormatter or GroupedConsoleFormatter interchangably because they have the same interface.

masamaru0513: Thanks for the quick review! I agree with the approach — I'll refactor to use a `GroupedConsoleFormatter` subclass that overrides `toString()` / `toJSON()` so the existing formatter interface stays clean.

I'll also drop `lastId` since the first message's ID is sufficient for identical messages.

I'll work on the changes and push an update — would appreciate another look when it's ready!

masamaru0513:
@OrKoN 
Refactored as suggested, here's a summary of the changes:

- Introduced `GroupedConsoleFormatter` subclass that overrides `toString()` / `toJSON()`
- Dropped `lastId`; only the first message's ID is used
- No new methods added to the formatter interface
- `ConsoleFormatter` and `GroupedConsoleFormatter` are interchangeable via the same interface

Ready for another look when you get a chance. Thanks!

OrKoN:
Please update the PR description accordingly.

masamaru0513:
@OrKoN 
Updated the PR description. Sorry about that.

OrKoN (approved):
Thanks!

PR #1945, opened 2026-04-23
feat: support MCP client roots feature

### LINKED ISSUE #1860: Support the roots feature from the MCP spec

See https://modelcontextprotocol.io/specification/2025-06-18/client/roots

we should declare that chrome-devtools-mcp supports this capability and check the filesystem access against the client supplied routes.

cc @natorion

OrKoN:
This PR implements https://modelcontextprotocol.io/specification/2025-11-25/client/roots If client specifies roots, all reads and writes to the file system originating in tool calls will denied (including tmp files). The client specified empty list of roots, all filesystem access will be restricted.

Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1860

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -6,6 +6,7 @@
 
 import fs from 'node:fs/promises';
 import path from 'node:path';
+import {fileURLToPath} from 'node:url';
 
 import type {TargetUniverse} from './DevtoolsUtils.js';
 import {UniverseManager} from './DevtoolsUtils.js';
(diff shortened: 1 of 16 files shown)


src/McpContext.ts
```diff
@@ -18,6 +19,8 @@ import {
   type ListenerMap,
   type UncaughtError,
 } from './PageCollector.js';
+import {Locator} from './third_party/index.js';
+import {PredefinedNetworkConditions} from './third_party/index.js';
```
wolfib: nit: `import {Locator, PredefinedNetworkConditions, type Root} from './third_party/index.js;`

wolfib (approved):

OrKoN:
I realized I might have had a bug in index.ts. PT(another)L.

wolfib (approved):

OrKoN:
Sorry, I realised again that we should treat empty roots differently from undefined roots (rejecting any file access with empty roots). PTAL

src/McpContext.ts
```diff
@@ -643,13 +674,18 @@ export class McpContext implements Context {
     data: Uint8Array<ArrayBufferLike>,
     filename: string,
   ): Promise<{filepath: string}> {
-    return await saveTemporaryFile(data, filename);
+    const filepath = await getTempFilePath(filename);
+    this.validatePath(filepath);
+    await fs.writeFile(filepath, data);
```
wolfib: Should this be wrapped in a try..catch?

wolfib (approved):

src/tools/ToolDefinition.ts
```diff
@@ -167,9 +167,10 @@ export type SupportedExtensions =
   | '.json.gz';
 
 /**
- * Only add methods required by tools/*.
+ * Only add methods used by tools/*.
  */
 export type Context = Readonly<{
+  validatePath(filePath: string): void;
```
Lightning00Blade: Should this be 
```suggestion
  validatePath(filePath?: string): void;
```
So we can ignore some if statements?

OrKoN: Good idea, let me re-visit this next week.

src/McpContext.ts
OrKoN: I think no, why would it need a try/catch?

wolfib: `fs.writeFile` might throw an exception. Maybe it's already caught elsewhere, but it's not obvious.

The previous `saveTemporaryFile` did catch exceptions too.

src/tools/ToolDefinition.ts
OrKoN: we should see how we could mark filepath parameters so that we can validate them centrally later on

Issue #1069, opened 2026-03-01
Dialog Handling is still not fixed


voycey:
### Description of the bug

Any kind of dialog that pops up when Claude Code is operating the browser causes the process to stall:

https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/713

This issue seemed to previously work on this but the issue still persists in the latest version.

### Reproduction

Have claude use any page that includes a standard HTML Confirm dialog

### Expectation

It is able to confirm the dialogbox and continue with the browser operation

### MCP configuration

Standard

### Chrome DevTools MCP version

latest

### Chrome version

Version 145.0.7632.109 (Official Build) (arm64)

### Coding agent version

latest

### Model version

Opus 4.6

### Operating system

macOS

### Extra checklist

- [ ] I want to provide a PR to fix this bug

OrKoN:
@voycey could you provide an example or a screencast? the tests for dialog handling are all passing. Please also double check the exact version of the MCP server used.

voycey:
@OrKoN hit this again today:

On any webpage generated "confirm" box it cannot handle it when using this MCP server.


```
⏺ chrome-devtools - click (MCP)(uid: "83_77")
  ⎿  # click response
     Successfully clicked on the element

⏺ chrome-devtools - wait_for (MCP)(text: ["successfully","Error","failed","Knowledge Bases"], timeout: 10000)
  ⎿  Error: Timed out after waiting 10000ms
     Cause: Locator.waitHandle

⏺ chrome-devtools - take_snapshot (MCP)
  ⎿  Error: Accessibility.getFullAXTree timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher
     timeout if needed.

⏺ chrome-devtools - take_snapshot (MCP)
  ⎿  # take_snapshot response
     # Open dialog
     alert: Knowledge base created successfully!.
     … +43 lines (ctrl+o to expand)

<Hung here for about 20 minutes until I manually went into Chrome to click the dialog box then it continued immediately>

⏺ KB created successfully. Let me dismiss the dialog and verify the per-user file.
```

Here's a fresh one:

<img width="1786" height="463" alt="Image" src="https://github.com/user-attachments/assets/e690affe-97d7-4087-a225-9ce863ca74d6" />

```

⏺ Good, reloaded. Now let me create a new KB in the browser — should be on the KB list page already as voycey+2:

[...]

PR #1953, opened 2026-04-24
fix: handle errors due to open dialogs during tool calls

samiyac:
This change is part of the fix for #1069 
- Errors during tool calls are now consumed and handled as part of McpResponse format

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -232,31 +232,35 @@ export async function createMcpServer(
             : new McpResponse(serverArgs);
 
           response.setRedactNetworkHeaders(serverArgs.redactNetworkHeaders);
-          if ('pageScoped' in tool && tool.pageScoped) {
-            const page =
-              serverArgs.experimentalPageIdRouting &&
-              params.pageId &&
-              !serverArgs.slim
-                ? context.getPageById(params.pageId)
-                : context.getSelectedMcpPage();
-            response.setPage(page);
-            await tool.handler(
-              {
-                params,
-                page,
-              },
-              response,
-              context,
-            );
-          } else {
-            await tool.handler(
-              // @ts-expect-error types do not match.
-              {
-                params,
-              },
-              response,
-              context,
-            );
+          try {
+            if ('pageScoped' in tool && tool.pageScoped) {
+              const page =
+                serverArgs.experimentalPageIdRouting &&
+                params.pageId &&
+                !serverArgs.slim
+                  ? context.getPageById(params.pageId)
+                  : context.getSelectedMcpPage();
+              response.setPage(page);
+              await tool.handler(
+                {
+                  params,
+                  page,
+                },
+                response,
+                context,
+              );
+            } else {
+              await tool.handler(
+                // @ts-expect-error types do not match.
+                {
+                  params,
+                },
+                response,
+                context,
+              );
+            }
+          } catch (err) {
+            response.setError(err);
           }
           const {content, structuredContent} = await response.handle(
             tool.name,
@@ -267,6 +271,9 @@ export async function createMcpServer(
           } = {
             content,
           };
+          if (response.error) {
+            result.isError = true;
+          }
           success = true;
           if (serverArgs.experimentalStructuredContent) {
             result.structuredContent = structuredContent as Record<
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -226,19 +226,22 @@ export async function createMcpServer(
...
           logger(`${tool.name} context: resolved`);
-          await context.detectOpenDevToolsWindows();
+          const page =
+            serverArgs.experimentalPageIdRouting &&
+            params.pageId &&
+            !serverArgs.slim
+              ? context.getPageById(params.pageId)
+              : context.getSelectedMcpPage();
+          // If there is a dialog open, we will skip devtools detection
+          if (!page.getDialog()) {
```
OrKoN: I think we should skip it if there is a dialog on any page not only on the current one?

OrKoN (commented):
Let's add tests?

src/tools/ToolDefinition.ts
```diff
@@ -348,6 +348,18 @@ export function definePageTool<
   return {
     ...definition,
     pageScoped: true,
+    handler: async (request, response, context) => {
```
OrKoN: there is a function style declaration above. Perhaps we should just add it to src/index.ts server.registerTool's callback?

src/index.ts
samiyac: I just tested if only checking current page for dialog is enough - and it seems like it is. I tried running new_page tool multiple times on different pages while the first page had a dialog open - and it worked fine.

OrKoN: interesting, what is the reason then why we time out here?

OrKoN (approved):

src/index.ts
samiyac: As discussed offline, removing this check for now as it might not actually be required.

src/tools/ToolDefinition.ts
samiyac: Done, moved to server.registerTool

Added to the pagescoped if block only since that block corresponds to both of these declarations.

samiyac:
Added a test

tests/index.test.ts
```diff
@@ -245,6 +245,38 @@ describe('e2e', () => {
...
+      // Trigger the dialog
+      const result = await client.callTool({
+        name: 'click',
+        arguments: {
+          uid,
+        },
+      });
+
+      const content = result.content as TextContent[];
+      assert.match(content[0].text, /\[Blocked\] Action triggered a dialog\./);
```
OrKoN: could we assert complete response via a snapshot assertion? I want to make sure there is no duplication in the full response

samiyac: Done

tests/index.test.js.snapshot
```diff
@@ -5,3 +5,7 @@ exports[`e2e > calls a tool 1`] = `
 exports[`e2e > calls a tool multiple times 1`] = `
 [{"type":"text","text":"## Pages\\n1: about:blank [selected]"}]
 `;
+
+exports[`e2e > returns blocked message when dialog is opened during tool execution 1`] = `
+[{"type":"text","text":"[Blocked] Action triggered a dialog.\\n# Open dialog\\nalert: test dialog.\\nCall handle_dialog to handle it before continuing."}]
```
OrKoN: so we already report `Open dialog\\nalert: test dialog.\\nCall handle_dialog to handle it before continuing.`? why do we need an additional blocked message?

samiyac: Updated logic as per our discussion.

tests/index.test.ts
```diff
@@ -245,6 +245,42 @@ describe('e2e', () => {
...
+      // Trigger the dialog
+      const result = await client.callTool({
+        name: 'click',
+        arguments: {
+          uid,
+        },
+      });
+
+      const content = result.content as TextContent[];
+      assert.match(
```
OrKoN: nit: we can remove assertion in favour of the snapshot

tests/index.test.ts
```diff
@@ -245,6 +245,42 @@ describe('e2e', () => {
...
+
+      // Trigger the dialog
+      const result = await client.callTool({
+        name: 'click',
+        arguments: {
+          uid,
+        },
+      });
+
+      const content = result.content as TextContent[];
```
OrKoN: I think we would still want to mark the response with isError=true (see https://modelcontextprotocol.io/specification/2025-11-25/server/tools#error-handling) that would mean that the Response handle()/format() should return an indication that we encountered an error.

samiyac: Makes sense, I have used `response.error` to set `result.isError`

OrKoN (approved):

PR #1954, opened 2026-04-24
feat: report navigated URL in action responses

### LINKED ISSUE #243: Actions that result in a navigation should report the new page URL in the response

masamaru0513:
## Summary

Report the new page URL when an action triggers a navigation, so there is no need to re-query the pages list.

Fixes #243

> Note: I am aware of the existing #1853 which addresses the same issue. This PR incorporates the review feedback from #1853 (by @Lightning00Blade) — specifically using `navigatedToUrl` naming, extracting a shared `appendNavigatedToUrl` helper in `WaitForHelper`, and applying it to `evaluate_script` as well.

## Changes
* `McpPage.waitForEventsAfterAction` now compares the page URL before and after the action, returning `{navigatedToUrl}` when a navigation occurred
* Added `appendNavigatedToUrl` helper in `WaitForHelper.ts` for consistent URL reporting across tools
* All input tools (`click`, `click_at`, `hover`, `fill`, `fill_form`, `type_text`, `drag`, `press_key`) use the helper to append `Navigated to <URL>` when a navigation is detected
* `evaluate_script` also reports navigated URLs using the same helper
* No change to `navigate_page` or `new_page` since they already report the URL explicitly

## Output example

Before:

`Successfully clicked on the element`

After:

```
Successfully clicked on the element
Navigated to https://www.iana.org/help/example-domains
```

## Testing
* Unit tests added for:
  * Click triggering navigation → reports URL
  * Click without navigation → no "Navigated to" in response
* Manual verification:
  * `click` on a link → reports navigated URL ✅
  * `press_key` Enter on a focused link → reports navigated URL ✅
  * `evaluate_script` with `window.location.href` change → reports navigated URL ✅
  * `evaluate_script` without navigation → no "Navigated to" ✅
  * Hash-only change (anchor link) → no "Navigated to" (same-document navigations are excluded by WaitForHelper) ✅
  * `navigate_page` back/forward → existing behavior preserved ✅

### DIFF
--- a/src/McpPage.ts
+++ b/src/McpPage.ts
@@ -129,15 +129,18 @@ export class McpPage implements ContextPage {
     return new WaitForHelper(this.pptrPage, cpuMultiplier, networkMultiplier);
   }
 
-  waitForEventsAfterAction(
+  async waitForEventsAfterAction(
     action: () => Promise<unknown>,
     options?: {timeout?: number; handleDialog?: 'accept' | 'dismiss' | string},
-  ): Promise<void> {
+  ): Promise<{navigatedToUrl?: string}> {
+    const urlBefore = this.pptrPage.url();
     const helper = this.createWaitForHelper(
       this.cpuThrottlingRate,
       getNetworkMultiplierFromString(this.networkConditions),
     );
-    return helper.waitForEventsAfterAction(action, options);
+    await helper.waitForEventsAfterAction(action, options);
+    const urlAfter = this.pptrPage.url();
+    return urlAfter === urlBefore ? {} : {navigatedToUrl: urlAfter};
   }
 
   dispose(): void {
--- a/tests/tools/input.test.ts
+++ b/tests/tools/input.test.ts
@@ -130,6 +130,60 @@ describe('input', () => {
       });
     });
 
+    it('reports navigated URL after click', async () => {
+      server.addHtmlRoute('/nav-link', html`<a href="/nav-target">Go</a>`);
+      server.addHtmlRoute('/nav-target', html`<main>Target</main>`);
+
+      await withMcpContext(async (response, context) => {
+        const page = context.getSelectedPptrPage();
+        await page.goto(server.getRoute('/nav-link'));
+        context.getSelectedMcpPage().textSnapshot = await TextSnapshot.create(
+          context.getSelectedMcpPage(),
+        );
+        await click.handler(
+          {
+            params: {uid: '1_1'},
+            page: context.getSelectedMcpPage(),
+          },
+          response,
+          context,
+        );
+        assert.strictEqual(
+          response.responseLines[0],
+          'Successfully clicked on the element',
+        );
+        assert.strictEqual(
+          response.responseLines[1],
+          `Navigated to ${server.getRoute('/nav-target')}`,
+        );
+      });
+    });
+
+    it('does not report navigated URL when no navigation occurs', async () => {
+      await withMcpContext(async (response, context) => {
+        const page = context.getSelectedPptrPage();
+        await page.setContent(
+          html`<button onclick="this.innerText = 'clicked';">test</button>`,
+        );
+        context.getSelectedMcpPage().textSnapshot = await TextSnapshot.create(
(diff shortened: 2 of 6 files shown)


src/tools/pages.ts
```diff
@@ -201,6 +201,9 @@ export const newPage = defineTool(args => {
         request.params.timeout,
       );
 
+      response.appendResponseLine(
```
wolfib: `new_page` already returns a list of open pages, no need to add anything

src/tools/pages.ts
```diff
@@ -303,8 +306,9 @@ export const navigatePage = definePageTool(args => {
                 }
                 try {
                   await page.pptrPage.goto(request.params.url, options);
+                  const finalUrl = page.pptrPage.url();
                   response.appendResponseLine(
-                    `Successfully navigated to ${request.params.url}.`,
+                    `Successfully navigated to ${finalUrl}.`,
```
wolfib: For redirects (e.g. from foo to bar), the tool currently returns: "Successfully navigated to foo.
## Pages:
1: bar"
I think this is already working as intended as it gives more info than just reporting the final URL twice.

src/McpPage.ts
```diff
@@ -121,15 +121,21 @@ export class McpPage implements ContextPage {
...
+  ): Promise<{navigatedToUrl?: string}> {
+    const urlBefore = this.pptrPage.url();
     const helper = this.createWaitForHelper(
       this.cpuThrottlingRate,
       getNetworkMultiplierFromString(this.networkConditions),
     );
-    return helper.waitForEventsAfterAction(action, options);
+    await helper.waitForEventsAfterAction(action, options);
+    const urlAfter = this.pptrPage.url();
+    if (urlAfter !== urlBefore) {
```
wolfib: nit: Could be shortened to
`return (urlAfter === urlBefore) ? {} : {navigatedToUrl: urlAfter};`

tests/tools/input.test.ts
```diff
@@ -130,6 +130,64 @@ describe('input', () => {
...
+          response.responseLines[0],
+          'Successfully clicked on the element',
+        );
+        assert.ok(
+          response.responseLines[1]?.startsWith('Navigated to '),
+          `Expected "Navigated to" but got: ${response.responseLines[1]}`,
+        );
+        assert.ok(
+          response.responseLines[1]?.includes('/nav-target'),
+          `Expected URL to contain /nav-target but got: ${response.responseLines[1]}`,
```
wolfib: Instead of comparing the line in 2 (imprecise) parts, could we construct the whole response line (with `server.getRoute`) and check for equality?

wolfib:
Thanks! This mostly looks good to me, added a few comments.
@Lightning00Blade PTAL since you reviewed the previous PR

masamaru0513:
Thanks for the review @wolfib! Addressed the feedback:
- Removed `new_page` URL reporting (page list already covers it)
- Reverted `navigate_page` to use the requested URL (showing both requested + final in page list is more informative)
- Shortened `McpPage.ts` to a ternary
- Updated test to use `server.getRoute` with `strictEqual`

Please take another look when you get a chance.

src/tools/pages.ts
masamaru0513: Done.

src/tools/pages.ts
masamaru0513: Done.

src/McpPage.ts
masamaru0513: Done.

tests/tools/input.test.ts
masamaru0513: Done.

wolfib:
Thanks! The update LGTM, but I defer to @Lightning00Blade

OrKoN:
@Lightning00Blade PTAL

OrKoN:
Closing in favour of https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1853

PR #1961, opened 2026-04-27
fix: respect npm_config_registry for update checks

### LINKED ISSUE #1943: Update check hardcodes registry.npmjs.org, ignoring user's npm registry configuration

The update check script [check-latest-version.ts](https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/src/bin/check-latest-version.ts) uses a hardcoded URL:

```js
const response = await fetch('https://registry.npmjs.org/chrome-devtools-mcp/latest');
```

This uses Node.js native `fetch()`, which completely bypasses the user's `.npmrc` registry configuration. For users behind corporate proxies or using private/internal registries (e.g. JFrog Artifactory), this request either gets blocked or fails silently on every startup.

The script should respect the user's npm registry configuration instead of hardcoding the public registry. Some options:

- Read the `npm_config_registry` environment variable (npm automatically sets this when invoked via `npx`)
- Fall back to `https://registry.npmjs.org` only if no custom registry is configured

For example:
```js
const registry = process.env.npm_config_registry?.replace(/\/$/, '') || 'https://registry.npmjs.org';
const response = await fetch(`${registry}/chrome-devtools-mcp/latest`);
```

- Setting `CHROME_DEVTOOLS_MCP_NO_UPDATE_CHECKS=1` to disable the check entirely. This works but means losing update notifications.
- Patching the script locally, but it gets overwritten whenever `npx` refreshes the package.

### Additional context

The issue also affects users who configure npm registries via environment variables or project-level `.npmrc` files. The `CHROME_DEVTOOLS_MCP_NO_UPDATE_CHECKS` env var serves as a workaround but shouldn't be necessary for users who simply have a different registry configured.

Will-hxw:
## Summary

Use `npm_config_registry` environment variable for npm registry URL in update checks, instead of hardcoding `registry.npmjs.org`.

## Fix

`check-latest-version.ts` now respects the user's npm registry configuration. This helps users behind corporate proxies or using private registries (e.g., JFrog Artifactory).

When `npm_config_registry` is not set, falls back to the default public registry.

## Validation

- [x] `npm run build` passes
- [x] `npm test` passes

Fixes #1943

### DIFF
--- a/src/bin/check-latest-version.ts
+++ b/src/bin/check-latest-version.ts
@@ -12,8 +12,11 @@ const cachePath = process.argv[2];
 
 if (cachePath) {
   try {
+    const registry =
+      process.env.npm_config_registry?.replace(/\/$/, '') ||
+      'https://registry.npmjs.org';
     const response = await fetch(
-      'https://registry.npmjs.org/chrome-devtools-mcp/latest',
+      `${registry}/chrome-devtools-mcp/latest`,
     );
     const data = response.ok ? await response.json() : null;
 
--- a/src/bin/check-latest-version.ts
+++ b/src/bin/check-latest-version.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import {execSync} from 'node:child_process';
 import fs from 'node:fs/promises';
 import path from 'node:path';
 import process from 'node:process';
@@ -12,9 +13,15 @@ const cachePath = process.argv[2];
 
 if (cachePath) {
   try {
-    const registry =
-      process.env.npm_config_registry?.replace(/\/$/, '') ||
-      'https://registry.npmjs.org';
+    let registry;
+    try {
+      registry = execSync('npm config get registry', {
+        encoding: 'utf8',
+      }).trim().replace(/\/$/, '');
+    } catch {
+      // npm not on PATH, fall back to default
+    }
+    registry ||= 'https://registry.npmjs.org';
     const response = await fetch(
       `${registry}/chrome-devtools-mcp/latest`,
     );


Will-hxw:
Hi @coty @precision , sorry to bother you. This PR has been open for a while, so I wanted to kindly check whether you might have time to review it or let me know if any changes are needed.I’m happy to make adjustments if required. Thanks!

wolfib:
Thanks!

As discussed in [here](https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1943#issuecomment-4303782483), this does not solve the case where the DevTools CLI is called directly. @mathiasbynens WDYT? Should the suggestion from [here](https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1943#issuecomment-4307831766) be added?

mathiasbynens:
> Thanks!
> 
> As discussed in [here](https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1943#issuecomment-4303782483), this does not solve the case where the DevTools CLI is called directly. @mathiasbynens WDYT? Should the suggestion from [here](https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1943#issuecomment-4307831766) be added?

Yeah, if we need to support this then let’s do it properly and support all use cases, not just those through `npx` or `npm run`.

The “proper” way to do it would be to use [`@npmcli/config`](https://www.npmjs.com/package/@npmcli/config), but this comes at the cost of a dependency.

Will-hxw:
Thanks for the suggestion. Adding @npmcli/config as a dependency would indeed handle all cases properly, but you're right that it introduces a new dependency. 

Given the PR is already behind and this would be a meaningful scope expansion, I think it makes sense to treat this as a follow-up PR rather than growing this one further. Would you be okay with merging the current fix as-is (for the npx/npm case) and addressing the direct-CLI case in a separate PR with the proper implementation?

Will-hxw:
@wolfib Thanks for the follow-up! The plan is to merge the current fix for the npx/npm case, then address the direct-CLI scenario in a follow-up with the proper `@npmcli/config` implementation. This keeps this PR focused and lets us do the full solution justice. I'll open a new issue to track the direct-CLI work after this merges.

PR #1962, opened 2026-04-27
feat(cli): generate commands for conditional tools

### LINKED ISSUE #1933: Update CLI generator to generate interfaces for all possible tools
- generate CLI for all possible tools (turning on all conditional tools)
- add additional skills for conditional CLI commands
- ensure that the tools can be still disabled via server start args but it is fine if they remain in the CLI interface (this is a discussion item).

nroscino:
This PR adds all tools in the CLI interface. When a tool is not enabled the server responds with an error guiding the user on how to enable the category or the experiment.

Closes: https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1933

### DIFF
--- a/scripts/generate-cli.ts
+++ b/scripts/generate-cli.ts
@@ -11,7 +11,12 @@ import {Client} from '@modelcontextprotocol/sdk/client/index.js';
 import {StdioClientTransport} from '@modelcontextprotocol/sdk/client/stdio.js';
 
 import {parseArguments} from '../build/src/bin/chrome-devtools-mcp-cli-options.js';
-import {labels, ToolCategory} from '../build/src/tools/categories.js';
+import {CONDITION_TO_FLAG, buildFlag} from '../build/src/index.js';
+import {
+  labels,
+  ToolCategory,
+  OFF_BY_DEFAULT_CATEGORIES,
+} from '../build/src/tools/categories.js';
 import {createTools} from '../build/src/tools/tools.js';
 
 const OUTPUT_PATH = path.join(
@@ -83,10 +88,7 @@ function schemaToCLIOptions(schema: JsonSchema): CliOption[] {
   const properties = schema.properties;
   return Object.entries(properties).map(([name, prop]) => {
     const isRequired = required.includes(name);
-    let description = prop.description || '';
-    if (isRequired) {
-      description += ' (required)';
-    }
+    const description = prop.description || '';
     if (typeof prop.type !== 'string') {
       throw new Error(
         `Property ${name} has a complex type not supported by CLI.`,
@@ -107,12 +109,12 @@ async function generateCli() {
   const tools = await fetchTools();
 
   const staticTools = createTools(parseArguments());
-  const toolNameToCategory = new Map<string, string>();
+  const toolNameToCategoryEnum = new Map<string, string>();
+  const toolNameToConditions = new Map<string, string[]>();
+
   for (const tool of staticTools) {
-    toolNameToCategory.set(
-      tool.name,
-      labels[tool.annotations.category as keyof typeof labels],
-    );
+    toolNameToCategoryEnum.set(tool.name, tool.annotations.category);
+    toolNameToConditions.set(tool.name, tool.annotations.conditions || []);
   }
 
   // Sort tools by name
@@ -134,7 +136,7 @@ async function generateCli() {
         return false;
       }
       // Skipping in_page tools as they are not launched yet
-      if (toolNameToCategory.get(tool.name) === labels[ToolCategory.IN_PAGE]) {
+      if (toolNameToCategoryEnum.get(tool.name) === ToolCategory.IN_PAGE) {
         return false;
       }
       return true;
@@ -151,15 +153,40 @@ async function generateCli() {
     for (const opt of options) {
       args[opt.name] = opt;
     }
-    const category = toolNameToCategory.get(tool.name);
-    if (!category) {
+
+    const categoryEnum = toolNameToCategoryEnum.get(tool.name);
+    if (!categoryEnum) {
       throw new Error(`Tool ${tool.name} has no category.`);
     }
+    const category = labels[categoryEnum as unknown as keyof typeof labels];
     if (!tool.description) {
-      throw new Error(`Tool ${tool.name} is missing descripttion`);
+      throw new Error(`Tool ${tool.name} is missing description`);
     }
+
+    let description = tool.description;
+    const requiredFlags: string[] = [];
+
+    const isOffByDefault = OFF_BY_DEFAULT_CATEGORIES.includes(categoryEnum);
+    if (isOffByDefault) {
+      const categoryFlag = buildFlag(categoryEnum);
+      requiredFlags.push(`--${categoryFlag}=true`);
+    }
+
+    const conditions = toolNameToConditions.get(tool.name) || [];
+    for (const condition of conditions) {
+      const flag =
+        CONDITION_TO_FLAG[condition as keyof typeof CONDITION_TO_FLAG];
+      if (flag) {
+        requiredFlags.push(`--${flag}=true`);
+      }
+    }
+
+    if (requiredFlags.length > 0) {
+      description += ` (requires flag: ${requiredFlags.join(', ')})`;
+    }
+
     commands[tool.name] = {
-      description: tool.description,
+      description,
       category,
       args,
     };
--- a/scripts/generate-cli.ts
+++ b/scripts/generate-cli.ts
@@ -11,7 +11,7 @@ import {Client} from '@modelcontextprotocol/sdk/client/index.js';
 import {StdioClientTransport} from '@modelcontextprotocol/sdk/client/stdio.js';
 
 import {parseArguments} from '../build/src/bin/chrome-devtools-mcp-cli-options.js';
-import {labels} from '../build/src/tools/categories.js';
+import {labels, ToolCategory} from '../build/src/tools/categories.js';
 import {createTools} from '../build/src/tools/tools.js';
 
 const OUTPUT_PATH = path.join(
@@ -29,7 +29,7 @@ async function fetchTools() {
 
   const transport = new StdioClientTransport({
     command: 'node',
-    args: [serverPath],
+    args: [serverPath, '--viaCli'],
     env: {...process.env, CHROME_DEVTOOLS_MCP_NO_USAGE_STATISTICS: 'true'},
   });
 
@@ -83,7 +83,10 @@ function schemaToCLIOptions(schema: JsonSchema): CliOption[] {
   const properties = schema.properties;
   return Object.entries(properties).map(([name, prop]) => {
     const isRequired = required.includes(name);
-    const description = prop.description || '';
+    let description = prop.description || '';
+    if (isRequired) {
+      description += ' (required)';
+    }
     if (typeof prop.type !== 'string') {
       throw new Error(
         `Property ${name} has a complex type not supported by CLI.`,
@@ -103,6 +106,15 @@ function schemaToCLIOptions(schema: JsonSchema): CliOption[] {
 async function generateCli() {
   const tools = await fetchTools();
 
+  const staticTools = createTools(parseArguments());
+  const toolNameToCategory = new Map<string, string>();
+  for (const tool of staticTools) {
+    toolNameToCategory.set(
+      tool.name,
+      labels[tool.annotations.category as keyof typeof labels],
+    );
+  }
+
   // Sort tools by name
   const sortedTools = tools
     .sort((a, b) => a.name.localeCompare(b.name))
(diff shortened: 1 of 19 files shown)


skills/chrome-devtools-cli/SKILL.md
```diff
@@ -123,6 +123,30 @@ chrome-devtools take_snapshot # Take a text snapshot of the page from the a11y t
...
+## Experimental Features
+
+Experimental tools are disabled by default. Enable them with the corresponding flag during `start`.
+
+```bash
+chrome-devtools click_at 100 200 # Clicks at the provided coordinates (requires --experimentalVision=true)
+chrome-devtools screencast_start # Starts a screencast recording (requires --experimentalScreencast=true and ffmpeg)
+chrome-devtools screencast_stop # Stops the active screencast
+chrome-devtools list_webmcp_tools # List all WebMCP tools (requires --experimentalWebmcp=true)
+chrome-devtools get_tab_id # Get the tab ID for the current page (requires --experimentalInteropTools=true)
```
OrKoN: we should exclude this one

OrKoN: this is for internal integrations and would not be of use for CL users

nroscino: Ok! I was not sure about this one either. will remove.

skills/chrome-devtools-cli/SKILL.md
```diff
@@ -123,6 +123,30 @@ chrome-devtools take_snapshot # Take a text snapshot of the page from the a11y t
 chrome-devtools take_snapshot --verbose true --filePath "s.txt" # Take a verbose snapshot and save to file
 ```
 
+## Extensions
+
+Extensions tools are disabled by default. Enable them with `chrome-devtools start --categoryExtensions=true`.
```
OrKoN: let's change the default for the CLI here so that in CLI the extension category is enabled by default (you can see other examples for changing defaults in chrome-devtools.ts).

src/bin/chrome-devtools-cli-options.ts
```diff
@@ -233,18 +314,31 @@ export const commands: Commands = {
         name: 'requestFilePath',
         type: 'string',
         description:
-          'The absolute or relative path to save the request body to. If omitted, the body is returned inline.',
+          'The absolute or relative path to a .network-request file to save the request body to. If omitted, the body is returned inline.',
```
OrKoN, line 301: Hm.. should not this update already be on the main branch?

src/bin/chrome-devtools-cli-options.ts
```diff
@@ -49,6 +49,37 @@ export const commands: Commands = {
       },
     },
   },
+  click_at: {
+    description: 'Clicks at the provided coordinates',
```
OrKoN: could we add required flags to the descriptions of the CLI commands?

src/bin/chrome-devtools-cli-options.ts
```diff
@@ -155,11 +186,55 @@ export const commands: Commands = {
...
+      dialogAction: {
+        name: 'dialogAction',
+        type: 'string',
+        description:
+          'Handle dialogs while execution. "accept", "dismiss", or string for response of window.prompt. Defaults to accept.',
+        required: false,
+      },
+    },
+  },
+  execute_in_page_tool: {
```
OrKoN: let's remove everything about in_page tools for now, it is not launched. cc @wolfib

src/bin/chrome-devtools-cli-options.ts
OrKoN, line 1: Did we have a bug and used the wrong file?

src/bin/chrome-devtools-cli-options.ts
nroscino: I understood what happened. I think i fixed a bug. The `chrome-devtools.ts` was importing the `chrome-devtools-cli-options.ts` so I was generating that one (changed the path in the `generate-cli.ts`). But I saw we also had the `cliDefinitions.ts` that was newer and mentioned in a older commit so decided to preserve that one and changed the import instead. Now I have all these updates that should have been there already and they are also reflected in the cli correctly. 
To avoid misunderstandings with this file I removed it from the repo. Let me know if I have misunderstood something here!

src/bin/chrome-devtools-cli-options.ts
nroscino: Yes, I explained it in the above comment. Let me know also what you think

src/bin/chrome-devtools-cli-options.ts
OrKoN: cliDefinitions is the wrong file. See https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1969

OrKoN: Thanks for catching this! Let's rebase after https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1969 lands to make sure we have a clean diff for this feature.

nroscino: No problem! 
Just did the rebase! PTAL

src/bin/chrome-devtools-cli-options.ts
OrKoN: I think this one was misunderstood, I meant that we should add a text saying which flags have to be passed to the `start` command to enable the command.

nroscino: ah 🤦  
Ok! Will changed this!

src/bin/chrome-devtools-cli-options.ts
```diff
@@ -58,7 +89,7 @@ export const commands: Commands = {
         name: 'pageId',
         type: 'number',
         description:
-          'The ID of the page to close. Call list_pages to list pages.',
+          'The ID of the page to close. Call list_pages to list pages. (required)',
```
OrKoN: please revert `(required)`, I think this is already passed via the required boolean attributes.

OrKoN:
The tests seem to fail still.

src/bin/chrome-devtools.ts
```diff
@@ -47,29 +47,12 @@ const startCliOptions = {
...
-delete startCliOptions.categoryExtensions;
-// Always on in CLI.
-delete startCliOptions.experimentalStructuredContent;
-// Change the defaults.
-if (!('default' in cliOptions.headless)) {
-  throw new Error('headless cli option unexpectedly does not have a default');
-}
-if ('default' in cliOptions.isolated) {
-  throw new Error('isolated cli option unexpectedly has a default');
-}
```
OrKoN, line 69: what was the reason to remove these lines?

src/bin/chrome-devtools.ts
```diff
@@ -47,29 +47,12 @@ const startCliOptions = {
...
-delete startCliOptions.experimentalVision;
-delete startCliOptions.experimentalWebmcp;
-delete startCliOptions.experimentalInteropTools;
-delete startCliOptions.experimentalScreencast;
-delete startCliOptions.categoryEmulation;
-delete startCliOptions.categoryPerformance;
-delete startCliOptions.categoryNetwork;
-delete startCliOptions.categoryExtensions;
-// Always on in CLI.
-delete startCliOptions.experimentalStructuredContent;
```
OrKoN, line 62: why was this line deleted?

src/bin/chrome-devtools.ts
```diff
@@ -47,29 +47,12 @@ const startCliOptions = {
 delete startCliOptions.autoConnect;
 // Missing CLI serialization.
 delete startCliOptions.viewport;
-// CLI is generated based on the default tool definitions. To enable conditional
-// tools, they need to be enabled during CLI generation.
-delete startCliOptions.experimentalPageIdRouting;
```
OrKoN, line 52: I think we should keep this one.

src/bin/chrome-devtools.ts
```diff
@@ -47,29 +47,12 @@ const startCliOptions = {
...
-delete startCliOptions.experimentalVision;
-delete startCliOptions.experimentalWebmcp;
-delete startCliOptions.experimentalInteropTools;
```
OrKoN, line 55: keep this one

src/index.ts
```diff
@@ -24,12 +24,124 @@ import {
...
 import {createTools} from './tools/tools.js';
 import {VERSION} from './version.js';
 
+export const CONDITION_TO_FLAG: Record<string, string> = {
+  computerVision: 'experimentalVision',
+  experimentalMemory: 'experimentalMemory',
+  experimentalInteropTools: 'experimentalInteropTools',
+  screencast: 'experimentalScreencast',
+  experimentalWebmcp: 'experimentalWebmcp',
+};
```
OrKoN: could we instead have a convention that a condition has to match the flag name?

src/bin/chrome-devtools.ts
nroscino: It was to simplify, the `cliOptions` have always those set in that way, but I have restored those line to enforce the format for `cliOptions`, just in case it will change in future.

src/bin/chrome-devtools.ts
nroscino: Restored also this one

src/index.ts
nroscino: Done! What do you think about doing a similar thing with the ToolCategory? 
All of them follow the category+[Capital category name], but IN_PAGE doesn't. I was thinking of changing IN_PAGE to `inPageTools`. WDYT?

OrKoN: Yes, I think we should match the naming for IN_PAGE cc @wolfib

nroscino: Ok, will change it, so I can remove the check!

OrKoN: I think we need to do the same for webmcp. Can we support `experimentalCategoryInPage`?

nroscino: you mean to change `experimentalInPageTools` to `experimentalCategoryInPage`?

nroscino: about webmcp not sure I understood either, you want me to add a different category for `webmcp` instead of using the debug?

OrKoN: Yeah, I think we can change webmcp to have its own category. But let's do it separately probably. For now, I wonder if we could just retain `experimental` prefix for the in-page tools category.

scripts/generate-cli.ts
```diff
@@ -117,18 +131,17 @@ async function generateCli() {
       if (tool.name === 'wait_for') {
         return false;
       }
+      // Skipping get_tab_id as it is for internal integrations
+      if (tool.name === 'get_tab_id') {
```
OrKoN, line 135: do we need this if we do `delete startCliOptions.experimentalInteropTools;`?

scripts/generate-cli.ts
```diff
@@ -117,18 +131,17 @@ async function generateCli() {
...
+        return false;
+      }
+      // Skipping in_page tools as they are not launched yet
+      if (toolNameToCategoryEnum.get(tool.name) === ToolCategory.IN_PAGE) {
```
OrKoN, line 139: I wonder if we can remove this if we delete the category cli flag from the cli options in chrome-devtools.ts?

src/index.ts
nroscino: Ok for webmcp.
Not sure I understood for in-page. Should I delete the category and rely only on the condition `experimentalCategoryInPage`? or we want to delete the condition altogether for the in-page and rename the category to: `experimentalCategoryInPage` relying only on that?

OrKoN: I mean the following:

- definition in tools:

```
annotations: {
    category: ToolCategory.IN_PAGE,
    readOnlyHint: false,
  },
```

- the flag for this category should be `experimentalCategoryInPage` 
(`experimentalCategory<CategoryName>`)

For purposes of filtering we should treat `experimentalCategory` the same way as `category`. WDYT?

nroscino: Ok, clear! What about inverting it to `categoryExperimentalInPage`? this way the `ToolCategory` will have   `IN_PAGE='experimentalInPage'` and we can keep the `buildFlag` method the same without adding a special condition for experimental categories. WDYT?

OrKoN: I think that works too!

scripts/generate-cli.ts
nroscino: if I remove these, it generates also the `get_tab_id` in the `chrome-devtools-cli-options.ts`

OrKoN: does it mean get_tab_id is available on the server we fetch tools from?

nroscino: Yes. the tool is present but if it's disabled it returns an error:
```
        if (disabledReason) {
          return {
            content: [
              {
                type: 'text',
                text: disabledReason,
              },
            ],
            isError: true,
          };
        }
```

This is only when --viaCli is passed. When we are not passing `--viaCli` the previous behaviour is preserved (removing the tool completely)
If we don't want these tool at all appearing then we have also to filter them out from the `generate-cli`. Unless we want to change the logic. I am open to change it

src/telemetry/flag_usage_metrics.json
```diff
@@ -192,11 +192,13 @@
   },
   {
     "name": "category_in_page_tools",
-    "flagType": "boolean"
+    "flagType": "boolean",
+    "isDeprecated": true
```
nroscino, line 196: Here do we want to deprecate the old category or we want to delete it completely? @OrKoN

OrKoN: Nothing should be deleted from these generated files, so deprecation is correct here.

OrKoN (approved):
LGTM Thanks!

OrKoN:
Looks like this PR needs a rebase

OrKoN:
@nroscino please link the issue it is closing

PR #1963, opened 2026-04-27
chore(memory): sort all by retainer size

Lightning00Blade:
This is a better measurement as if object self size is X and the retainerSize Y. X will always be smaller then Y, there Y will be the actual size on the Heap.

### DIFF
--- a/src/formatters/HeapSnapshotFormatter.ts
+++ b/src/formatters/HeapSnapshotFormatter.ts
@@ -16,7 +16,7 @@ export interface FormattedSnapshotEntry {
   retainedSize: number;
 }
 
-function isNodeLike(
+export function isNodeLike(
   item: unknown,
 ): item is DevTools.HeapSnapshotModel.HeapSnapshotModel.Node {
   return (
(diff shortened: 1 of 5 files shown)


tests/third_party_notices.test.js.snapshot
```diff
@@ -1,31 +1,4 @@
 exports[`THIRD_PARTY_NOTICES > matches snapshot if exists 1`] = `
-Name: urlpattern-polyfill
```
OrKoN, line 2: I think you need to revert this change

src/formatters/HeapSnapshotFormatter.ts
```diff
@@ -55,7 +55,7 @@ export class HeapSnapshotFormatter {
   }
 
   #getSortedAggregates(): AggregatedInfoWithUid[] {
-    return Object.values(this.#aggregates).sort((a, b) => b.self - a.self);
+    return Object.values(this.#aggregates).sort((a, b) => b.maxRet - a.maxRet);
```
OrKoN, line 58: can we re-use any code from the dtf to handle the various sorting and lookups?

Lightning00Blade: I will double check the look ups, and if they can be re-used will do a follow-up.
For the Sorting it's all done on the Frontend part, as it relies on the DataGrid component handling, I don't think it's worth extracting.

OrKoN (approved):

PR #1967, opened 2026-04-28
fix: replace stale chrome-devtools refs in tests and scripts

triuzzi:
## Summary

Cleans up stale `chrome-devtools` references that were missed during the Brave port and were causing test failures.

- `tests/utils.ts`: `CLI_PATH` now points at the actual `build/src/bin/brave-devtools.js` binary
- `tests/check-for-updates.test.ts`: env var renamed from `CHROME_DEVTOOLS_MCP_NO_UPDATE_CHECKS` to `BRAVE_DEVTOOLS_MCP_NO_UPDATE_CHECKS` to match the implementation
- `tests/tools/{screencast,pagesNavigateAllowlist}.test.ts`: type-only `ParsedArguments` import points at `brave-devtools-mcp-cli-options.js` (came in via the upstream sync but kept the chrome path)
- `scripts/{generate-docs,generate-cli,eval_gemini}.ts`: spawn `build/src/bin/brave-devtools-mcp.js` instead of the chrome binary so dev/eval tooling actually starts the server
- `tests/third_party_notices.test.js.snapshot`: regenerated after recent upstream dependency bumps

## Test plan

- [x] `npm run typecheck` passes
- [x] `npm run build` passes
- [x] `npm test` passes — full suite green on a Mac with Brave Browser installed (after clearing stale `/tmp/brave-devtools-mcp-*` state from previous runs)


triuzzi:
Wrong repo — closing.

PR #1969, opened 2026-04-28
fix(cli): re-generate cli correctly

OrKoN:
Regressed in https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1120


nroscino (approved):
LGTM

PR #1971, opened 2026-04-28
fix: Make fill_form more appealing when filling forms with checkboxes

### LINKED ISSUE #1942: `fill_form` tool is not used to fill multiple form fields at once
### Description of the bug

In evals, out of 9 runs on form filling eval, there was no attempt at using `fill_form` tool. Moreover, 5 of those trajectories had tool call failures related to stale UIDs and subsequent extra turns to fix those issues (by performing extra snapshots). This causes significant extra token burn.

### Reproduction

Ask the agent to fill multiple form fields.

### Expectation

The agent should use `fill-form` tool to fill all fields in one tool call to save tokens.

### MCP configuration

_No response_

v0.21.0

### Chrome version

_No response_

### Coding agent version

gemini-cli

### Model version

gemini-2.5-flash

### Chat log

_No response_

### Node version

_No response_

### Operating system

None

### Extra checklist

zyzyzyryxy:
Fixes #1942

Verified using `npm run eval -- scripts/eval_scenarios/fill_select_and_checkboxes_test.ts`

Without this change, I observed 7 runs using fill_form for all controls at once, 14 runs using click to select checkboxes and 10 runs that did nothing (total 31 runs)

After this change: 9 fill_form using runs (passes), 1 click based approach and 10 no-attempt fails (20 runs total)

Depending how we count the no-attempt runs, its either increase from 23% to 45% or 33% to 90% in eval pass rate.

### DIFF
--- a/src/tools/input.ts
+++ b/src/tools/input.ts
@@ -260,11 +260,29 @@ async function fillFormElement(
     if (aXNode && aXNode.role === 'combobox' && hasOptionChildren(aXNode)) {
       await selectOption(handle, aXNode, value);
     } else {
-      // Increase timeout for longer input values.
-      const timeoutPerChar = 10; // ms
-      const fillTimeout =
-        page.pptrPage.getDefaultTimeout() + value.length * timeoutPerChar;
-      await handle.asLocator().setTimeout(fillTimeout).fill(value);
+      const isToggle = await handle.evaluate(el => {
+        if (el instanceof HTMLInputElement) {
+          return el.type === 'checkbox' || el.type === 'radio';
+        }
+        const role = el.getAttribute('role');
+        return role === 'checkbox' || role === 'radio' || role === 'switch';
+      });
+
+      if (isToggle) {
+        if (['true', 'false'].includes(value)) {
+          await handle.asLocator().fill(value === 'true');
+        } else {
+          throw new Error(
+            `Checkboxes, radio boxes and toggles require "true" or "false" value, but ${value} was used`,
+          );
+        }
+      } else {
+        // Increase timeout for longer input values.
+        const timeoutPerChar = 10; // ms
+        const fillTimeout =
+          page.pptrPage.getDefaultTimeout() + value.length * timeoutPerChar;
+        await handle.asLocator().setTimeout(fillTimeout).fill(value);
+      }
     }
   } catch (error) {
     handleActionError(error, uid);
@@ -286,7 +304,11 @@ export const fill = definePageTool({
       .describe(
         'The uid of an element on the page from the page content snapshot',
       ),
-    value: zod.string().describe('The value to fill in'),
+    value: zod
+      .string()
+      .describe(
+        'The value to fill in. "true" or "false" for checkboxes and toggles, "true" for radio buttons.',
+      ),
     includeSnapshot: includeSnapshotSchema,
   },
   blockedByDialog: true,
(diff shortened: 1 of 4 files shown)


src/tools/input.ts
```diff
@@ -313,7 +313,7 @@ export const drag = definePageTool({
 
 export const fillForm = definePageTool({
   name: 'fill_form',
-  description: `Fill out multiple form elements at once`,
+  description: `Fill out multiple form elements at once. To set checkboxes fill them with 'true'`,
```
OrKoN: let's add a test to tests/tools/input.test.ts that tests checkboxes and verify that it actually accepts a string `true` and `false`

zyzyzyryxy: Good catch, I misread some eval results thinking it already works this way. Pushed an update to make fill and fill_form handle those, together with tests.

src/tools/input.ts
```diff
@@ -204,11 +204,33 @@ async function fillFormElement(
...
+        const isChecked = await handle.evaluate(el => {
+          if (el instanceof HTMLInputElement) {
+            return el.checked;
+          }
+          return el.getAttribute('aria-checked') === 'true';
+        });
+
+        if (isChecked !== shouldBeChecked) {
+          await handle.asLocator().click();
+        }
```
OrKoN, line 278: this should be fixed in puppeteer's fill()

zyzyzyryxy: https://github.com/puppeteer/puppeteer/issues/14938

zyzyzyryxy: Updated to new puppeteer with Locator.fill now handling boolean inputs.

src/tools/input.ts
```diff
@@ -383,7 +387,11 @@ export const fillForm = definePageTool({
         // eslint-disable-next-line @local/enforce-zod-schema
         zod.object({
           uid: zod.string().describe('The uid of the element to fill out'),
-          value: zod.string().describe('Value for the element'),
+          value: zod
+            .string()
+            .describe(
+              'Value for the element. "true" or "false" for checkboxes and toggles, "true" for radio buttons.',
```
OrKoN, line 411: so where do we convert true/false to booleans for puppeteer?

OrKoN (approved):

zyzyzyryxy:
With `Locator.fill` now handling boolean inputs too, I tweaked the responses and added more tests.

Basing on updated eval in #2011 and gemini-3, success rate on that eval rises from 40% to 100% in local tests.

src/tools/input.ts
zyzyzyryxy: Must've accidentally dropped that while splitting two PRs, will fix in a moment.

zyzyzyryxy: Should be fixed now, PTAL

OrKoN (approved):

OrKoN:
> With `Locator.fill` now handling boolean inputs too, I tweaked the responses and added more tests.
> 
> Basing on updated eval in #2011 and gemini-3, success rate on that eval rises from 40% to 100% in local tests.

this is very nice!

zyzyzyryxy:
> > With `Locator.fill` now handling boolean inputs too, I tweaked the responses and added more tests.
> > Basing on updated eval in #2011 and gemini-3, success rate on that eval rises from 40% to 100% in local tests.
> 
> this is very nice!

Keeping my fingers crossed this stat will hold for larger samples than just 10 runs, but looks promising 😄

PR #1978, opened 2026-04-29
fix: add proactive tool rejection when dialog is open

samiyac:
Adding proactive rejection of tool execution for tools which currently get blocked due to open dialogs (related to #1069 )

### DIFF
--- a/src/McpPage.ts
+++ b/src/McpPage.ts
@@ -81,6 +81,14 @@ export class McpPage implements ContextPage {
     this.#dialog = undefined;
   }
 
+  throwIfDialogOpen(): void {
+    if (this.#dialog) {
+      throw new Error(
+        `A dialog is open (${this.#dialog.type()}: ${this.#dialog.message()}).`,
+      );
+    }
+  }
+
   getInPageTools(): ToolGroup<ToolDefinition> | undefined {
     return this.inPageTools;
   }
--- a/src/index.ts
+++ b/src/index.ts
@@ -233,14 +233,17 @@ export async function createMcpServer(
 
           response.setRedactNetworkHeaders(serverArgs.redactNetworkHeaders);
           try {
+            const page =
+              serverArgs.experimentalPageIdRouting &&
+              params.pageId &&
+              !serverArgs.slim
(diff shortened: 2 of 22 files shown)


src/tools/emulation.ts
```diff
@@ -68,6 +68,12 @@ export const emulate = definePageTool({
   },
   handler: async (request, _response, context) => {
     const page = request.page;
+    const dialog = page.getDialog();
+    if (dialog) {
+      throw new Error(
+        `A dialog is open (${dialog.type()}: ${dialog.message()}).`,
+      );
```
OrKoN: let's throw it from a method on the page? `page.throwIfDialogOpen()` to avoid ifs and message duplication in each tool?

samiyac: Updated, thanks for the suggestion.

src/McpPage.ts
```diff
@@ -81,6 +81,14 @@ export class McpPage implements ContextPage {
     this.#dialog = undefined;
   }
 
+  throwIfDialogOpen(): void {
+    if (this.#dialog) {
+      throw new Error(
+        `A dialog is open (${this.#dialog.type()}: ${this.#dialog.message()}).`,
```
OrKoN, line 87: let's include instruction to use handle_dialog to handle. We should re-use the message in the McpResponse perhaps?

samiyac: Updated the implementation as per our discussion. Added 2 tests in index.test.ts to check behavior in both cases

src/index.ts
```diff
@@ -233,13 +233,17 @@ export async function createMcpServer(
 
           response.setRedactNetworkHeaders(serverArgs.redactNetworkHeaders);
           try {
+            const page =
+              serverArgs.experimentalPageIdRouting &&
+              params.pageId &&
+              !serverArgs.slim
+                ? context.getPageById(params.pageId)
+                : context.getSelectedMcpPage();
+            if (tool.blockedByDialog) {
```
OrKoN, line 243: let's move response.setPage(page); outside of this if and remove it on line 247

OrKoN (approved):

PR #1981, opened 2026-04-29
feat: add Claude plugin executablePath user config

maruen:
## Summary

- Add a Claude plugin `executablePath` user config option so users can point the MCP at a custom Chrome binary (e.g. Chrome for Testing) without editing the plugin manifest in their plugin cache.
- Pass the option through with `--executablePath=${user_config.executablePath}`. A blank value keeps the default auto-detected-stable-Chrome behavior.
- Document the Claude Code `pluginConfigs` settings shape, mirroring #1851.

## Motivation

On macOS, when the user's regular Chrome and the MCP-launched Chrome share bundle ID `com.google.Chrome`, AppleScript `tell application "Google Chrome"` becomes ambiguous and OS-automation menu commands (e.g. *Tab → Pin Tab*) silently no-op on the wrong window. Pointing the MCP at `/Applications/Google Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing` (bundle ID `com.google.chrome.for.testing`) gives the MCP-launched Chrome a separate AppleScript identity and resolves the conflict — but currently requires patching the plugin manifest in the user's plugin cache, which gets reset on plugin upgrade.

## Testing

- `jq . .claude-plugin/plugin.json` parses cleanly
- `npm run check-format` (eslint + prettier) passes

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -2,11 +2,18 @@
   "name": "chrome-devtools-mcp",
   "version": "0.23.0",
   "description": "Reliable automation, in-depth debugging, and performance analysis in Chrome using Chrome DevTools and Puppeteer",
+  "userConfig": {
+    "executablePath": {
+      "description": "Optional path to a custom Chrome executable. Useful when the system has multiple Chrome installations (e.g. point at the Chrome for Testing binary so the MCP-launched Chrome stays separate from the user's regular Chrome and can be scripted unambiguously by AppleScript / OS automation tools). Leave blank to use the auto-detected stable Chrome.",
+      "sensitive": false
+    }
+  },
   "mcpServers": {
     "chrome-devtools": {
       "command": "npx",
       "args": [
-        "chrome-devtools-mcp@latest"
+        "chrome-devtools-mcp@latest",
+        "--executablePath=${user_config.executablePath}"
       ]
     }
   }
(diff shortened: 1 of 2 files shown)


.claude-plugin/plugin.json
```diff
@@ -2,11 +2,18 @@
   "name": "chrome-devtools-mcp",
   "version": "0.23.0",
   "description": "Reliable automation, in-depth debugging, and performance analysis in Chrome using Chrome DevTools and Puppeteer",
+  "userConfig": {
+    "executablePath": {
+      "description": "Optional path to a custom Chrome executable. Useful when the system has multiple Chrome installations (e.g. point at the Chrome for Testing binary so the MCP-launched Chrome stays separate from the user's regular Chrome and can be scripted unambiguously by AppleScript / OS automation tools). Leave blank to use the auto-detected stable Chrome.",
+      "sensitive": false
+    }
+  },
```
OrKoN, line 10: can we update it as part of `npm run gen`? so that it does not need to be manually maintained? perhaps a new script to update plugin configurations would be great.

cc @natorion

natorion:
I think adding this makes sense. Any other client integrations we have to update?

OrKoN:
> I think adding this makes sense. Any other client integrations we have to update?

AFAIK not all clients support something like this in their plugins. So it would have to be done per client.

OrKoN:
Closing in favour of https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1851

PR #1984, opened 2026-04-30
fix: always allow tmpdir access with client roots

OrKoN:
This PR changes the implementation to allow access to tmpdir even if the client is not configured it explicitly because several tools default to the tmpdir for outputs. Many clients like gemini-cli/claude do not configure the tmp dir as a root by default.

### DIFF
--- a/tests/roots.test.ts
+++ b/tests/roots.test.ts
@@ -1,0 +1,49 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'node:assert';
+import os from 'node:os';
+import path from 'node:path';
+import {describe, it} from 'node:test';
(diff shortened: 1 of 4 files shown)


tests/roots.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+      const tmpPath = path.join(os.tmpdir(), 'test-file.txt');
+      // This should not throw
+      context.validatePath(tmpPath);
+
+      // Other root should also be allowed
+      context.validatePath(path.join(otherRoot, 'file.txt'));
+
+      // Outside should still be denied. Use a path that is definitely not a root or temp dir.
+      // We use a sibling directory to the temp dir or home dir.
```
wolfib: The comment's first sentence is already enough imo. And the used directory doesn't look like a sibling to me.

wolfib (approved):

PR #1991, opened 2026-05-04
docs: update generate-docs.ts tools output order

### LINKED ISSUE #1932: Update doc generator to list of all available tools in the tool reference
Tools within a category that do not require any args should go first followed by tools that require arguments alongside with the required args.

Tool categories where the entire category requires a flag should be listed after the default categories.

Exception: `--slim` tool mode should remain in a separate reference file as it completely changes the nature of the tools exposed.

nroscino:
Closes: #1932

### DIFF
--- a/scripts/generate-docs.ts
+++ b/scripts/generate-docs.ts
@@ -132,10 +132,8 @@ function addCrossLinks(text: string, tools: ToolWithAnnotations[]): string {
 }
 
 function sortTools(a: ToolWithAnnotations, b: ToolWithAnnotations): number {
-  const aHasConditions =
-    (a.annotations?.conditions && a.annotations.conditions.length > 0) || false;
-  const bHasConditions =
-    (b.annotations?.conditions && b.annotations.conditions.length > 0) || false;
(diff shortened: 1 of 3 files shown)


OrKoN (changes requested):
As discussed offline, the issue is about CLI arguments that are needed to enable a specific tool or category. We should not sort based on tool args.

scripts/generate-docs.ts
```diff
@@ -131,6 +131,22 @@ function addCrossLinks(text: string, tools: ToolWithAnnotations[]): string {
   return result;
 }
 
+function sortTools(a: ToolWithAnnotations, b: ToolWithAnnotations): number {
+  const aHasConditions =
+    (a.annotations?.conditions && a.annotations.conditions.length > 0) || false;
```
OrKoN:
```suggestion
  const aHasConditions =  Boolean(a.annotations?.conditions.length > 0);
```

OrKoN (approved):
LGTM thanks!

PR #1994, opened 2026-05-05
fix: make sure env variables are consistently applied when parsing args

OrKoN:
This PR moves the checks for CHROME_DEVTOOLS_MCP_NO_USAGE_STATISTICS and CI env vars to parseArguments making sure it is correctly applied by the daemon and in tests. It also updates the tests to set explicit CHROME_DEVTOOLS_MCP_NO_USAGE_STATISTICS as it might not always be inherited. This change is likely to affect the metrics collected and might explain some fluctuations we observed.


OrKoN:
cc @yulunz @natorion

wolfib (approved):

PR #1999, opened 2026-05-06
chore: make Clearcut logger a global singleton

yulunz:
We will add a error logging method to ClearcutLogger in a follow-up PR. Since the error can happen anywhere in the stack, the logger instance has to be readily available (i.e. w/o passing the logger instance everywhere). This commit registers the instantiated logger as a global singleton, and makes it possible to retrieve it by a static method (`ClearcutLogger.get()`) wherever we need it.

### DIFF
--- a/src/telemetry/ClearcutLogger.ts
+++ b/src/telemetry/ClearcutLogger.ts
@@ -171,20 +171,41 @@ function detectOsType(): OsType {
   }
 }
 
+export interface ClearcutLoggerOptions {
+  appVersion: string;
+  logFile?: string;
+  persistence?: Persistence;
+  watchdogClient?: WatchdogClient;
+  clearcutEndpoint?: string;
+  clearcutForceFlushIntervalMs?: number;
+  clearcutIncludePidHeader?: boolean;
+}
+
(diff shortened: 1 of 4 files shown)


src/telemetry/ClearcutLogger.ts
```diff
@@ -171,20 +171,42 @@ function detectOsType(): OsType {
...
-    appVersion: string;
-    logFile?: string;
-    persistence?: Persistence;
-    watchdogClient?: WatchdogClient;
-    clearcutEndpoint?: string;
-    clearcutForceFlushIntervalMs?: number;
-    clearcutIncludePidHeader?: boolean;
-  }) {
+  static initialize(options: ClearcutLoggerOptions): ClearcutLogger {
+    const symbolKey = Symbol.for('ClearcutLogger.instance');
```
OrKoN: could we use a module level variable instead of keying it by a symbol on globalThis to avoid type casts?

yulunz: yeah, that'd be better. and done!

OrKoN (approved):

PR #1982, opened 2026-04-29
feat: support third-party developer tools

wolfib:
Enables "third-party developer tools" feature. This allows the inspected web page to expose tools which provide debugging information to Chrome DevTools for Agents.

Third-party developer tools enable web applications to expose internal state, component hierarchies, or specific debug data that cannot be deduced through static analysis. This allows Chrome DevTools for Agents to provide richer, more actionable context to AI agents during debugging sessions.

2 additional tools are enabled in Chrome DevTools for Agents for interacting with third-party developer tools: `list_3p_developer_tools()` and `execute_3p_developer_tool`.

Code changes in this PR:
- Rename "in-page tools" to "third-party developer tools"
- Unhide
- Make available in CLI
- Add documentation

PR #2005, opened 2026-05-07
fix(telemetry): re-run the update metrics script

yulunz:
the updates made to these files from #1982 came out a bit weird. So I reverted changes on these files and re-run `npm run gen` (which includes `npm run update-metrics`) to correct them.

we'd always want old entries to be marked as deprecated and new entries to be added to the very last. Don't manually resolve conflicts in these files. When there are conflicts, just revert back to a clean state and `npm run gen` will take care of them 😉

### DIFF
--- a/src/telemetry/flag_usage_metrics.json
+++ b/src/telemetry/flag_usage_metrics.json
@@ -281,19 +281,19 @@
     "isDeprecated": true
   },
   {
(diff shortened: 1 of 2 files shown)


wolfib (approved):
Thanks for the fix and the explanation!
I did indeed manually edit the files in my PR.

PR #2009, opened 2026-05-07
fix: protocolTimeout for heavy pages + attach-mode arg cleanup

triuzzi:
## Summary

Two reliability fixes for real-world failures hit during a Brightcove Studio dev session:

### 1. `protocolTimeout` raised from puppeteer's 180s default to 10min

Heavy pages (e.g. Studio module dev bundles ~160MB JS) cannot ack `Network.enable` and other auto-attached CDP domain calls within 180s. Once the timeout fires, puppeteer marks the CDP connection dead and every subsequent call throws `Network.enable timed out`. The daemon shows as alive (`process.kill(pid, 0)` returns true), so client retries reuse the same wedged connection until manually killed.

Bumped to 600s on both `puppeteer.connect()` and `puppeteer.launch()`. Env-overridable via `BRAVE_DEVTOOLS_PROTOCOL_TIMEOUT_MS`.

### 2. Attach-mode `start` no longer emits spawn-only flags

`brave-devtools start --browserUrl http://127.0.0.1:9222` was producing daemon args:

```
--browser-url http://127.0.0.1:9222 --headless --isolated ...
```

The runtime resolver ignores `--headless` / `--isolated` under attach mode, so behaviour was correct — but `status` showed misleading args and the silent ignore would become a real bug if the resolver order ever changed. Root cause: spawn-mode defaults (`headless: true`, `isolated: true`) applied unconditionally after yargs parsing, with `headless` having `default: true` overridden in the CLI bin file (line 61).

Detect attach mode (`browserUrl` / `wsEndpoint` / `autoConnect`) and strip spawn-only flags before `serializeArgs`.

## Smoke test

```
$ brave-devtools start --browserUrl http://127.0.0.1:9222 && brave-devtools status | grep args=
args=["--browser-url","http://127.0.0.1:9222","--category-emulation",...]   # no --headless / --isolated

$ brave-devtools start --headless && brave-devtools status | grep args=
args=["--headless","--isolated","--category-emulation",...]   # defaults still apply in spawn mode
```

## Out of scope (separate PRs)

These came up in the same audit but are bigger surgery:

- **Per-tty daemon cleanup** — multiple stale daemons accumulate (10× observed on this machine). Needs parent-PID watchdog in `daemon.ts`.
- **Hang recovery / health probe** — once the daemon's MCP child wedges, no auto-recovery. Needs a fast-probe before reuse in `brave-devtools.ts:230` and a real `status` round-trip.

## Test plan

- [x] `npm run build` green
- [x] Smoke test attach mode (correct args)
- [x] Smoke test spawn mode (defaults still apply)
- [ ] CI green
- [ ] Verify on a real heavy-page session that `Network.enable timed out` no longer surfaces


triuzzi:
Closing — head branch was based on the brave-devtools-mcp fork's main and pulled in 9 fork-specific commits (Brave port, version bumps, README rebrand). Re-opening a clean PR branched from upstream/main with only the protocolTimeout change.

PR #2035, opened 2026-05-11
fix: Patch puppeteer for missing session IDs

wolfib:


src/puppeteer-patches.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+import './third_party/index.js';
+
+import {
+  InternalConnection as Connection,
+  type InternalCallbackRegistry as CallbackRegistry,
+} from './third_party/index.js';
+
+// 1. Map active command IDs directly to their sessionId
```
OrKoN, line 14: is it a reason for the lighthouse timeouts? sounds like it would be easier to fix in Puppeter?

OrKoN:
@wolfib can you confirm that the issue is fixed?

wolfib:
The issue isn't fully fixed yet, I'm currently on it.

PR #2037, opened 2026-05-11
feat: support allowedUrlPattern & blockedUrlPattern Options

### LINKED ISSUE #239: Feature Request: Support Allowed Origins & Blocked Origins list options
**Is your feature request related to a problem? Please describe.**

When adopting browser automation MCPs in a corporate environment, considering security is especially crucial. Any possibility of automatic/unintentional access to unknown/malicious content can be a reason for preventing adoption. 
Playwright MCP, for example, has `--allowed-origins` & `--blocked-origins` to filter in/out which contents the MCP is allowed to access. This gives us more certainty on which content the tool will access, and this can bring a good reason to use the browser automation tool corporate-wide.
For example, we can restrict access except to localhost.

**Describe the solution you'd like**
Like Playwright MCP, providing CLI options to filter origins the MCP server can access (e.g. localhost only) prevents unintended navigation to unauthorized or potentially malicious sites while maintaining safe internal tool usage.

Playwright uses [`context.route()`](https://github.com/microsoft/playwright/blob/main/packages/playwright/src/mcp/browser/context.ts#L194-L206), which internally relies on the same CDP methods in the identical way as Pupeteer [` page.setRequestInterception()`](https://github.com/puppeteer/puppeteer/blob/48902651782c712a90cba51e913f1adf78933766/packages/puppeteer-core/src/cdp/Page.ts#L530) does.

- https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/chromium/crNetworkManager.ts#L155
-  https://github.com/puppeteer/puppeteer/blob/48902651782c712a90cba51e913f1adf78933766/packages/puppeteer-core/src/cdp/NetworkManager.ts#L312

**Additional context**
- [microsoft/playwright-mcp: Playwright MCP server](https://github.com/microsoft/playwright-mcp?tab=readme-ov-file#configuration)

nattallius:
## Support for Network Blocklists and Allowlists (`--blocked-url-pattern` & `--allowed-url-pattern` arguments)

This PR adds support for CLI options to restrict network access in the browser session via URL patterns.

### Key Features & How It Works

- **Pattern Matching:** Utilizes the [URLPattern Standard](https://urlpattern.spec.whatwg.org/) for pattern matching.
- **Target Detachment:** Silently detaches from targets (pages/tabs) whose URLs match blocked patterns (or do not match allowed patterns) upon connection.
- **Runtime Blocking:** Prevents navigations and blocks runtime requests (such as fetch/XHR and subresources) if they violate the pattern rules.
- **Mutual Exclusivity:** `--blocked-url-pattern` and `--allowed-url-pattern` conflict with each other and cannot be configured simultaneously.
- **Browser Requirements:**
  - **`--allowed-url-pattern`**: Requires **Chrome 149+**.
  - **`--blocked-url-pattern`**: Works on Chrome versions older than 149, but **Chrome 149+ is highly recommended**. 

### Important Limitations & Side Effects

- **Network Emulation/Throttling Conflict:** Network throttling is disabled when a network blocklist/allowlist is configured, to avoid conflicting with Puppeteer's underlying blocking mechanisms. 
  - Using the `emulate` tool to modify `networkConditions` (e.g. setting to `Offline`) will throw an error: *`Network throttling is not supported when network blocking (allowlist/blocklist) is configured.`*
  - Other emulation settings (e.g., `cpuThrottlingRate`, `geolocation`, `viewport`) are unaffected and remain fully functional.
- **Service Worker Request Bypass:** Service worker-initiated fetches bypass these restrictions and are not blocked.

---

#### 1. Blocking specific domains or endpoints (Blocklist)
Add the `--blocked-url-pattern` options to the `args` list in your MCP settings file:

```json
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": [
        "chrome-devtools-mcp@latest",
        "--blocked-url-pattern=*://*.blocked-example.com/*",
        "--blocked-url-pattern=*://*.another-blocked-example.com/*"
      ]
    }
  }
}
```

#### 2. Restricting access to authorized domains (Allowlist)
Add the `--allowed-url-pattern` options to restrict the browser to permitted hosts (requires Chrome 149+):

```json
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": [
        "chrome-devtools-mcp@latest",
        "--allowed-url-pattern=https://*.allowed-example.com/*",
        "--allowed-url-pattern=https://*.another-allowed-example.com/*"
      ]
    }
  }
}
```

### DIFF
--- a/scripts/test.mjs
+++ b/scripts/test.mjs
@@ -103,7 +103,6 @@ async function runTests(attempt) {
   });
 }
 
-
 const maxAttempts = shouldRetry ? 3 : 1;
 let exitCode = 1;
 
--- a/scripts/test.mjs
+++ b/scripts/test.mjs
@@ -103,6 +103,9 @@ async function runTests(attempt) {
   });
 }
 
+const chromePath = _installChrome('149.0.7827.14');
+process.env.CHROME_M149_EXECUTABLE_PATH = chromePath;
+
 const maxAttempts = shouldRetry ? 3 : 1;
 let exitCode = 1;
 
--- a/scripts/test.mjs
+++ b/scripts/test.mjs
@@ -103,8 +103,6 @@ async function runTests(attempt) {
   });
 }
 
-const chromePath = _installChrome('149.0.7827.14');
-process.env.CHROME_M149_EXECUTABLE_PATH = chromePath;
 
 const maxAttempts = shouldRetry ? 3 : 1;
 let exitCode = 1;
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -60,6 +60,8 @@ interface McpContextOptions {
   experimentalIncludeAllPages?: boolean;
   // Whether CrUX data should be fetched.
   performanceCrux: boolean;
+  // Whether allowlist/blocklist is configured.
+  hasNetworkBlockOrAllowlist?: boolean;
 }
 
 const DEFAULT_TIMEOUT = 5_000;
@@ -345,24 +347,27 @@ export class McpContext implements Context {
     const mcpPage = this.#getMcpPage(page);
     const newSettings: EmulationSettings = {...mcpPage.emulationSettings};
 
-    if (!options.networkConditions) {
-      await page.emulateNetworkConditions(null);
-      delete newSettings.networkConditions;
-    } else if (options.networkConditions === 'Offline') {
-      await page.emulateNetworkConditions({
-        offline: true,
-        download: 0,
-        upload: 0,
-        latency: 0,
-      });
-      newSettings.networkConditions = 'Offline';
-    } else if (options.networkConditions in PredefinedNetworkConditions) {
-      const networkCondition =
-        PredefinedNetworkConditions[
-          options.networkConditions as keyof typeof PredefinedNetworkConditions
-        ];
-      await page.emulateNetworkConditions(networkCondition);
-      newSettings.networkConditions = options.networkConditions;
+    // Skip network emulation if blocklist/allowlist is configured, as it is rejected by Puppeteer.
+    if (!this.#options.hasNetworkBlockOrAllowlist) {
+      if (!options.networkConditions) {
+        await page.emulateNetworkConditions(null);
+        delete newSettings.networkConditions;
+      } else if (options.networkConditions === 'Offline') {
+        await page.emulateNetworkConditions({
+          offline: true,
+          download: 0,
+          upload: 0,
+          latency: 0,
+        });
+        newSettings.networkConditions = 'Offline';
+      } else if (options.networkConditions in PredefinedNetworkConditions) {
+        const networkCondition =
+          PredefinedNetworkConditions[
+            options.networkConditions as keyof typeof PredefinedNetworkConditions
+          ];
+        await page.emulateNetworkConditions(networkCondition);
+        newSettings.networkConditions = options.networkConditions;
+      }
     }
 
     const secondarySession = this.getDevToolsUniverse(mcpPage)?.session;
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -347,27 +347,31 @@ export class McpContext implements Context {
     const mcpPage = this.#getMcpPage(page);
     const newSettings: EmulationSettings = {...mcpPage.emulationSettings};
 
-    // Skip network emulation if blocklist/allowlist is configured, as it is rejected by Puppeteer.
-    if (!this.#options.hasNetworkBlockOrAllowlist) {
-      if (!options.networkConditions) {
-        await page.emulateNetworkConditions(null);
-        delete newSettings.networkConditions;
-      } else if (options.networkConditions === 'Offline') {
-        await page.emulateNetworkConditions({
-          offline: true,
-          download: 0,
-          upload: 0,
-          latency: 0,
-        });
-        newSettings.networkConditions = 'Offline';
-      } else if (options.networkConditions in PredefinedNetworkConditions) {
-        const networkCondition =
-          PredefinedNetworkConditions[
-            options.networkConditions as keyof typeof PredefinedNetworkConditions
-          ];
-        await page.emulateNetworkConditions(networkCondition);
-        newSettings.networkConditions = options.networkConditions;
+    // Skip network emulation if blocklist/allowlist is configured, as it conflicts with blocking rules in Puppeteer.
+    if (this.#options.hasNetworkBlockOrAllowlist) {
+      if (options.networkConditions !== undefined) {
+        throw new Error(
+          'Network throttling is not supported when network blocking (allowlist/blocklist) is configured.',
+        );
       }
+    } else if (!options.networkConditions) {
+      await page.emulateNetworkConditions(null);
+      delete newSettings.networkConditions;
+    } else if (options.networkConditions === 'Offline') {
+      await page.emulateNetworkConditions({
+        offline: true,
+        download: 0,
+        upload: 0,
+        latency: 0,
+      });
+      newSettings.networkConditions = 'Offline';
+    } else if (options.networkConditions in PredefinedNetworkConditions) {
+      const networkCondition =
+        PredefinedNetworkConditions[
+          options.networkConditions as keyof typeof PredefinedNetworkConditions
+        ];
+      await page.emulateNetworkConditions(networkCondition);
+      newSettings.networkConditions = options.networkConditions;
     }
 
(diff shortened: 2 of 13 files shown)


src/bin/chrome-devtools-mcp-cli-options.ts
```diff
@@ -210,6 +210,17 @@ export const cliOptions = {
     describe:
       'Additional arguments for Chrome. Only applies when Chrome is launched by chrome-devtools-mcp.',
   },
+  blocklist: {
+    type: 'array',
+    describe:
+      'URL patterns to block access to. Uses standard URLPattern API. Cannot be used with --blocklist',
```
OrKoN: Please document CLI args in details, what will be blocked and what will not be blocked and starting from which Chrome version.

src/bin/chrome-devtools-mcp-cli-options.ts
```diff
@@ -210,6 +210,17 @@ export const cliOptions = {
...
+  },
+  allowlist: {
+    type: 'array',
+    describe:
+      'URL patterns to allow access to (blocks everything else). Uses standard URLPattern API. Cannot be used with --blocklist.',
```
OrKoN: We do not need to mention `Cannot be used with` as 'conflicts' already would handle this.

src/bin/chrome-devtools-mcp-cli-options.ts
```diff
@@ -210,6 +210,17 @@ export const cliOptions = {
     describe:
       'Additional arguments for Chrome. Only applies when Chrome is launched by chrome-devtools-mcp.',
   },
+  blocklist: {
+    type: 'array',
```
OrKoN, line 209: how does it compare to using a string type? Does `array` support comma separated input? I think if we go with the array we should rename the argument from blocklist/allowlist to `blocked-url-pattern`, `allowed-url-pattern`.

OrKoN: I see the current version is using comma separation. I wonder if we will run into issues if the pattern itself contains a comma?

tests/security_policies.test.ts
```diff
@@ -0,0 +1,238 @@
...
+import {describe, it} from 'node:test';
+
+import {lighthouseAudit} from '../src/tools/lighthouse.js';
+import {navigatePage} from '../src/tools/pages.js';
+import {evaluateScript} from '../src/tools/script.js';
+
+import {serverHooks} from './server.js';
+import {withMcpContext} from './utils.js';
+
+describe('Security Policies Integration', () => {
```
OrKoN: let's rename this and the file to something more specific like `network_blocking.test.ts`

tests/network_blocking.test.ts
```diff
@@ -0,0 +1,241 @@
...
+        );
+        assert.strictEqual(
+          JSON.parse(response.responseLines.at(2)!),
+          'Failed to fetch',
+          'Fetch should still be blocked after audit',
+        );
+      },
+      {
+        blockedUrlPattern: [server.getRoute('/blocked.html')],
+        executablePath: process.env.CHROME_M149_EXECUTABLE_PATH,
```
OrKoN: block list should not need M149, right?

tests/network_blocking.test.ts
```diff
@@ -0,0 +1,241 @@
...
+        );
+
+        assert.strictEqual(
+          JSON.parse(response.responseLines.at(2)!),
+          'Failed to fetch',
+        );
+      },
+      {
+        blockedUrlPattern: [server.getRoute('/blocked.html')],
+        executablePath: process.env.CHROME_M149_EXECUTABLE_PATH,
```
OrKoN: blocklist should not need M149?

tests/network_blocking.test.ts
nattallius: No,  it also fails if runs against stable

scripts/test.mjs
```diff
@@ -103,6 +103,9 @@ async function runTests(attempt) {
   });
 }
 
+const chromePath = _installChrome('149.0.7827.14');
```
Lightning00Blade: Do we need this this is already stable https://chromestatus.com/roadmap
And that is the version pinned by Puppeteer.

src/McpContext.ts
```diff
@@ -345,24 +347,27 @@ export class McpContext implements Context {
...
-      newSettings.networkConditions = 'Offline';
-    } else if (options.networkConditions in PredefinedNetworkConditions) {
-      const networkCondition =
-        PredefinedNetworkConditions[
-          options.networkConditions as keyof typeof PredefinedNetworkConditions
-        ];
-      await page.emulateNetworkConditions(networkCondition);
-      newSettings.networkConditions = options.networkConditions;
+    // Skip network emulation if blocklist/allowlist is configured, as it is rejected by Puppeteer.
+    if (!this.#options.hasNetworkBlockOrAllowlist) {
```
Lightning00Blade: Should we throw if we try to set emulation? Reason is I try to emulate network conditions and this just silently ignores it, an LLM most likely will be confused.

tests/browser.test.ts
```diff
@@ -100,4 +102,88 @@ describe('browser', () => {
...
+        const content = await page.evaluate(() => document.body.textContent);
+        assert.strictEqual(content, 'Allowed');
+
+        // Fetch of blocked URL from the page
+        const fetchResult = await page.evaluate(async url => {
+          try {
+            await fetch(url);
+            return 'SUCCESS';
+          } catch (err) {
+            return err instanceof Error ? err.message : String(err);
```
Lightning00Blade: Nit: We can just return true or false as the message may change.

tests/browser.test.ts
```diff
@@ -100,4 +102,88 @@ describe('browser', () => {
...
+
+        assert.strictEqual(fetchResult, 'Failed to fetch');
+      } finally {
+        await browser.close();
+      }
+    });
+
+    it(
+      'blocks URLs not in allowlist',
+      {skip: 'Requires Chrome 149 or greater'},
```
Lightning00Blade: Is this also still needed?

tests/cli.test.ts
```diff
@@ -345,4 +345,62 @@ describe('cli args parsing', () => {
...
+    assert.deepStrictEqual(singleArgs.blockedUrlPattern, [
+      'https://example.com/*',
+    ]);
+
+    const repeatedArgs = parseArguments(
+      '1.0.0',
+      [
+        'node',
+        'main.js',
+        '--blocked-url-pattern=https://a.com/*',
```
Lightning00Blade, line 367: Can you double check if ` '--blocked-url-pattern', '<pattern-1>', <pattern-2>` works? I think that is also valid CLI args.

tests/network_blocking.test.ts
```diff
@@ -0,0 +1,240 @@
...
+
+        const blockedUrl = server.getRoute('/blocked.html');
+        response.resetResponseLineForTesting();
+        await evaluateScript().handler(
+          {
+            params: {
+              function: `async () => {
+                try {
+                  await fetch("${blockedUrl}");
+                  return 'SUCCESS';
```
Lightning00Blade, line 63: Same here. we can use true/false.

tests/network_blocking.test.ts
```diff
@@ -0,0 +1,240 @@
...
+          context,
+        );
+
+        assert.strictEqual(
+          JSON.parse(response.responseLines.at(2)!),
+          'Failed to fetch',
+        );
+      },
+      {
+        blockedUrlPattern: [server.getRoute('/blocked.html')],
```
Lightning00Blade: This was confusingly at the end we should move it to the start, some where easier to find.

tests/network_blocking.test.ts
```diff
@@ -0,0 +1,240 @@
...
+        );
+
+        assert.strictEqual(
+          JSON.parse(response.responseLines.at(2)!),
+          'Failed to fetch',
+        );
+      },
+      {
+        allowedUrlPattern: [server.getRoute('/allowed.html')],
+        executablePath: process.env.CHROME_M149_EXECUTABLE_PATH,
```
Lightning00Blade: Do we need this?

Lightning00Blade (approved):
LGTM with some small comment and questions.

tests/browser.test.ts
nattallius: removed

scripts/test.mjs
nattallius: removed

PR #2041, opened 2026-05-12
refactor: waitForResult helpers

Lightning00Blade:


src/McpResponse.ts
```diff
@@ -388,6 +389,12 @@ export class McpResponse implements Response {
     this.#textResponseLines.push(value);
   }
 
+  appendWaitForResult(result: WaitForEventsResult): void {
+    if (result.navigatedToUrl) {
+      this.appendResponseLine(`Page navigated to ${result.navigatedToUrl}.`);
```
OrKoN: nit: can we do it during the format() stage?

src/WaitForHelper.ts
```diff
@@ -16,6 +16,9 @@ export class WaitForHelper {
   #expectNavigationIn: number;
   #navigationTimeout: number;
 
+  #dialogOpened = false;
```
OrKoN, line 19: is each helper instance used only for one wait operation?

Lightning00Blade: Yes it should be and I added line 135 for that.

Note this will break if one uses the tool but not awaits it.

src/McpResponse.ts
```diff
@@ -747,6 +754,14 @@ export class McpResponse implements Response {
       response.push(...this.#textResponseLines);
     }
 
+    if (this.#attachedWaitForResult) {
+      if (this.#attachedWaitForResult.navigatedToUrl) {
+        response.push(
```
OrKoN, line 759: add url to structuredContent as well?

Lightning00Blade: Should we, we already push it to the response text, would that not mean we have it two times?

OrKoN: is the response used in structured content?

Lightning00Blade: Actually not, sorry fixed

OrKoN (approved):
LGTM thanks!

PR #2042, opened 2026-05-12
feat: add experimental TOON support for structured content output

zyzyzyryxy:
Adds `--experimentalToonFormat` boolean flag, controlling how structured content is formatted in text response.
By default, custom shorthand format is used. With this flag, TOON format (see https://github.com/toon-format/toon) is used instead.

TOON format is supposed to be more token-efficient and less error-prone for agents to understand than json, but it's not clear if it will be better on those metrics than the custom format used so far. (Evals pending)

One clear benefit over custom format would be less code to maintain if we decide to fully switch to TOON, due to dropping custom formatters and reusing json formatters used for structuredContent anyways.


Lightning00Blade (approved):
LGTM

PR #463, opened 2025-10-23
feat: support saving snapshots to file

OrKoN:
Refs https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/153

PR #248, opened 2025-10-01
feat: add filePath parameter to take_snapshot and evaluate_script

Coquinate:
## Summary
Adds optional `filePath` parameter to `take_snapshot` and `evaluate_script` tools, enabling direct file output instead of response inclusion.

Resolves #153

## Features
- `take_snapshot` can save accessibility tree to file
- `evaluate_script` can save JSON result to file
- Reduces token usage for large outputs
- Enables offline analysis and processing
- Fully backwards compatible

## Implementation
- Added optional `filePath` parameter to both tool schemas
- Write formatted output using `writeFile` from `node:fs/promises`
- Updated `Context` interface to expose `createTextSnapshot` and `getTextSnapshot`
- Follows same pattern as `take_screenshot` tool

## Testing
- ✅ New test: saves snapshot to file
- ✅ New test: saves script result to file  
- ✅ All existing tests still pass
- ✅ Backwards compatibility verified
- ✅ Manual testing with local MCP server

## Test Plan
- [ ] Build succeeds
- [ ] All existing tests pass
- [ ] New file-saving tests pass
- [ ] Documentation generated correctly

Issue #153, opened 2025-09-25
Feature Request: Add Save to File for take_snapshot and evaluate_script tools


brian-ln:
**Is your feature request related to a problem? Please describe.**
I am unable to save off screenshots using the *chrome-devtools-mcp* server. Sometimes I also get too many token errors on tool results.

**Describe the solution you'd like**
Add a save to file optional parameter to tools that return images  (_take_screenshot_) or large amounts of data  (_take_snapshot_, _evaluate_script_) .

The [mcp-chrome](https://github.com/hangwin/mcp-chrome/) server (and several others) supports saving images as files using the *asPng* optional parameter.

*From the [mcp-chrome](https://github.com/hangwin/mcp-chrome/) server:*
```json
{
    name: TOOL_NAMES.BROWSER.SCREENSHOT,
    description:
      'Take a screenshot of the current page or a specific element(if you want to see the page, recommend to use chrome_get_web_content first)',
    inputSchema: {
      type: 'object',
      properties: {
        name: { type: 'string', description: 'Name for the screenshot, if saving as PNG' },
        selector: { type: 'string', description: 'CSS selector for element to screenshot' },
        width: { type: 'number', description: 'Width in pixels (default: 800)' },
        height: { type: 'number', description: 'Height in pixels (default: 600)' },
        storeBase64: {
          type: 'boolean',
          description:
            'return screenshot in base64 format (default: false) if you want to see the page, recommend set this to be true',
        },
        fullPage: {
          type: 'boolean',
          description: 'Store screenshot of the entire page (default: true)',
        },
        savePng: {
          type: 'boolean',
          description:
            'Save screenshot as PNG file (default: true)，if you want to see the page, recommend set this to be false, and set storeBase64 to be true',
        },
      },
      required: [],
    },
  },
```

**Describe alternatives you've considered**
Claude code can't (or won't) let me save the _[Image]_ data returned from the tool call to _take_screenshot_...

**Additional context**
The [mcp-chrome](https://github.com/hangwin/mcp-chrome/) server (and others) also support returning images as base 64 content using the _storeBase64_ optional parameter.

OrKoN:
I think it makes sense to have this. What we currently do:

1) if the image is less than 2 MB we return it as an image content according to the MCP.
2) if it is larger than 2 MB, we save it to a tmp file (automatically generated filename).

[...]

PR #2054, opened 2026-05-13
feat: support filePath in evaluate_script

masamaru0513:
## Summary

Adds an optional `filePath` parameter to `evaluate_script` that saves the script output to a file instead of returning it inline.

Refs #153

## Motivation

Issue #153 requested `filePath` support for `take_snapshot` and `evaluate_script`. `take_snapshot` was addressed in #463. PR #248 previously attempted this but was closed due to conflicts. This PR implements the same feature on the current codebase, completing the remaining piece.

## Changes

- Add optional `filePath` parameter to the `evaluate_script` schema
- Add `context.validatePath(filePath)` call for path validation
- Pass `{filePath, context}` options to `performEvaluation()`
- In `performEvaluation()`: when `filePath` is provided, save output via `context.saveFile()` with `.json` extension; otherwise return inline as before
- Update `docs/tool-reference.md` via `npm run docs:generate`
- Add unit test for file output

## Key design decisions

- **Same pattern as existing tools**: Follows the `context.saveFile()` pattern established by `take_snapshot` (#463), `take_screenshot`, `get_network_request` (#795), and performance tools (#686).
- **Minimal change surface**: Only `performEvaluation()` gains an optional `options` parameter. No new interfaces or abstractions.
- **Backwards compatible**: `filePath` is optional. When omitted, behavior is identical to before.

## Testing

**Unit test added** (`tests/tools/script.test.ts`):
- Call `evaluate_script` with `filePath` set to a temp file
- Assert response contains "Output saved to"
- Assert file content matches the JSON-serialized return value
- Clean up temp file in `finally` block

**Manual testing performed**:
- `() => document.title` with `filePath: /tmp/test.json` → file contains `"Example Domain"`
- `() => document.title` without `filePath` → inline ```json block returned (no regression)
- `() => Array.from({length: 100}, ...)` with `filePath` → 100-item array saved correctly
- `filePath` pointing to non-existent directory → directory auto-created, file saved
- Relative path (`test.json`) → resolved to CWD, absolute path shown in response
- Function that throws → error returned, no partial file created
- Existing file as `filePath` → file overwritten completely

### DIFF
--- a/src/tools/script.ts
+++ b/src/tools/script.ts
@@ -147,7 +147,7 @@ const performEvaluation = async (
   fnString: string,
   args: Array<JSHandle<unknown>>,
   response: Response,
-  options?: {filePath?: string; context?: Context},
+  options?: {filePath: string; context: Context},
 ) => {
   const fn = await evaluatable.evaluateHandle(`(${fnString})`);
   try {
@@ -159,7 +159,7 @@ const performEvaluation = async (
       fn,
       ...args,
     );
-    if (options?.filePath && options.context) {
+    if (options?.filePath) {
       const data = new TextEncoder().encode(result ?? 'undefined');
       const {filename} = await options.context.saveFile(
         data,
--- a/src/tools/script.ts
+++ b/src/tools/script.ts
@@ -46,6 +46,12 @@ Example with arguments: \`(el) => {
         )
         .optional()
         .describe(`An optional list of arguments to pass to the function.`),
+      filePath: zod
+        .string()
+        .optional()
+        .describe(
+          'The absolute or relative path to a file to save the script output to. If omitted, the output is returned inline.',
+        ),
       dialogAction: zod
         .string()
         .optional()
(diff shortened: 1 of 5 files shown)


src/tools/script.ts
```diff
@@ -132,6 +147,7 @@ const performEvaluation = async (
   fnString: string,
   args: Array<JSHandle<unknown>>,
   response: Response,
+  options?: {filePath?: string; context?: Context},
```
OrKoN: can context be undefined?

masamaru0513: You're right — `context` is always provided when `filePath` is set. Fixed: made both `filePath` and `context` required within the options object (the options object itself remains optional). Pushed in b94a532.

OrKoN (approved):

OrKoN:
LGTM thanks!

masamaru0513:
Thank you for the review!

Issue #2052, opened 2026-05-13
Impossible to start two Chrome instances from DevTools for agents MCP server when using Claude Code CLI


gilbertococchi:
### Description of the bug

It looks like it's not possible to start multiple sessions using DevTools for agents MCP server with Claude Code CLI.

### Reproduction

- Open Claude Code CLI
- Insert prompt asking to test a page on localhost
- Claude Code CLI invocates DevTools for agents MCP server opening Chrome Browser instance.
- In the meantime, open another tab with a different prompt asking to browse another page
- CHrome is locked and the secont tab is not triggering a new Chrome instance.

### Expectation

Claude Code CLI should be able to trigger multiple Chrome Browser instances, only just one per time

### Chrome DevTools MCP version

0.26.0

### Operating system

None

### Extra checklist

- [ ] I want to provide a PR to fix this bug

natorion:
We discussed this when we worked on https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1019. @OrKoN thoughts?

OrKoN:
Yes, if ClaudeCode does not start a new instance and shares the instance between conversations, the `--experimentalPageIdRouting` is the way to make it work for concurrent conversations.

natorion:
We should document this feature in our docs. Seems like we have nothing.

OrKoN:
it is still experimental and we have not done evals. Let's bring to planning if it is important. Although with this bug I am not sure if the expectation is that we should manage multiple Chrome instances with one MCP server instance. We decided against this in https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1019

passtas:
@gilbertococchi, if you would start MCP with `--isolated` flag, you can run as many parallel individual sessions as you'd like even now. Each Claude session will use its own context when spinning up the mcp.

There are basically two features available atm:
- For "subagents" (who shares the tools from parent main agent) but within the same session - you can use pageId routing - so they will reuse the same browser context and parallel work will be managed in separate tabs
- For "Team of Agents"/"Swarm" or just two or more instances of the CLI - use the `isolated` flag - each session will use its own instance of the mcp and browser. 

My default now is:
```
"chrome-devtools": {
            "command": "npx",
            "args": ["-y", "chrome-devtools-mcp@latest", "--experimentalPageIdRouting", "--isolated"]
        }
```

Basically it works everywhere enabling both modes of operation.

[...]

PR #2062, opened 2026-05-15
docs: explain concurrent session options

MukundaKatta:
## Summary
- Add a README section explaining how to configure concurrent MCP sessions
- Document when to use `--experimentalPageIdRouting` for shared server instances
- Point independent sessions at `--isolated` to avoid sharing the default user data directory

Refs #2052

## Testing
- `npm run check-format`

### DIFF
--- a/README.md
+++ b/README.md
@@ -728,6 +728,34 @@ You can also run `npx chrome-devtools-mcp@latest --help` to see all available co
 
 ## Concepts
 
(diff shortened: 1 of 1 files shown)


OrKoN:
Thanks for the PR. Adding this sounds fine to me. @natorion PTAL

natorion (approved):

PR #2064, opened 2026-05-15
fix: report unknown tool arguments

### LINKED ISSUE #1940: Update error handling to help agents understand errors they make

# The problem

Agents using the tools sometimes hallucinate extra parameters. This causes the tool call to fail. In some cases, this leads to several rounds of trial-and-error until the agent finally finds the correct combination or pivots. This causes unnecessary token burn.

# Examples

- #1896 - agent hallucinated `--expression` flag instead of positional parameters. There were 6 failed attempts, 2 of them can be directly attributed to the hallucinated flag and unclear error message. Two other could be a consequence of failed fixes to this issue.
- Across 72 evals, 3 trajectories contained extra "description" parameter to `navigate_page` MCP tool that could be safely ignored. The error message was "params must NOT have additional properties" which is ambiguous. In total this accounted for 5 failed tool calls. This seems to be aggravated by the evals asking for the agent to be verbose and describe each tool usage, but that is not an unreasonable usage scenario in itself.
- Similar to the above, 1 run using CLI version had 2 attempts to use `click` tool with extra `--filePath ...` parameter, likely hallucinated from `take_snapshot` tool (both calls used `--includeSnapshot true`). In this case however error message clearly stated "Unknown argument: filePath." and model fixed the call on the next try.

- Allow extra named flags and ignore them for tools that do not use positional arguments. This seems to be in line with some other MCPs and tools that tend to allow agents to be extra chatty with tool invocations
- Improve parsing error messages to list unexpected parameter names explicitly in all cases.

 NA

popsiclelmlm:
## Summary
- register tool input schemas as passthrough so extra named arguments reach ToolHandler validation
- report unknown arguments with an explicit error that names the unknown and expected arguments
- stop before invoking tool handlers when unknown arguments are present
- add ToolHandler coverage for reporting an extra argument

Fixes #1940

## Tests
- `npm run check-format`
- `npx tsc --noEmitOnError false` *(emits build artifacts but still reports the existing `chrome-devtools-frontend` type conflict in `ModelImpl.ts`)*
- `node --experimental-strip-types --no-warnings=ExperimentalWarning scripts/post-build.ts`
- `NODE_TEST_REPORTER=spec npm run test:no-build -- tests/ToolHandler.test.ts`

### DIFF
--- a/src/ToolHandler.ts
+++ b/src/ToolHandler.ts
@@ -12,7 +12,8 @@ import type {Mutex} from './Mutex.js';
 import {SlimMcpResponse} from './SlimMcpResponse.js';
 import {ClearcutLogger} from './telemetry/ClearcutLogger.js';
 import {bucketizeLatency} from './telemetry/transformation.js';
-import type {CallToolResult, zod} from './third_party/index.js';
+import type {CallToolResult} from './third_party/index.js';
+import {zod} from './third_party/index.js';
 import type {ToolCategory} from './tools/categories.js';
 import {labels, OFF_BY_DEFAULT_CATEGORIES} from './tools/categories.js';
 import type {DefinedPageTool, ToolDefinition} from './tools/ToolDefinition.js';
@@ -121,8 +122,29 @@ function isPageScopedTool(
   return 'pageScoped' in tool && tool.pageScoped === true;
 }
 
+function formatArgumentNames(names: string[]): string {
+  return names.map(name => `"${name}"`).join(', ');
+}
+
+function buildUnknownArgumentsMessage(
+  toolName: string,
+  unknownArgumentNames: string[],
+  expectedArgumentNames: string[],
+): string {
+  const unknownLabel =
+    unknownArgumentNames.length === 1 ? 'argument' : 'arguments';
+  const expectedArguments = expectedArgumentNames.length
+    ? `Expected arguments: ${formatArgumentNames(expectedArgumentNames)}.`
+    : 'This tool does not accept any arguments.';
+  const correction =
+    unknownArgumentNames.length === 1 ? 'Remove it' : 'Remove them';
+
+  return `Unknown ${unknownLabel} for tool "${toolName}": ${formatArgumentNames(unknownArgumentNames)}. ${expectedArguments} ${correction} and retry.`;
+}
+
 export class ToolHandler {
   readonly inputSchema: zod.ZodRawShape;
+  readonly registeredInputSchema: zod.ZodTypeAny;
   readonly shouldRegister: boolean;
   private readonly disabledReason?: string;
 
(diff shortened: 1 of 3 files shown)


OrKoN (commented):
per https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1940 I do not think we want to silently swallow unknown parameters instead we want to improve the error message? cc @zyzyzyryxy

popsiclelmlm:
Thanks, that makes sense. I updated the PR so unknown arguments are no longer silently ignored. The registered schema now lets the call reach the tool handler, but the handler returns an explicit error naming the unknown argument(s) and the expected arguments, then stops before invoking the tool.\n\nI also retitled the PR and amended the commit to reflect that behavior. Verified with:\n- `npm run check-format`\n- `NODE_TEST_REPORTER=spec npm run test:no-build -- tests/ToolHandler.test.ts` after rebuilding the emitted JS with `npx tsc --noEmitOnError false` and `scripts/post-build.ts`\n\nThe full TypeScript check still hits the existing `chrome-devtools-frontend` duplicate type declaration reported earlier.

src/ToolHandler.ts
```diff
@@ -121,8 +121,29 @@ function isPageScopedTool(
...
+): string {
+  const unknownLabel =
+    unknownArgumentNames.length === 1 ? 'argument' : 'arguments';
+  const expectedArguments = expectedArgumentNames.length
+    ? ` Expected arguments: ${formatArgumentNames(expectedArgumentNames)}.`
+    : ' This tool does not accept any arguments.';
+  const correction =
+    unknownArgumentNames.length === 1 ? 'Remove it' : 'Remove them';
+
+  return `Unknown ${unknownLabel} for tool "${toolName}": ${formatArgumentNames(unknownArgumentNames)}.${expectedArguments} ${correction} and retry.`;
```
zyzyzyryxy: Missing space after the dot
```suggestion
  return `Unknown ${unknownLabel} for tool "${toolName}": ${formatArgumentNames(unknownArgumentNames)}. ${expectedArguments} ${correction} and retry.`;
```

src/ToolHandler.ts
```diff
@@ -12,7 +12,7 @@ import type {Mutex} from './Mutex.js';
 import {SlimMcpResponse} from './SlimMcpResponse.js';
 import {ClearcutLogger} from './telemetry/ClearcutLogger.js';
 import {bucketizeLatency} from './telemetry/transformation.js';
-import type {CallToolResult, zod} from './third_party/index.js';
+import {type CallToolResult, zod} from './third_party/index.js';
```
zyzyzyryxy: Let's keep type-only imports as separate statement, consistently with other imports.
```suggestion
import type {CallToolResult} from './third_party/index.js';
import {zod} from './third_party/index.js';
```

zyzyzyryxy (commented):
Looks about right.
CLA still seem to be not resolved, so cannot approve just yet.

popsiclelmlm:
Thanks again. I applied both review suggestions and updated the PR description to match the current behavior.

Verified again with:
- `npm run check-format`
- `npx tsc --noEmitOnError false` (still emits build artifacts and still reports the existing `chrome-devtools-frontend` duplicate type declaration in `ModelImpl.ts`)
- `node --experimental-strip-types --no-warnings=ExperimentalWarning scripts/post-build.ts`
- `NODE_TEST_REPORTER=spec npm run test:no-build -- tests/ToolHandler.test.ts`

The `cla/google` check is passing now. Could you please take another look when you have a chance?

zyzyzyryxy (approved):

PR #2070, opened 2026-05-17
fix: include saved image paths in CLI JSON output

### LINKED ISSUE #1910: feat(cli): add --output-format=json flag for binary tool responses
## Problem

When using the Chrome DevTools MCP CLI programmatically (e.g. in scripts or CI pipelines), binary tool responses like screenshots are returned as human-readable text summaries (e.g. `Saved to /tmp/abc.png.`) rather than structured machine-readable data. This makes it difficult to parse and use the actual file paths or base64 data in downstream tools.

## Solution

Extend the existing `--output-format=json` flag (documented for `list_pages`) so it applies to **all** CLI commands — including those that return binary content (screenshot, performance_analyze_insight, etc.). When `--output-format=json` is set, the CLI should return a structured JSON object that includes:

- `tool`: the tool name
- `filepath`: (for file outputs) the absolute path to the saved file
- `data` or `base64`: (optionally) the raw base64 data
- `metadata`: any relevant metadata (dimensions, format, etc.)

## Use case

Scripted regression tests or CI pipelines that call `chrome-devtools take_screenshot --output-format=json` and need to programmatically assert on the output file path or upload the resulting image — without parsing human-readable text.

## References

- CLI docs already mention `--output-format=json` for `list_pages` (`docs/cli.md`)
- Binary tools currently return text like `Saved to /tmp/abc.png.` regardless of format flag

popsiclelmlm:
Fixes #1910

Summary:
- save image content before returning structured CLI JSON output
- include saved image metadata in `structuredContent.images`
- add coverage for JSON output with image content

Tests:
- `npm test -- tests/daemon/client.test.ts`
- `npx eslint src/daemon/client.ts tests/daemon/client.test.ts && npx prettier --check src/daemon/client.ts tests/daemon/client.test.ts`


nroscino (approved):
Hey! Thank you for the PR! 
LGTM!

PR #2084, opened 2026-05-19
fix: make pageId required

samiyac:
Refs: #2052 

Without having pageId as required, when the flag experimentalPageIdRouting is enabled, the pageId field is not always populated. This was found while running eval scenarios.

We only add `pageIdSchema` when tool is page scoped and experimentalPageIdRouting is enabled, so it should be safe to make this change.


OrKoN (approved):
Thanks, can you please update PR description with details on why it was needed and how it works.

PR #2085, opened 2026-05-19
test: update all the eval scenarios to work with experimentalPageIdRouting flag

samiyac:
Refs #2052 

This change updates all the eval scenarios so that they work with and without experimentalPageIdRouting flag

### DIFF
--- a/scripts/eval_result.ts
+++ b/scripts/eval_result.ts
@@ -1,0 +1,107 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'node:assert';
+
+export interface CapturedFunctionCall {
+  name: string;
+  args: Record<string, unknown>;
+}
+
+export class Result {
+  private nextCallIndex = 0;
+  public readonly calls: CapturedFunctionCall[];
+  public readonly serverArgs: string[];
+
+  constructor(calls: CapturedFunctionCall[], serverArgs: string[]) {
+    this.calls = calls;
+    this.serverArgs = serverArgs;
+  }
+
+  get hasPageIdRouting(): boolean {
+    return this.serverArgs.includes('--experimental-page-id-routing');
+  }
+
+  get remainingCalls(): CapturedFunctionCall[] {
+    return this.calls.slice(this.nextCallIndex);
+  }
+
+  /**
+   * Consumes initial page navigation/setup boilerplate.
+   * - Ignores/skips leading list_pages calls.
+   * - Asserts that new_page or navigate_page was called.
+   * - Determines the expected pageId.
+   * - Returns the active pageId.
+   */
+  consumePageNavigation(): number | undefined {
+    if (this.calls[this.nextCallIndex]?.name === 'list_pages') {
+      this.nextCallIndex++;
+    }
+
+    const navCall = this.calls[this.nextCallIndex];
+    assert.ok(
+      navCall &&
+        (navCall.name === 'new_page' || navCall.name === 'navigate_page'),
+      `Expected navigation call (new_page or navigate_page), but got: ${navCall?.name || 'none'}`,
+    );
+    this.nextCallIndex++;
+
+    const isNewPage = navCall.name === 'new_page';
+    let pageId: number | undefined;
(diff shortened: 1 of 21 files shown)


scripts/eval_result.ts
```diff
@@ -0,0 +1,100 @@
...
+      remaining: new Result(this.calls.slice(index + 1), this.serverArgs),
+      pageId,
+    };
+  }
+
+  /**
+   * Asserts that a call exists at index, has the correct name, and matches expected arguments.
+   */
+  assertCall(
+    index: number,
```
OrKoN: can we drop the index and use `assertNextCall` as the name?

OrKoN: we would rely on the order of assertions

scripts/eval_scenarios/console_test.ts
```diff
@@ -20,16 +20,13 @@ export const scenario: TestScenario = {
...
-  expectations: calls => {
-    assert.strictEqual(calls.length, 2);
-    assert.ok(
-      calls[0].name === 'navigate_page' || calls[0].name === 'new_page',
-      'First call should be navigation',
-    );
-    assert.strictEqual(
-      calls[1].name,
+  expectations: result => {
+    const {remaining, pageId} = result.consumePageNavigation();
```
OrKoN: I wonder if we should get rid of remaining and just mutate the result?

OrKoN:
```
const {pageId} = result.assertCallsNavigation();
result.asserRemainingCalls(1)
result.assertNextCall('list_console_messages', 
   remaining.hasPageIdRouting ? {pageId} : undefined,)
``

File: scripts/eval_result.ts (hunk 1 above)
Comment (samiyac) [reply scripts/eval_result.ts]: Done

File: scripts/eval_scenarios/console_test.ts (hunk 2 above)
Comment (samiyac) [reply scripts/eval_scenarios/console_test.ts]: Done

File: scripts/eval_scenarios/emulation_test.ts [hunk 3]
```diff
@@ -11,9 +11,15 @@ import type {TestScenario} from '../eval_gemini.ts';
...
   prompt: 'Emulate offline network conditions.',
   maxTurns: 2,
-  expectations: calls => {
-    assert.strictEqual(calls.length, 1);
-    assert.strictEqual(calls[0].name, 'emulate');
-    assert.strictEqual(calls[0].args.networkConditions, 'Offline');
+  expectations: result => {
+    assert.ok(result.remainingCalls.length >= 2);
+    result.assertNextCall('list_pages');
+    const expectedArgs: Record<string, unknown> = {
```
OrKoN: nit: but it would really nice to do it inline instead of a tmp variable. If undefined needs special handling in assertions, perhaps we could do it within assert?

```
result.assertNextCall('emulate', {
  networkConditions: 'Offline',
  pageId: result.hasPageIdRouting ? 1 : undefined,
});
```

OrKoN (approved):
LGTM thanks!

scripts/eval_scenarios/emulation_viewport_test.ts
```diff
@@ -9,11 +9,19 @@ import assert from 'node:assert';
...
+  prompt: 'Emulate current page with iPhone 14 viewport',
   maxTurns: 2,
-  expectations: calls => {
-    assert.strictEqual(calls.length, 1);
-    assert.strictEqual(calls[0].name, 'emulate');
-    assert.deepStrictEqual(calls[0].args.viewport, '390x844x3,mobile,touch');
+  expectations: result => {
+    assert.ok(result.remainingCalls.length >= 1);
+    if (result.hasPageIdRouting) {
+      result.assertNextCall('list_pages');
```
OrKoN, line 20: I sometimes get list_pages without pageId routing.

scripts/eval_scenarios/emulation_test.ts
samiyac: Done

scripts/eval_scenarios/emulation_viewport_test.ts
samiyac: Updated to allow for this

PR #2089, opened 2026-05-20
[Prototype] Heapsnapshot diff tool

Lightning00Blade:
This is rebase of on old stash, updated to use the current tool structure. 

One thing that is not used and utilize is the `createDeletedNodesProvider` which can show cases how the heapsnapshot shrunk. 

When testing the LLM consistently used the tool to get the diff, but failed often to invoke the specialized tool for debugging it.


Lightning00Blade:
Closed by https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2198

Issue #1955, opened 2026-04-25
performance_start_trace records 'CPU throttling: none / Network throttling: none' despite emulate setting throttling — measurements come back partially or fully unthrottled


tracedef:
## Description

When `emulate` is called with `cpuThrottlingRate=4` and `networkConditions="Slow 4G"` and a subsequent `performance_start_trace` records a page load, the trace summary's metadata reports `CPU throttling: none / Network throttling: none` regardless of the emulation state. The recorded LCP comes back substantially faster than PSI Lighthouse 12 mobile measurements with identical configured throttling — typically 3–6× faster — suggesting the throttling either isn't propagating to the trace agent or is being applied incompletely.

## Environment

- **chrome-devtools-mcp version:** 0.23.0 (`npx chrome-devtools-mcp@latest`, latest as of 2026-04-22)
- **Coding agent:** Claude Code (Opus 4.7)
- **OS:** macOS 26.4.1 (Darwin 25.4.0)
- **Configuration:** `--channel=stable --isolated=true --viewport=1920x1080`

## Reproduction

Target: `https://jumpermedia.co/` (a CrUX-available, real-world WordPress + WP Rocket + Cloudflare site). Mobile field LCP from CrUX is 1,712 ms; PSI Lighthouse 12 mobile lab LCP is 6,831 ms.

```
1. mcp__chrome-devtools__new_page url="about:blank" isolatedContext="repro-1"
2. mcp__chrome-devtools__emulate
     viewport="412x823x2.625,mobile,touch"
     userAgent="Mozilla/5.0 (Linux; Android 11; moto g power (2022)) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/147.0.0.0 Mobile Safari/537.36"
     networkConditions="Slow 4G"
     cpuThrottlingRate=4
3. mcp__chrome-devtools__performance_start_trace reload=false autoStop=false
4. mcp__chrome-devtools__navigate_page type="url" url="https://jumpermedia.co/"
5. (sleep ~30 seconds for full load + delay-JS firing after a synthetic interaction event)
6. mcp__chrome-devtools__performance_stop_trace filePath=/tmp/test.json
```

Trace summary returned:

```
URL: about:blank
Trace bounds: {min: ..., max: ...}
CPU throttling: none
Network throttling: none

## insight set id: NAVIGATION_0
URL: https://jumpermedia.co/
Metrics (lab / observed):
  - LCP: 1173 ms, ...
[...]

PR #2092, opened 2026-05-20
fix: Apply CPU throttling to secondary CDP session

wolfib:
This addresses #1955 

CPU throttling needs to be applied to both the primary puppeteer session and the secondary CDP session from the DevTools universe to have an effect.

For network throttling this does not seem to be the case, I can see a slowdown with the current implementation which only applies the network throttling to the primary CDP session.

I also had to increase the navigation timeout to prevent timeout errors.

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -481,7 +481,7 @@ export class McpContext implements Context {
       page.networkConditions,
     );
     page.pptrPage.setDefaultNavigationTimeout(
-      NAVIGATION_TIMEOUT * networkMultiplier,
+      NAVIGATION_TIMEOUT * networkMultiplier * cpuMultiplier,
     );
   }
 
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -329,11 +329,22 @@ export class McpContext implements Context {
       newSettings.networkConditions = options.networkConditions;
     }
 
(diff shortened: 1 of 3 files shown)


src/McpContext.ts
```diff
@@ -329,11 +329,22 @@ export class McpContext implements Context {
       newSettings.networkConditions = options.networkConditions;
     }
 
+    const secondarySession = this.getDevToolsUniverse(mcpPage)?.session;
     if (!options.cpuThrottlingRate) {
       await page.emulateCPUThrottling(1);
+      if (secondarySession) {
```
OrKoN, line 335: I am surprised this is needed because CPUthrottling is global IIRC. Are we sure DevTools code does not reset CPU throttling upon instantiation?

tests/tools/emulation.test.ts
```diff
@@ -209,6 +209,36 @@ describe('emulation', () => {
       });
     });
 
+    it('applies cpu throttling to secondary session', async () => {
```
OrKoN, line 212: can we test instead that throttling apply correctly to performance trace results?

OrKoN (approved):
LGTM as a temporary solution but let's fix it upstream and file a task to clean up.

PR #2096, opened 2026-05-21
fix: Fix throttling info in performance trace output

wolfib:
This addresses #1955.

Throttling information needs to be passed to the parser for the output to be correct.


OrKoN (approved):

abhishekftpshah-collab:
Call detector

abhishekftpshah-collab:
Phone hack

Issue #1765, opened 2026-03-30
Chrome process persists after MCP plugin is disabled


hengrumay:
## Summary

When the `chrome-devtools-mcp` plugin is disabled (set to `false` in Claude Code `enabledPlugins`), the Chrome instance it spawned continues running indefinitely. Had to manually kill the processes days later.

## Steps to reproduce

1. Install `chrome-devtools-mcp@claude-plugins-official` in Claude Code
2. Use it (Chrome spawns with `--user-data-dir=~/.cache/chrome-devtools-mcp/chrome-profile`)
3. Disable the plugin in settings: `"chrome-devtools-mcp@claude-plugins-official": false`
4. Chrome processes persist — still running days later

## Expected behavior

Disabling the plugin (or exiting Claude Code) should terminate the spawned Chrome instance.

## Workaround

```bash
pkill -f 'chrome-devtools-mcp/chrome-profile'
```

## Environment

- macOS (Darwin 25.3.0)
- Claude Code CLI
- Chrome 146.x

OrKoN:
@hengrumay could you please confirm that the chrome-devtoos-mcp server instance is not running? I think claude code might not be shutting down the MCP server gracefully or not shutting it down at all. E.g., calling `kill -s SIGTERM <pid>` or `kill -9 <pid>` allows chrome-devtools-mcp to clean up the child browser process.

hengrumay:
Thanks for the suggested resolution!

Confirmed — the `chrome-devtools-mcp` server instances were not being shut down by Claude Code.

I had **30 orphaned processes** (11 MCP server instances + their watchdog/npm parent processes) accumulated over ~2 weeks across v0.20.0 and v0.20.2.

After running `kill -s SIGTERM` on all PIDs:
- 28/30 terminated cleanly on first SIGTERM
- 2 required a second SIGTERM (these were the active MCP instance for the current session)
- All watchdog children cleaned up properly — no SIGKILL needed

So SIGTERM cleanup works fine on chrome-devtools-mcp's side. The issue seems to be on the Claude Code side — not sending SIGTERM when a session exits or the plugin is disabled.

As a workaround, I've added a Stop hook in my Claude Code settings that runs `pkill -f 'chrome-devtools-mcp'` on session exit.

OrKoN:
[...]

PR #2117, opened 2026-05-24
fix: exit on stdin EOF and SIGTERM/SIGINT/SIGHUP, closing the browser cleanly

### LINKED ISSUE #2116: Server process does not exit on stdin EOF after a page has been opened
### Description of the bug

`chrome-devtools-mcp` does not exit when its stdin is closed if a `navigate_page` (or any tool that opens a Chrome instance) has been issued during the session. The MCP protocol does not define a `shutdown` RPC and convention across stdio-based MCP servers is to exit on stdin EOF; without that, callers that cleanly close stdin have to fall back to `SIGTERM` / `SIGKILL` on every session that touched the browser.

Sessions that never opened a page (e.g. just `tools/list`) do exit on stdin EOF, in ~30 ms — so the transport's read loop terminates correctly. The issue is that opening Chrome adds a ref to the Node event loop and no shutdown handler closes it.

### Reproduction

Minimal Node script + setup, with no MCP client involved:

```bash
mkdir cdmcp-repro && cd cdmcp-repro
npm init -y
npm i chrome-devtools-mcp
# save repro.mjs (see below)
node repro.mjs
```

`repro.mjs`:

```js
// Spawn chrome-devtools-mcp, do the MCP handshake, run a tool call,
// close stdin, and time how long until the child exits (5s SIGTERM
// fallback so the script always ends).
import { spawn } from "node:child_process"
import { performance } from "node:perf_hooks"
import { fileURLToPath } from "node:url"
import path from "node:path"

const here = path.dirname(fileURLToPath(import.meta.url))
const BIN = path.join(here, "node_modules", ".bin", "chrome-devtools-mcp")
const KILL_AFTER_MS = 5000

function send(child, msg) {
  child.stdin.write(JSON.stringify(msg) + "\n")
}

function drive(requests, label) {
  return new Promise((resolve) => {
    const child = spawn(BIN, [], { stdio: ["pipe", "pipe", "pipe"] })
    const responses = new Map()
    let buf = ""
    child.stdout.on("data", (b) => {
      buf += b.toString()
      let nl
      while ((nl = buf.indexOf("\n")) >= 0) {
        const line = buf.slice(0, nl); buf = buf.slice(nl + 1)
        if (!line.trim()) continue
        try { const p = JSON.parse(line); if (p.id !== undefined) responses.set(p.id, p) } catch {}
      }
    })
    child.stderr.on("data", () => {})
    send(child, { jsonrpc: "2.0", id: 1, method: "initialize",
      params: { protocolVersion: "2024-11-05", capabilities: {},
        clientInfo: { name: "repro", version: "0.0.0" } } })
    child.stdin.write(JSON.stringify({ jsonrpc: "2.0", method: "notifications/initialized" }) + "\n")
    const tick = setInterval(() => {
      if (!responses.has(1)) return
      clearInterval(tick)
      for (const r of requests) send(child, { jsonrpc: "2.0", ...r })
      const tick2 = setInterval(() => {
        if (!requests.every((r) => responses.has(r.id))) return
        clearInterval(tick2)
        const t0 = performance.now()
        child.stdin.end()
        let killed = false
        const kill = setTimeout(() => { killed = true; child.kill("SIGTERM") }, KILL_AFTER_MS)
        child.on("exit", () => {
          clearTimeout(kill)
          console.log(`${label}: exit ${(performance.now() - t0).toFixed(0)}ms ${killed ? "(SIGTERM)" : "(clean)"}`)
          resolve()
        })
      }, 50)
    }, 50)
  })
}

await drive([{ id: 2, method: "tools/list" }], "tools/list only      ")
await drive(
  [{ id: 2, method: "tools/call", params: { name: "navigate_page", arguments: { url: "https://example.com" } } }],
  "navigate example.com ",
)
```

### Observed

Across 10 invocations (40 total scenarios; clean Linux box, no other tabs, no MCP client):

```
iter 1:  tools/list only      : exit 30ms (clean)  | navigate example.com : exit 5078ms (SIGTERM)
iter 2:  tools/list only      : exit 36ms (clean)  | navigate example.com : exit 5081ms (SIGTERM)
iter 3:  tools/list only      : exit 35ms (clean)  | navigate example.com : exit 5085ms (SIGTERM)
iter 4:  tools/list only      : exit 34ms (clean)  | navigate example.com : exit 5097ms (SIGTERM)
iter 5:  tools/list only      : exit 36ms (clean)  | navigate example.com : exit 5100ms (SIGTERM)
iter 6:  tools/list only      : exit 33ms (clean)  | navigate example.com : exit 5084ms (SIGTERM)
iter 7:  tools/list only      : exit 33ms (clean)  | navigate example.com : exit 5078ms (SIGTERM)
iter 8:  tools/list only      : exit 37ms (clean)  | navigate example.com : exit 5077ms (SIGTERM)
iter 9:  tools/list only      : exit 30ms (clean)  | navigate example.com : exit 5090ms (SIGTERM)
iter 10: tools/list only      : exit 34ms (clean)  | navigate example.com : exit 5082ms (SIGTERM)
```

100% / 100% — never flakes either way.

### Expectation

After stdin EOF, the server closes the browser (and any other long-lived resources) and exits cleanly within a small bounded time (~1 s seems generous given the Chrome shutdown cost), in line with stdio MCP convention. Callers should not need to `SIGTERM` every page-loaded session to reclaim the process.

### Investigation

`build/src/bin/chrome-devtools-mcp-main.js` has no shutdown handler — no `process.on('SIGTERM' | 'SIGINT' | 'exit')`, no `process.stdin.on('end' | 'close')`. It does `await server.connect(transport)` and ends. When stdin closes, the StdioServerTransport's reader returns EOF but nothing closes the browser, so the Chrome subprocess keeps the Node event loop ref'd and the process never exits.

There IS a `process.stdin.on('end' / 'close')` handler in `build/src/telemetry/watchdog/main.js` (the `"Parent death detected (stdin end). Sending shutdown event..."` log line some users have seen), but that's a separate telemetry sub-process; it doesn't reap the main server or the browser. There's also `SIGTERM` / `SIGINT` / `SIGHUP` handling in `build/src/daemon/daemon.js`, but the default stdio flow doesn't go through the daemon.

Tried as a workaround: sending `tools/call close_page` before `stdin.end()` (5/5 still SIGTERM) — closing the page doesn't reap Chrome itself.

### Suggested fix

In `chrome-devtools-mcp-main.ts` (or wherever `createMcpServer` constructs the browser-owning context), register `process.stdin.on('end', ...)` and `process.stdin.on('close', ...)` handlers that close the browser and then `process.exit(0)`. Same handler can be wired to `SIGTERM` and `SIGINT` so the bounded teardown also works for clients that signal the process instead of closing stdin.

### MCP configuration

Direct spawn (no MCP client / coding agent involved). Default flags.

1.0.1

### Chrome version

148.0.7778.178

### Coding agent version

N/A — reproduced by piping JSON-RPC directly into the MCP server's stdin.

### Model version

N/A

### Chat log

N/A

### Node version

v24.11.1

### Operating system

Linux

### Extra checklist

dballesteros7:
Fixes #2116.

`chrome-devtools-mcp-main.ts` currently has no shutdown handler. After a session calls `navigate_page` (or anything else that launches Chrome), the Chrome subprocess keeps the Node event loop ref'd, so closing stdin (the stdio MCP convention for "I'm done") doesn't make the server exit. Callers that close stdin to terminate the server have to fall back to SIGTERM / SIGKILL on every page-loaded session — deterministically, not flakily.

This change:

1. Adds `closeBrowser()` in `browser.ts` that calls `browser.close()` for launched instances (reaps the Chrome subprocess) and `browser.disconnect()` for attached instances (leaves the user's Chrome alive). No-op if no browser is active or the connection has already been dropped.

2. Registers shutdown handlers in `chrome-devtools-mcp-main.ts` for:
   - `stdin.on('end' | 'close')` — stdio MCP transport convention
   - `SIGTERM` / `SIGINT` / `SIGHUP` — clients that signal instead of closing stdin (`SIGHUP` for parity with `src/daemon/daemon.ts`)

   The handler is idempotent (guarded `shuttingDown` flag), and has an unref'd 10s timeout backstop in case Chrome teardown hangs (slow `beforeunload` handlers, many tabs, etc.).

### Note on scope

This complements (does not replace) the client-side fixes filed against #1765, e.g. google-gemini/gemini-cli#13391 and anthropics/claude-code#42300. The MCP stdio convention is that closing stdin signals shutdown; a server that doesn't honor that forces every client to special-case it. The watchdog sub-process (`src/telemetry/watchdog/main.ts:145-146`) and the daemon (`src/daemon/daemon.ts:224-230`) both already implement this for the same reason — this PR extends the same pattern to the main entry point so all three execution paths behave consistently.

### Measurement

Repro script in #2116, same env (chrome-devtools-mcp@1.0.1, Chrome 148.0.7778.178, Node v24.11.1, Linux), 10 iterations:

| Scenario | Before | After |
|---|---|---|
| `tools/list only` (no navigation) | 10/10 clean, 30-37 ms | 10/10 clean, 29-40 ms |
| `navigate example.com` | 10/10 SIGTERM at ~5080 ms | 10/10 clean at 145-180 ms |

### Notes

- I didn't add a subprocess-based test for this; the existing `tests/utils.ts:runCli` infrastructure targets the `chrome-devtools` CLI, not the stdio MCP server, and a shutdown-timing test would introduce non-trivial Chrome-startup flakiness in CI. Happy to add one if maintainers want it — pointer to the right test directory appreciated.

### DIFF
--- a/src/bin/chrome-devtools-mcp-main.ts
+++ b/src/bin/chrome-devtools-mcp-main.ts
@@ -8,6 +8,7 @@ import '../polyfill.js';
 
 import process from 'node:process';
 
+import {closeBrowser} from '../browser.js';
 import {createMcpServer, logDisclaimers} from '../index.js';
 import {logger, saveLogsToFile} from '../logger.js';
 import {ClearcutLogger} from '../telemetry/ClearcutLogger.js';
@@ -31,6 +32,44 @@ if (process.env['CHROME_DEVTOOLS_MCP_CRASH_ON_UNCAUGHT'] !== 'true') {
     logger('Unhandled promise rejection', promise, reason);
   });
 }
+
+// Shutdown on stdin EOF (stdio MCP convention — the client closes the
+// transport to signal exit) and on standard termination signals. Without
+// this, an active Chrome subprocess keeps the Node event loop ref'd after
+// stdin closes and the server hangs until something else kills it.
+let shuttingDown = false;
+async function shutdown(reason: string): Promise<void> {
+  if (shuttingDown) {
+    return;
+  }
+  shuttingDown = true;
+  logger(`Shutting down (${reason})`);
+  // Backstop in case browser teardown hangs (e.g. unresponsive Chrome,
+  // slow beforeunload handlers, many tabs). Exits 0 because we still
+  // honored the shutdown request; the log line preserves observability.
+  // Unref'd so it doesn't keep the loop alive on the clean path.
+  setTimeout(() => {
+    logger('Shutdown timeout exceeded, forcing exit');
+    process.exit(0);
+  }, 10000).unref();
+  await closeBrowser();
+  process.exit(0);
+}
+process.stdin.on('end', () => {
+  void shutdown('stdin end');
+});
+process.stdin.on('close', () => {
+  void shutdown('stdin close');
+});
+process.on('SIGTERM', () => {
+  void shutdown('SIGTERM');
+});
+process.on('SIGINT', () => {
+  void shutdown('SIGINT');
+});
+process.on('SIGHUP', () => {
+  void shutdown('SIGHUP');
+});
 
 logger(`Starting Chrome DevTools MCP Server v${VERSION}`);
 const {server} = await createMcpServer(args, {
--- a/tests/shutdown.test.ts
+++ b/tests/shutdown.test.ts
@@ -1,0 +1,180 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'node:assert';
+import type {ChildProcessByStdio} from 'node:child_process';
+import {spawn} from 'node:child_process';
+import type {Readable, Writable} from 'node:stream';
+import {describe, it} from 'node:test';
+
+import {executablePath} from 'puppeteer';
+
+type Server = ChildProcessByStdio<Writable, Readable, Readable>;
+
+// Once shutdown is signalled, the server should be fully gone within this
+// budget. The actual fast path is well under 500ms; the budget is set to be
+// generous against CI noise without being so loose that it would hide a hang.
+const SHUTDOWN_BUDGET_MS = 3000;
+// Outer test timeout. If exit doesn't happen within this, treat as a hang
+// (the bug we're guarding against) and SIGKILL the subprocess.
+const EXIT_TIMEOUT_MS = 15000;
+
+async function spawnServer(): Promise<Server> {
+  const child = spawn(
+    'node',
+    [
+      'build/src/bin/chrome-devtools-mcp.js',
+      '--headless',
+      '--isolated',
+      '--executable-path',
+      await executablePath(),
+    ],
+    {
+      env: {
+        ...process.env,
+        CHROME_DEVTOOLS_MCP_NO_USAGE_STATISTICS: 'true',
+      },
+      stdio: ['pipe', 'pipe', 'pipe'],
+    },
+  ) as Server;
+  // Drain stderr to avoid pipe-buffer backpressure stalling the server.
+  child.stderr.on('data', () => {
+    // discard
+  });
+  return child;
+}
+
+async function waitForExit(
+  child: Server,
+  timeoutMs: number,
+): Promise<{
+  code: number | null;
+  signal: NodeJS.Signals | null;
+  elapsedMs: number;
+}> {
+  const start = Date.now();
+  return await new Promise((resolve, reject) => {
+    const timer = setTimeout(() => {
+      child.kill('SIGKILL');
+      reject(new Error(`server did not exit within ${timeoutMs}ms`));
+    }, timeoutMs);
+    child.once('exit', (code, signal) => {
+      clearTimeout(timer);
+      resolve({code, signal, elapsedMs: Date.now() - start});
+    });
+  });
+}
+
+async function rpc(
+  child: Server,
+  msg: {method: string; params?: unknown},
+): Promise<unknown> {
+  const id = Math.floor(Math.random() * 1e9);
+  const payload = JSON.stringify({jsonrpc: '2.0', id, ...msg}) + '\n';
+  return await new Promise((resolve, reject) => {
+    let buf = '';
+    const onData = (chunk: Buffer) => {
+      buf += chunk.toString();
+      const lines = buf.split('\n');
+      buf = lines.pop() ?? '';
+      for (const line of lines) {
+        if (!line.trim()) {
+          continue;
+        }
+        try {
+          const parsed = JSON.parse(line) as {id?: number};
+          if (parsed.id === id) {
+            child.stdout.off('data', onData);
+            resolve(parsed);
+            return;
+          }
+        } catch {
+          // Not a JSON message; ignore.
+        }
+      }
+    };
+    child.stdout.on('data', onData);
+    const onExit = () => {
+      child.stdout.off('data', onData);
+      reject(new Error('server exited before RPC response'));
+    };
+    child.once('exit', onExit);
+    child.stdin.write(payload);
+  });
+}
+
+function notify(child: Server, msg: {method: string; params?: unknown}): void {
+  child.stdin.write(JSON.stringify({jsonrpc: '2.0', ...msg}) + '\n');
+}
+
+async function initializeAndLaunchBrowser(child: Server): Promise<void> {
+  await rpc(child, {
+    method: 'initialize',
+    params: {
+      protocolVersion: '2024-11-05',
+      capabilities: {},
+      clientInfo: {name: 'shutdown-test', version: '0.0.1'},
+    },
+  });
+  notify(child, {method: 'notifications/initialized'});
+  // navigate_page forces a real Chrome launch — this is what reproduces
+  // the hang in #2116. Without an active Chrome subprocess, stdin EOF
+  // would close the event loop on its own and shutdown would look fine
+  // even with broken handlers.
+  await rpc(child, {
+    method: 'tools/call',
+    params: {
+      name: 'navigate_page',
+      arguments: {url: 'about:blank'},
+    },
+  });
+}
+
+describe('shutdown', () => {
+  it('exits within budget on stdin EOF after Chrome launch', async () => {
+    const child = await spawnServer();
+    await initializeAndLaunchBrowser(child);
+    child.stdin.end();
+    const {elapsedMs} = await waitForExit(child, EXIT_TIMEOUT_MS);
+    assert.ok(
+      elapsedMs < SHUTDOWN_BUDGET_MS,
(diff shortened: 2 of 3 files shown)


src/bin/chrome-devtools-mcp-main.ts
```diff
@@ -32,6 +33,44 @@ if (process.env['CHROME_DEVTOOLS_MCP_CRASH_ON_UNCAUGHT'] !== 'true') {
...
+  await closeBrowser();
+  process.exit(0);
+}
+process.stdin.on('end', () => {
+  void shutdown('stdin end');
+});
+process.stdin.on('close', () => {
+  void shutdown('stdin close');
+});
+process.on('SIGTERM', () => {
```
OrKoN, line 64: how does this work together with signal handling installed by Puppeteer for launched browsers?

OrKoN: I'd say it should be sufficient to handle stdin only on the mcp server as I expect the browser shutdown to be handled by puppeteer. Would you mind adding a test for these situations?

OrKoN (commented):
Thanks, generally looks good to me!

src/bin/chrome-devtools-mcp-main.ts
dballesteros7: Good question — I investigated to be sure. Findings:

Tested puppeteer 25.0.4 directly (script reads stdin to simulate the MCP transport keeping the loop alive, then sends each signal):

| Mode    | Signal  | Puppeteer listeners | Outcome without our handler |
|---------|---------|---------------------|-----------------------------|
| launch  | SIGTERM | 1                   | **Hangs forever** — Puppeteer fires `void this.close()` and returns; suppresses Node's default-terminate but never replaces it with `process.exit`. Stdin keeps the loop alive. |
| launch  | SIGHUP  | 1                   | Same as SIGTERM. |
| launch  | SIGINT  | 1                   | Exits 130 — the one case Puppeteer fully handles (`this.kill()` + `process.exit(130)`). |
| connect | SIGTERM | 0                   | Default terminates (143), no `disconnect()` call. |
| connect | SIGHUP  | 0                   | Default terminates (129), no `disconnect()` call. |
| connect | SIGINT  | 0                   | Default terminates (130), no `disconnect()` call. |

The launched-browser hang on SIGTERM/SIGHUP comes from [`packages/browsers/src/launch.ts#L412-L426`](https://github.com/puppeteer/puppeteer/blob/puppeteer-core-v25.0.4/packages/browsers/src/launch.ts#L412-L426), which fires `void this.close()` and returns without exiting. The `puppeteer.connect()` path installs no signal handlers at all (only `BrowserLauncher.launch()` does), so in attached mode signals fall through to Node's default — process exits without ever calling `browser.disconnect()`.

So "stdin only" would regress 4 of these 6 cases.

Added the test in b2311a6 — spawns the binary, runs the MCP handshake, calls `navigate_page` to force a real Chrome launch (the bug only reproduces with an active Chrome subprocess), then triggers each shutdown path and asserts exit within 3s (vs a 15s hang detector). Same test pattern as `tests/index.test.ts`, no new infrastructure. Validated locally that patching out the SIGTERM/SIGHUP handlers makes those two tests time out at 15s while stdin-EOF and SIGINT still pass.

OrKoN: Thanks!

> so in attached mode signals fall through to Node's default — process exits without ever calling browser.disconnect().

process existing causes implicit disconnect, right? 

> The launched-browser hang on SIGTERM/SIGHUP comes from [packages/browsers/src/launch.ts#L412-L426](https://github.com/puppeteer/puppeteer/blob/puppeteer-core-v25.0.4/packages/browsers/src/launch.ts#L412-L426), which fires void this.close() and returns without exiting.

right, but closing the browser should release resources keeping the process alive allowing it to exit. I am curious why it would still hang?

OrKoN:
There is currently an outage with GitHub Actions https://www.githubstatus.com/incidents/gnftqj9htp0g I will trigger the test jobs once it is resolved.

src/bin/chrome-devtools-mcp-main.ts
OrKoN: It looks like there are some bugs to fix in Puppeteer as well.

OrKoN (approved):

OrKoN:
Thanks!

src/bin/chrome-devtools-mcp-main.ts
dballesteros7: Thanks for the review!

> process existing causes implicit disconnect, right?

Effectively yes for the `connect` path — when the process dies the WebSocket to the remote Chrome is dropped at the OS level, which the browser sees as a disconnect, and the remote/user-attached Chrome stays alive (the desired behavior). So no leak, just no observability — no `logger` line, no chance to flush telemetry, no symmetric "we're shutting down" hook. The explicit `disconnect()` is mostly cosmetic in attach mode.

> closing the browser should release resources keeping the process alive allowing it to exit. I am curious why it would still hang?

The chain is subtle. Puppeteer's signal listener is `void this.close()` and returns synchronously. Per [Node's signal docs](https://nodejs.org/api/process.html#signal-events), once *any* listener is registered for SIGTERM/SIGHUP, Node skips its default terminate. So:

1. Signal arrives → Puppeteer's listener runs → fires `void close()` → returns → control back to event loop.
2. `close()` eventually completes async and reaps the Chrome subprocess. So far so good.
3. But the MCP stdio transport has `process.stdin` actively being read, which independently keeps the event loop ref'd.
4. Chrome is gone, but stdin is still alive → loop never drains → process hangs.

The repro in #2116 isolates this — it's the stdin reader (simulating the transport) that turns "Chrome gets reaped" into "process never exits". The fix is the explicit `process.exit(0)` in our handler; without it, loop drain is at the mercy of whoever owns stdin.

dballesteros7: Agreed. The two upstream issues that fall out of this:

1. `BrowserLauncher.launch()` registers SIGTERM/SIGHUP listeners as `void this.close()` (no `.then(() => process.exit(...))`). SIGINT correctly exits 130; SIGTERM/SIGHUP suppress Node's default terminate without replacing it, so any caller with a non-empty event loop is left hanging.
2. `puppeteer.connect()` installs no signal handling at all, leaving attach-mode callers to handle it themselves.

tests/shutdown.test.ts
```diff
@@ -0,0 +1,180 @@
...
+  await rpc(child, {
+    method: 'tools/call',
+    params: {
+      name: 'navigate_page',
+      arguments: {url: 'about:blank'},
+    },
+  });
+}
+
+describe('shutdown', () => {
```
OrKoN, line 136: seems to be flaky on windows https://github.com/ChromeDevTools/chrome-devtools-mcp/actions/runs/26454201052/job/77883330043?pr=2140#step:9:1235

PR #2118, opened 2026-05-24
fix: pin npx version in plugin.json to avoid supply-chain drift

xiaolai:
> **Automated**: drive-by fix from [NLPM](https://github.com/xiaolai/nlpm-for-claude), an NL artifact linter. Reviewed and reproduced before submission.

**Bug**: `.claude-plugin/plugin.json` runs `npx chrome-devtools-mcp@latest`, so the resolved version floats on every MCP client restart — any future `latest` tag (including a compromised one) is picked up automatically.

**Evidence**: `"args": ["chrome-devtools-mcp@latest"]` confirmed at line 9; `package.json` declares `"version": "1.0.1"`, so a matching pinned version is available.

**Fix**: Replaces `chrome-devtools-mcp@latest` with `chrome-devtools-mcp@1.0.1` to lock to the current release; update in lockstep with future version bumps.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -6,7 +6,7 @@
     "chrome-devtools": {
       "command": "npx",
       "args": [
-        "chrome-devtools-mcp@latest"
+        "chrome-devtools-mcp@1.0.1"
       ]
     }
   }


OrKoN (changes requested):
Thanks for the PR but we would like to keep using the latest version for now as there is a delay in publishing plugins that want to avoid. Additionally, to implement this properly we should update the version here at the same time as release-please generates are a new release. cc @natorion

natorion:
> Thanks for the PR but we would like to keep using the latest version for now as there is a delay in publishing plugins that want to avoid. Additionally, to implement this properly we should update the version here at the same time as release-please generates are a new release. cc @natorion

I think mid-term we have to move to fixed versions in our various extension/plugin offerings.

Official marketplaces from vendors AFAIK already require pinned versions anyways.

OrKoN:
in that case, we should look into properly integrating plugins into the release.

OrKoN:
Filed https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2121

OrKoN:
Fixed via https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2121

PR #2127, opened 2026-05-26
fix: use realpath for MCP roots validation

OrKoN:


src/McpContext.ts
```diff
@@ -175,27 +176,83 @@ export class McpContext implements Context {
...
+        err.code === 'ENOENT'
       ) {
-        return;
+        // The file/directory doesn't exist. This is expected if we are about to create it.
+        // To handle this, canonicalize the parent directory and append the basename.
+        const dirname = path.dirname(absolutePath);
+        const basename = path.basename(absolutePath);
+        try {
+          const canonicalDir = await fsPromises.realpath(dirname);
+          canonicalPath = path.join(canonicalDir, basename);
```
wolfib: Could there be cases where a directory structure of 2 or more directories need to be created?
`/existing/first_new/second_new/foo.bar`

wolfib (approved):

src/McpContext.ts
OrKoN: Good catch, PTAL.

wolfib (approved):

OrKoN:
There is currently an outage of GitHub actions https://www.githubstatus.com/incidents/gnftqj9htp0g

PR #2133, opened 2026-05-26
fix: add missing validatePath call in getHeapSnapshotRetainers

mhuzaifashah:
All other getHeapSnapshot* methods in McpContext call this.validatePath(filePath) before delegating to HeapSnapshotManager, but getHeapSnapshotRetainers (added in b2b05a0) was missing it. Without this check, a path outside the workspace roots could bypass the access-control guard.

This aligns getHeapSnapshotRetainers with getHeapSnapshotAggregates, getHeapSnapshotStats, getHeapSnapshotStaticData, and getHeapSnapshotNodesById.


OrKoN:
I believe this is stale. The actual checks are in the tools code, not in the McpContext. The checks in the McpContext are redundant.

Issue #2115, opened 2026-05-24
emulate is not atomic: mobile viewport + network throttle in one call throws a navigation timeout and desyncs tracked state


decisive-engineering:
### Environment
- `chrome-devtools-mcp` **1.0.1** (via `npx -y chrome-devtools-mcp@latest`)
- Connected to a running Chrome via `--browser-url=http://127.0.0.1:9222`
- Windows 11, Chrome stable

### Summary
A single `emulate` call that sets **both** a mobile viewport **and** a network throttle (e.g. `Slow 3G`) reliably throws `Navigation timeout of 10000ms exceeded`. Because the handler throws partway through, the page's tracked `emulationSettings` is never updated and the per-page timeouts are never refreshed — leaving the tool's tracked emulation state out of sync with the actual CDP overrides on the page.

### Repro
1. Navigate to any page.
2. Call `emulate` in one shot:
   ```json
   {
     "viewport": "390x844x3,mobile,touch",
     "userAgent": "Mozilla/5.0 (iPhone; CPU iPhone OS 17_4 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.4 Mobile/15E148 Safari/604.1",
     "networkConditions": "Slow 3G"
   }
   ```
3. → `Error: Navigation timeout of 10000ms exceeded`.

A subsequent `emulate` call that omits `userAgent` (intending to clear it) then leaves a phantom mobile (Nexus 5) UA, because mobile device-metrics remained active from the partially-applied previous call.

### Likely cause
In `src/McpContext.js`, `emulate()` (≈L194–271) applies settings in this order:
1. network throttle (≈L213)
2. … cpu / geolocation / userAgent / colorScheme …
3. `setViewport(...)` (≈L264) — when `isMobile`/`hasTouch` change, Puppeteer applies this via a **page reload**
4. `mcpPage.emulationSettings = …` (≈L267)
5. `#updateSelectedPageTimeouts()` (≈L270), which is what finally calls `setDefaultNavigationTimeout(NAVIGATION_TIMEOUT * networkMultiplier)` (≈L338; `NAVIGATION_TIMEOUT = 10_000` at L20)

[...]

PR #2134, opened 2026-05-26
fix: Set viewport after updating timeouts when setting emulation

wolfib:
Tentatively addresses #2115

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -413,7 +413,6 @@ export class McpContext implements Context {
     }
 
     if (!options.viewport) {
-      await page.setViewport(null);
       delete newSettings.viewport;
     } else {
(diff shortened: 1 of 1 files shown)


src/McpContext.ts
```diff
@@ -404,6 +401,10 @@ export class McpContext implements Context {
       : {};
 
     this.#updateSelectedPageTimeouts();
+
+    // This should happen after updating the page timeouts.
+    // Setting the viewport can trigger a reload which we don't want to timeout.    
+    await page.setViewport(newSettings.viewport ?? null);
```
OrKoN, line 443: I think we should avoid running setViewport if it was not changed in the current update.

OrKoN (approved):

PR #2150, opened 2026-05-28
fix: centralize file path validation for tools

### LINKED ISSUE #2138: [Task]: Refactor path validation
### Task to do:

Currently we call validatePath in any tool with filePath inputs. Instead, we should add a new required annotation called `filePathFields: array<string>` requiring each tool to list field paths that require validation. The validation can then be done centrally when handling a request.

cc @Lightning00Blade

msnandhis:
Fixes #2138.

## Summary

This centralizes workspace-root validation for request file path fields instead of having individual tool handlers call `context.validatePath(...)` directly.

Changes included:

- add required `annotations.filePathFields` metadata to tool definitions
- validate annotated request fields centrally in `ToolHandler` before handler execution
- annotate existing tools that accept request path fields such as `filePath`, `requestFilePath`, `responseFilePath`, `outputDirPath`, and extension `path`
- remove duplicated per-tool validation calls for direct request path fields
- keep non-request validation in `reload_extension`, because that path is resolved from existing extension state rather than from request params
- add `ToolHandler` coverage to ensure annotated file path fields are validated before the handler runs

## Testing

- `npm run format`
- `npm run check-format`
- `npm run test tests/ToolHandler.test.ts`

I also started a full `npm run test`; it built and ran many tests locally, then remained quiet in a daemon/browser section, so I stopped the local run and cleaned up the spawned daemon processes rather than leaving them running.

### DIFF
--- a/src/tools/slim/tools.ts
+++ b/src/tools/slim/tools.ts
@@ -16,6 +16,7 @@ export const screenshot = definePageTool({
     category: ToolCategory.DEBUGGING,
     // Not read-only due to filePath param.
     readOnlyHint: false,
+    filePathFields: [],
   },
   schema: {},
   blockedByDialog: true,
@@ -39,6 +40,7 @@ export const navigate = definePageTool({
   annotations: {
     category: ToolCategory.NAVIGATION,
     readOnlyHint: false,
+    filePathFields: [],
   },
   schema: {
     url: zod.string().describe('URL to navigate to'),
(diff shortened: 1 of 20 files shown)


src/tools/slim/tools.ts
```diff
@@ -16,6 +16,7 @@ export const screenshot = definePageTool({
     category: ToolCategory.DEBUGGING,
     // Not read-only due to filePath param.
     readOnlyHint: false,
+    filePathFields: [],
```
OrKoN, line 19: Thanks for the PR, we discussed it offline and we actually want to move filePathFields to the same level as blockedByDialog (i.e., outside of annotations). cc @Lightning00Blade

PR #2152, opened 2026-05-28
refactor: use validate files on the tool level

### LINKED ISSUE #2138: [Task]: Refactor path validation

Lightning00Blade:
Fixes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2138
Closes #2150

### DIFF
--- a/src/tools/ToolDefinition.ts
+++ b/src/tools/ToolDefinition.ts
@@ -32,9 +32,7 @@ import type {
   ToolDefinition as ThirdPartyDeveloperToolDefinition,
 } from './thirdPartyDeveloper.js';
 
-export interface BaseToolDefinition<
-  Schema extends zod.ZodRawShape = zod.ZodRawShape,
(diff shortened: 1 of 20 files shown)


src/tools/ToolDefinition.ts
```diff
@@ -48,6 +48,7 @@ export interface BaseToolDefinition<
   };
   schema: Schema;
   blockedByDialog: boolean;
+  verifyFilesSchema?: Array<keyof Schema>;
```
OrKoN: let's make it required so that it is less frequently overlooked

OrKoN (approved):

PR #2159, opened 2026-05-31
input: allow drag drop-point biasing within target bounding box

innocentnortey:
## Summary

Extend the existing `drag` tool with four optional, backwards-compatible parameters that let callers bias the drop point inside the target element's bounding rectangle instead of always landing at the geometric centre:

- `to_offset_x` — CSS-pixel offset from target bounding-box top-left
- `to_offset_y` — same, vertical
- `to_fraction_x` — fractional x position in `[0, 1]`
- `to_fraction_y` — fractional y position in `[0, 1]`

When omitted, `drag(from_uid, to_uid)` behaves exactly as it does today.

## Motivation

Some drag-and-drop UIs expose a single visible drop-zone element whose geometric centre is not the safest or correct drop hotspot.

A concrete repro is Zoho Analytics Pivot View construction. The Pivot editor has three drop shelves stacked vertically (Columns / Rows / Data). Dropping a measure onto the visible Data shelf container can land in the neighbouring Rows shelf when the shelf's bounding-box centre sits close to the Rows region. Dropping onto a child element located deeper inside the intended Data shelf — e.g., an existing chip body or its remove-button — works reliably, which points at target-point geometry rather than source selection or timing as the failure mode.

Today `chrome-devtools-mcp`'s `drag` tool uses Puppeteer's default `ElementHandle.drag(toHandle)` / `toHandle.drop(fromHandle)` path, which computes the drop point as the element's bounding-rect centroid via [`clickablePoint()`](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/api/ElementHandle.ts). That makes deceptive or overlapping drop-zone layouts hard to automate reliably.

### New optional drag parameters

Mutually exclusive per axis (offset OR fraction, not both).

Examples:

```jsonc
// Bias near the top interior of the target
{ "from_uid": "...", "to_uid": "...", "to_offset_y": 12 }

// Bias toward the lower-right interior of the target
{ "from_uid": "...", "to_uid": "...", "to_fraction_x": 0.85, "to_fraction_y": 0.9 }
```

### Backwards compatibility

If none of the new parameters are provided, the tool keeps the existing implementation verbatim:

- `fromHandle.drag(toHandle)`
- fixed 50 ms pause
- `toHandle.drop(fromHandle)`

This preserves behaviour for every existing caller.

### Custom-target code path

If any of the new parameters is present, the tool:

1. Scrolls source and target into view.
2. Computes the drop point from the target element's bounding box (offset / fraction / centroid).
3. Hovers the source, presses the mouse, moves the mouse to the computed drop point, waits 150 ms for `dragover` handlers, releases the mouse.

This mirrors Puppeteer's existing HTML5-drag emulation path (the non-intercepted branch of `ElementHandle.drag`) but lets the caller specify the target point. It intentionally avoids `setDragInterception(true)` — enabling drag interception would make `ElementHandle.drag(toHandle)` (used by the default path) hang waiting for `Input.dragIntercepted`, breaking every existing call site.

The 150 ms delay is intentional for HTML5 dragover-based UIs that need processing time between move and release.

## Tests

Added a regression test in `tests/tools/input.test.ts` that verifies `drag` with `to_offset_y: 10` lands in the top region of a 120×120 drop target (i.e. the bias actually takes effect).

Existing drag test continues to pass unchanged.

## Generated artefacts

Regenerated via `npm run gen`:

- `docs/tool-reference.md` — new params documented
- `src/bin/chrome-devtools-cli-options.ts` — new CLI options entries
- `src/telemetry/tool_call_metrics.json` — new arg-type metrics entries

## Verification

```bash
npm install
npm run typecheck   # passes
npm run build       # passes
npm run gen         # regenerates docs/cli/metrics; format clean
npm run check-format  # passes
node scripts/test.mjs build/tests/tools/input.test.js  # all 46 input tests pass
```

## References

- `chrome-devtools-mcp` current drag implementation: `src/tools/input.ts`
- Puppeteer `ElementHandle.drag` non-intercepted branch: `node_modules/puppeteer-core/lib/puppeteer/api/ElementHandle.js`
- Zoho Analytics Pivot View repro discussed during a real automation session

### DIFF
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -81,6 +81,10 @@
 - **from_uid** (string) **(required)**: The uid of the element to [`drag`](#drag)
 - **to_uid** (string) **(required)**: The uid of the element to drop into
 - **includeSnapshot** (boolean) _(optional)_: Whether to include a snapshot in the response. Default is false.
+- **to_fraction_x** (number) _(optional)_: Optional x fraction within the target element bounding rect. 0 is the left edge and 1 is the right edge. Cannot be combined with to_offset_x.
+- **to_fraction_y** (number) _(optional)_: Optional y fraction within the target element bounding rect. 0 is the top edge and 1 is the bottom edge. Cannot be combined with to_offset_y.
+- **to_offset_x** (number) _(optional)_: Optional x offset in CSS pixels from the target element bounding rect top-left corner. Cannot be combined with to_fraction_x.
+- **to_offset_y** (number) _(optional)_: Optional y offset in CSS pixels from the target element bounding rect top-left corner. Cannot be combined with to_fraction_y.
 
 ---
 
--- a/src/bin/chrome-devtools-cli-options.ts
+++ b/src/bin/chrome-devtools-cli-options.ts
@@ -111,6 +111,34 @@ export const commands: Commands = {
         description: 'The uid of the element to drop into',
         required: true,
       },
+      to_offset_x: {
+        name: 'to_offset_x',
+        type: 'number',
+        description:
+          'Optional x offset in CSS pixels from the target element bounding rect top-left corner. Cannot be combined with to_fraction_x.',
+        required: false,
+      },
+      to_offset_y: {
+        name: 'to_offset_y',
+        type: 'number',
+        description:
+          'Optional y offset in CSS pixels from the target element bounding rect top-left corner. Cannot be combined with to_fraction_y.',
+        required: false,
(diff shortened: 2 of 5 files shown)


zyzyzyryxy:
Please sign in the CLA for us to review your PR, otherwise we'll need to close as stale.

Lightning00Blade:
Closing as stale.

PR #2165, opened 2026-06-01
refactor: change type of logger

Lightning00Blade:
This change aligns most of our logger allowing them to be omitted when needed meaning code inside the args will not be executed unconditionally - example `logger?.(JSON.stringy(X))`.

Currently only changes the type, while in a follow up I will remove the `debug` package and update the file logs.


zyzyzyryxy (approved):
LGTM in general, build should catch remaining missed calls.

PR #2166, opened 2026-06-01
feat: include page title in list_pages output

### LINKED ISSUE #2156: Include page title in `list_pages` output

`list_pages` shows each page's URL but not its title, which is painful when pages share a host — e.g. several accounts of the same product, all under `app.example.com/u/0/…`, `/u/1/…`, `/u/2/…`. There's no way to tell which is which without opening each, even though every page has a usable `document.title`.

The page id isn't a stable handle either — it's a counter that resets when the server reconnects to a fresh browser (e.g. the app I'm debugging restarts under `--browserUrl`), so pages come back renumbered and an id I grabbed a minute ago can point somewhere else.

Include the page title in `list_pages` output. Today the line (in `McpResponse.ts`) is:

```js
`${context.getPageId(page)}: ${page.url()}${selected ? ' [selected]' : ''}${contextLabel}`
```

`page.title()` is already available via Puppeteer, so `${id}: ${title} (${url})` would do it, plus a `title` field on the structured entry from `createStructuredPage`.

Matching by URL — works until multiple pages share a host (the case above). Caching the page id — unreliable across reconnects, as noted. Neither gives a stable, human-readable handle the way the title would.

For reference, Playwright MCP already lists titles — its `browser_tabs` renders each line as `[title](url)`.

### Additional context

I'm on `chrome-devtools-mcp@1.0.1`.

Related: #933, #1173

Happy to send a PR (read CONTRIBUTING, can sign the CLA) if you're open to it.

serhiizghama:
## Motivation

`list_pages` shows each page's URL but not its title, which is painful when multiple pages share a host — e.g. several tabs under `app.example.com/u/0/`, `/u/1/`, `/u/2/`. There's no way to tell which is which without visiting each one, even though every page has a usable `document.title`.

Fixes #2156.
Closes #2175

## What this changes

- `list_pages` text output now shows the title before the URL when available: `1: My Page (https://example.com) [selected]`. If the page has no title (e.g. `about:blank`), the format is unchanged.
- The structured content entry for each page now includes a `title` field alongside `id`, `url`, and `selected`.
- `page.title()` is awaited with a `.catch(() => '')` so a closed or erroring page silently falls back to the URL-only format.
- `format()` is made `async` to support the `await` inside the page loop; `createStructuredPage()` likewise becomes `async`.

## Testing

Start the MCP server with multiple tabs open. Call `list_pages` — pages with titles now display as `id: Title (url)`. Pages without titles (`about:blank`, data URLs) display as before.


src/McpResponse.ts
```diff
@@ -849,10 +810,12 @@ Call ${handleDialog.name} to handle it before continuing.`);
           const contextLabel = isolatedContextName
             ? ` isolatedContext=${isolatedContextName}`
             : '';
+          const title = await page.title().catch(() => '');
```
zyzyzyryxy: Please add a character limit as requested in https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2156#issuecomment-4610053690
It would be good to extract this to a separate function since you'll need to do this in three places.

zyzyzyryxy (changes requested):
Overall the change looks good, let's just add a limit to the title (50 characters should be good imo) and address the issues reported by the CI (formatting, failing tests)

serhiizghama:
Thanks for the review! Pushed the fixes:

- added a `truncateTitle()` helper that caps titles at 50 chars with an ellipsis
- fixed the formatting failure and updated the pages test snapshot
- title fetching now races `page.title()` against a 1s timeout so `list_pages` can't hang while a dialog is open (a showing dialog blocks script evaluation, including the isolated realm)

The remaining CI failures come from the new webmcp snapshot tests on main that predate the title output — I'll update the branch with main and the regenerated snapshots in a follow-up push.

zyzyzyryxy (approved):

PR #2168, opened 2026-06-02
feat: Handle multiple providers of third-party developer tools

wolfib:
This allows a page to have multiple providers of third-party developer tools, which each respond to the `devtoolstooldiscovery` event.

- Multiple `ToolGroup`s
- MCP tool responses only mention third-party developer tools, if there are any. Otherwise this part of the output is skipped.

### DIFF
--- a/src/McpPage.ts
+++ b/src/McpPage.ts
@@ -14,7 +14,7 @@ import type {
   WebMCPTool,
 } from './third_party/index.js';
 import {takeSnapshot} from './tools/snapshot.js';
-import type {ToolGroup, ToolDefinition} from './tools/thirdPartyDeveloper.js';
+import type {ToolGroups} from './tools/thirdPartyDeveloper.js';
 import type {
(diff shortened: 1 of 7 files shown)


src/McpPage.ts
```diff
@@ -59,7 +59,7 @@ export class McpPage implements ContextPage {
   #dialog?: Dialog;
   #dialogHandler: (dialog: Dialog) => void;
 
-  thirdPartyDeveloperTools: ToolGroup<ToolDefinition> | undefined;
+  thirdPartyDeveloperTools: Array<ToolGroup<ToolDefinition>> = [];
```
Lightning00Blade: Nit: Maybe we can extract a type as I see it used very often.

Lightning00Blade (approved):
LGTM

src/McpPage.ts
wolfib: Done

PR #2169, opened 2026-06-02
feat: memory debugging tools

Lightning00Blade:
Updates the flag for the memory tooling to remove the experimental bit (keep alias for backwards compatibility).
And updates the SKILLs to reflected the update names and point to the available tools.

Q: Should it be called `take_heapsnapshot` or `take_heap_snapshot`?


nroscino (approved):
LGTM!

PR #2174, opened 2026-06-03
feat: Adds close_heapsnapshot MCP tool

dinfuehr:
This commit adds the close_heapsnapshot MCP tool such that the coding agent can close heap snapshots again.


src/tools/memory.ts
```diff
@@ -154,3 +154,32 @@ export const getHeapSnapshotRetainers = defineTool({
...
   },
 });
+
+export const closeHeapSnapshot = defineTool({
+  name: 'close_heapsnapshot',
+  description:
+    'Closes a previously loaded memory heapsnapshot, freeing its memory.',
+  annotations: {
+    category: ToolCategory.MEMORY,
+    readOnlyHint: false,
```
dinfuehr, line 164: I am not really sure about this. I guess depending on how one interprets that flag, it might be considered `true` as well.

Quiroz2020 (approved):

src/tools/memory.ts
Lightning00Blade: This is where the spec defines it - https://modelcontextprotocol.io/specification/2025-11-25/schema#toolannotations
Yeah maybe it's more of a `destructiveHint` not sure personally, we can leave it as is for now.

Lightning00Blade (approved):
LGTM, thanks!

PR #2175, opened 2026-06-04
fix: include page titles in list_pages

### LINKED ISSUE #2156: Include page title in `list_pages` output

Sushanth012:
## Summary
- include each page title in list_pages text output as id: title (url)
- add title to structured page entries
- use target metadata so list_pages still works while dialogs are open

Fixes #2156

## Testing
- npm run test:update-snapshots -- tests/McpResponse.test.ts tests/tools/pages.test.ts
- npm run test -- tests/McpResponse.test.ts tests/tools/pages.test.ts
- npm run format

### DIFF
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -879,14 +879,15 @@ Call ${handleDialog.name} to handle it before continuing.`);
         const parts = [`## Pages`];
         const structuredPages = [];
         for (const page of regularPages) {
+          const title = getPageTitle(page);
           const isolatedContextName = context.getIsolatedContextName(page);
           const contextLabel = isolatedContextName
             ? ` isolatedContext=${isolatedContextName}`
             : '';
           parts.push(
-            `${context.getPageId(page)}: ${page.url()}${context.isPageSelected(page) ? ' [selected]' : ''}${contextLabel}`,
+            `${context.getPageId(page)}: ${title} (${page.url()})${context.isPageSelected(page) ? ' [selected]' : ''}${contextLabel}`,
           );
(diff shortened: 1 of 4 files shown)


nroscino:
Hey @Sushanth012 ! Thank you for the PR! 
Can you please sign the CLA before we proceed with the review? 

Thank you!

PR #2187, opened 2026-06-08
feat: Add retaining paths MCP tool

dinfuehr:
This PR adds the `get_heapsnapshot_retaining_paths` MCP tool. This can be used to find the paths from the target object to the GC roots which keep that object alive.


dinfuehr:
Note that I added this as a separate MCP tool. We've discussed merging them but get_heapsnapshot_retaining_paths only prints paths to the GC roots. But it doesn't print all retainers (=incoming edges) of an object like get_heapsnapshot_retainers does.

src/McpResponse.ts
```diff
@@ -1070,6 +1068,26 @@ Call ${handleDialog.name} to handle it before continuing.`);
 
         structuredContent.heapSnapshotNodes = paginationData.items;
       }
+      const retainingPaths = this.#heapSnapshotOptions.retainingPaths;
+      if (retainingPaths) {
+        response.push('### Retaining Paths');
```
Lightning00Blade, line 1073: I am curious if it make sense to provide a simple explanation of structure.
Once we have a proper benchmarks I think this would be worth trying.

Lightning00Blade (approved):
LGTM

PR #2198, opened 2026-06-11
feat: Add MCP tools for heap snapshot comparison

dinfuehr:
This commit adds two MCP tools for comparing heap snapshots. `compare_heapsnapshot_summary` compares two memory snapshot and returns which classes have new/deleted objects. `compare_heapsnapshot_class_nodes` can then be used to list the object ids added and deleted for a specific class.


src/HeapSnapshotManager.ts
```diff
@@ -173,6 +195,56 @@ export class HeapSnapshotManager {
...
+      baseFilePath,
+      currentFilePath,
+    );
+    const result = classDiffs[classIndex];
+    if (!result) {
+      throw new Error(
+        `Invalid classIndex: ${classIndex}. Total classes with changes: ${classDiffs.length}`,
+      );
+    }
+    const rawDiff = result;
```
Lightning00Blade: Nit, not needed reassign

dinfuehr:
I've rebased the PR again. Should be ready for review again.

src/HeapSnapshotManager.ts
```diff
@@ -182,6 +254,33 @@ export class HeapSnapshotManager {
...
+
+    const interfaceDefinitions = await currentSnapshot.interfaceDefinitions();
+    const aggregatesForDiff =
+      await baseSnapshot.aggregatesForDiff(interfaceDefinitions);
+    const cachedBaseSnapshot = this.#getCachedSnapshot(baseFilePath);
+    // DevTools calculateSnapshotDiff uses the first parameter (baseSnapshotId)
+    // as a cache key. We pass the unique diffCacheKey we generated when loading
+    // the base snapshot.
+    const rawDiffs = await currentSnapshot.calculateSnapshotDiff(
+      cachedBaseSnapshot.diffCacheKey,
```
Lightning00Blade: I think I overlook the UID we provide as at time it did not matter and it's currently hard-coded for all to `1` (workerProxy.createLoader).
Let's update that part and pull the value from the uid like DevTools done.

Lightning00Blade (approved):
LGTM, with 2 small comments.

src/HeapSnapshotManager.ts
dinfuehr: Thanks for your suggestion! This indeed makes a lot of sense. I believe I've addressed your comment. But it is delicate enough, that you might want to take another look.

dinfuehr:
Sorry, I missed that you already reviewed this PR. I've addressed your comments now. PTALA

Lightning00Blade (approved):
LGTM

dinfuehr:
Thanks for the reviews!

PR #2200, opened 2026-06-11
fix: Reset toolGroups before gathering toolGroups

wolfib:
`getToolGroups` can be called multiple times. We want to reset the `window.__dtmcp.toolGroups` array each time to prevent duplications.


Lightning00Blade (approved):

kassebasse3-ai (commented):
src/McpResponse.ts

PR #2202, opened 2026-06-12
Update snapshot tests for dialog and tool execution

kkm9wd6vzj-stack:
Updated snapshot tests to reflect changes in dialog error messages and tool execution outputs.


Lightning00Blade:
Given that the test pass on main I think this is not needed.

PR #2203, opened 2026-06-12
revert: support experimental allowlist for navigate tool calls (#1935)

Lightning00Blade:
This reverts commit d50255778a6dcb30fb702755f0dd38f8ee2cd858.


OrKoN:
Thanks for the PR. Is there more context?

Lightning00Blade:
With the introduction of https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2037
We can rely on that functionality rather then this.

OrKoN:
> With the introduction of #2037 We can rely on that functionality rather then this.

I am not sure we can because that would require restarting the MCP server so this feature might be still in use. I would need to confirm separately if it is still in use.

OrKoN:
Let's land this. I have confirmed it's not in use.

OrKoN (approved):

PR #2207, opened 2026-06-13
fix: handle screencast file extensions case-insensitively and clean up temp dir on failure

### LINKED ISSUE #2206: screencast_start silently changes format and path for uppercase/unsupported file extensions
### Summary

`screencast_start` validates the requested file extension with a **case-sensitive** `endsWith()` against `['.webm', '.mp4']` and **silently falls back to `.mp4`** when nothing matches, instead of erroring. Combined with `ensureExtension()` (which *replaces* the existing extension), the recording can be written in a **different format** and to a **different path** than requested, with no warning.

(Reported originally by @ch-bas while testing; filing upstream with a reproduction and a fix.)

### Affected code

- `src/tools/screencast.ts` — the `supportedExtensions` loop uses `filePath.endsWith(supportedExtension)` (case-sensitive) with a silent `.mp4` default.
- `src/utils/files.ts` — `ensureExtension()` strips the existing extension via `path.extname` and appends the enforced one.

Present in `v1.2.0` and on current `main` (the two files are unchanged between them).

### Steps to reproduce

1. Start the server with `--experimentalScreencast=true` (and `--experimentalFfmpegPath <ffmpeg>`).
2. `screencast_start` with `filePath: "demo.WEBM"`.
3. `screencast_stop`.

### Actual

Records **MP4** to **`demo.mp4`** — neither the requested format nor the requested path. The same happens for any unsupported extension, e.g. `recording.avi` silently becomes `recording.mp4`.

Reproduced with the exact `endsWith` loop + `ensureExtension` run through `node:path`:

| requested      | resolved path     | format |             |
| -------------- | ----------------- | ------ | ----------- |
| `demo.webm`    | `demo.webm`       | webm   | ok          |
| `demo.WEBM`    | `demo.mp4`        | mp4    | wrong       |
| `demo.Webm`    | `demo.mp4`        | mp4    | wrong       |
| `recording.avi`| `recording.mp4`   | mp4    | wrong       |
| `clip.MP4`     | `clip.mp4`        | mp4    | silently renamed |

### Expected

- `.WEBM` / `.Webm` etc. treated as WebM (case-insensitive match).
- An explicitly requested but unsupported extension rejected with an explicit error listing the supported formats, rather than silently switching format and path.

### Minor related issue

When `screencast_start` is called **without** a `filePath`, it creates a temp directory via `mkdtemp()`. If `page.screencast()` then throws (e.g. ffmpeg missing), that directory is never cleaned up, leaking an empty temp dir per failed attempt.

### Fix

PR incoming that (1) matches the extension case-insensitively, (2) rejects unsupported explicit extensions with a clear error, and (3) removes the generated temp directory when recording fails to start. Includes regression tests using the existing screencast test harness.

### Environment

`chrome-devtools-mcp` v1.2.0 / current `main`.

nebrass:
Fixes #2206

### Problem

`screencast_start` matched the requested file extension with a **case-sensitive** `endsWith()` against `['.webm', '.mp4']` and **silently fell back to `.mp4`** when nothing matched. Combined with `ensureExtension()` (which replaces the extension), a request for `demo.WEBM` was recorded as **MP4** to **`demo.mp4`** — a different format *and* path than requested — and any unsupported extension (e.g. `recording.avi`) silently became `.mp4`.

Separately, when `screencast_start` is called without a `filePath`, it creates a temp directory via `mkdtemp()`. If `page.screencast()` then throws (e.g. ffmpeg missing), that directory was leaked.

### Changes

Two commits:

1. **`fix: match screencast extension case-insensitively and reject unsupported ones`** — match via `path.extname().toLowerCase()`; reject an explicitly requested but unsupported extension with an explicit error listing the supported formats; a missing extension still defaults to `.mp4`.
2. **`fix: clean up screencast temp directory when recording fails to start`** — remove the generated temp dir in the `catch` handler, but only when we own the generated path (never when the caller supplied `filePath`).

| requested       | before          | after          |
| --------------- | --------------- | -------------- |
| `demo.WEBM`     | mp4 → `demo.mp4`| webm → `demo.webm` |
| `recording.avi` | mp4 → `recording.mp4` | error (rejected) |
| `demo.webm`     | webm → `demo.webm` | unchanged |
| *(no filePath)* | mp4 temp        | unchanged      |

The matched extension is normalized to lower case (`demo.WEBM` → `demo.webm`).

### Testing

Added three regression tests to `tests/tools/screencast.test.ts` using the existing `sinon`/`withMcpContext` harness. Verified locally against Chrome for Testing 149 (`PUPPETEER_EXECUTABLE_PATH`):

- With the fix reverted, the two extension tests fail (uppercase `.WEBM` → mp4, `.avi` not rejected) and the cleanup test fails (temp dir left behind) — i.e. they fail for the right reason.
- With the fix applied, the full `screencast.test.ts` suite passes (11/11).
- `tsc --noEmit` and `npm run check-format` (eslint + prettier) are clean.

> Note: I ran the `screencast` test file (which stubs `page.screencast`) plus typecheck/lint locally; the rest of the browser-based suite I left to CI.

### Notes for reviewers

- I chose to **`throw`** for an unsupported explicit extension (consistent with the ffmpeg-missing `throw` in the same handler and with the issue's "reject with an explicit error"). Happy to switch to the softer `appendResponseLine(...) + return` style used by the in-progress guard if you'd prefer.
- The two commits are independent and can be split if you'd rather take them separately.
- I left the pre-existing `as \`${string}.webm\`` assertion on `resolvedPath` untouched to keep the diff focused, though it's slightly misleading now that the default is `.mp4`.

### DIFF
--- a/src/tools/screencast.ts
+++ b/src/tools/screencast.ts
@@ -87,6 +87,18 @@ export const startScreencast = definePageTool(args => ({
         ffmpegPath: args?.experimentalFfmpegPath,
       });
     } catch (err) {
+      // If we generated a temporary directory for this recording, remove it so
+      // a failed start (e.g. ffmpeg missing) does not leak an empty directory.
+      if (requestedFilePath === undefined) {
+        try {
+          await fs.rm(path.dirname(resolvedPath), {
+            recursive: true,
+            force: true,
+          });
+        } catch {
+          // no-op
+        }
+      }
       const message = err instanceof Error ? err.message : String(err);
       if (message.includes('ENOENT') && message.includes('ffmpeg')) {
         throw new Error(
--- a/src/tools/screencast.ts
+++ b/src/tools/screencast.ts
@@ -48,17 +48,29 @@ export const startScreencast = definePageTool(args => ({
       return;
     }
 
-    const filePath = request.params.filePath ?? (await generateTempFilePath());
-    let enforcedExtension = '.mp4' as `.${string}`;
-    let format: VideoFormat = 'mp4';
+    const requestedFilePath = request.params.filePath;
+    const filePath = requestedFilePath ?? (await generateTempFilePath());
 
-    for (const supportedExtension of supportedExtensions) {
-      if (filePath.endsWith(supportedExtension)) {
-        enforcedExtension = supportedExtension;
-        format = supportedExtension.substring(1) as VideoFormat;
-        break;
-      }
+    // Match the extension case-insensitively so e.g. `.WEBM` is recognized as
+    // WebM. An explicitly requested but unsupported extension is rejected
+    // rather than being silently rewritten to `.mp4` (which would change both
+    // the format and the output path from what was requested). A missing
+    // extension falls back to `.mp4`. The matched extension is normalized to
+    // lower case.
+    const requestedExtension = path.extname(filePath);
+    const matchedExtension = supportedExtensions.find(
+      supportedExtension =>
+        supportedExtension === requestedExtension.toLowerCase(),
+    );
+    if (!matchedExtension && requestedExtension !== '') {
+      throw new Error(
+        `Unsupported screencast file extension "${requestedExtension}". ` +
+          `Supported formats: ${supportedExtensions.join(', ')} (case-insensitive).`,
+      );
     }
+    const enforcedExtension: `.${string}` = matchedExtension ?? '.mp4';
+    const format: VideoFormat = (matchedExtension?.substring(1) ??
+      'mp4') as VideoFormat;
 
     const resolvedPath = ensureExtension(
       path.resolve(filePath),
--- a/tests/tools/screencast.test.ts
+++ b/tests/tools/screencast.test.ts
@@ -5,6 +5,8 @@
  */
 
 import assert from 'node:assert';
+import fs from 'node:fs/promises';
+import path from 'node:path';
 import {describe, it, afterEach} from 'node:test';
 
 import sinon from 'sinon';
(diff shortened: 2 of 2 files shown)


nebrass:
### CI failure looks like a pre-existing Windows flake, not from this PR

The only red cell is **Tests on windows-latest with node 24**. It fails in `tests/tools/screenshot.test.ts` (a file this PR doesn't touch) during browser setup:

```
✖ with unwritable filePath   (screenshot.test.ts:142)
✖ with malformed filePath    (screenshot.test.ts:181)
  ConnectionClosedError: Connection closed.
    at CdpBrowser.newPage (puppeteer-core/.../Browser.js:199)
    at async withBrowser (build/tests/utils.js:64)
```

The CDP connection dropped on `newPage`. (The `process ... could not be terminated` lines are just Puppeteer's Windows browser-tree cleanup — harmless.)

Evidence it's unrelated to this change:

- **Sibling cells passed.** Windows **node 22 and node 26 passed the same tests in the same run** — only node 24 flaked. A real bug would break all three.
- **`main` flakes the same way.** Recent `main` runs fail this workflow Windows-only with the same class of CDP instability (e.g. `extensions.test.js`: `ProtocolError (Target.attachToTarget): No target with given id found`).
- **This PR is scoped + safe.** It only changes `screencast.ts`/`screencast.test.ts`. The cleanup `fs.rm(...)` removes only the unique `mkdtemp` subdir from `generateTempFilePath()`, never `os.tmpdir()`; no `browser.close()`/disconnect; sinon stubs restored in `afterEach`.
- **Local repro is green.** Running `screencast.test.ts` → `screenshot.test.ts` back-to-back in one process: **21/21 pass**, including the two tests that failed in CI.

Could a maintainer please **re-run the failed job**? I don't have re-run permissions on the repo. (Side note: the CI step runs `npm run test:no-build` without the `--retry` flag that `scripts/test.mjs` already supports, so a single transient browser hiccup fails the whole required job — enabling `--retry` in CI would cut down these spurious failures.)

nroscino (approved):
LGTM, Thank you!

PR #2209, opened 2026-06-13
fix: return error message when screencast_stop is called with no active recording

ch-bas:
## Summary

`screencast_stop` returns an empty response when no recording is active, making it impossible for the calling agent to distinguish "stopped successfully" from "nothing was recording."

`screencast_start` already handles its inverse case with an explicit error (`"a screencast recording is already in progress"`), so this makes `stop` consistent.

## Change

Added an error message when `screencast_stop` is called without an active recording:

```ts
if (!data) {
  response.appendResponseLine(
    'Error: no active screencast recording to stop.',
  );
  return;
}
```

## Before

Empty tool response — agent cannot tell what happened.

## After

`Error: no active screencast recording to stop.`

### DIFF
--- a/src/tools/screencast.ts
+++ b/src/tools/screencast.ts
@@ -107,6 +107,9 @@ export const stopScreencast = definePageTool({
   handler: async (_request, response, context) => {
     const data = context.getScreenRecorder();
     if (!data) {
+      response.appendResponseLine(
+        'Error: no active screencast recording to stop.',
+      );
       return;
     }
     try {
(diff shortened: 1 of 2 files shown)


ch-bas:
@googlebot I signed it!

nroscino:
Hello @ch-bas! 

Thank you for submitting the PR. The change looks fine but test must also be updated to ensure CI passes correctly. Can you handle this?

ch-bas:
Hi @nroscino , I have updated the tests as well. Could you check it now? Thanks!

nroscino (approved):
LGTM! Thank you!

PR #2212, opened 2026-06-14
fix: clarify evaluate_script selector guidance

### LINKED ISSUE #1892: The agent attempts to use MCP-specific uids (from accessibility tree snapshots) as if they were standard HTML attributes or valid element references in evaluate_script.
### Description of the bug

Example of attempted tool (mis)use:
```
{
  "name": "mcp_chrome-devtools_evaluate_script",
  "args": { "function": "() => document.querySelector('[uid=\"3_125\"]').innerText" }
}
```

Observed once after activation of `chrome-devtools` skill and once without prior skill activation.

In both cases, the agent immediately realized uid is not a DOM attribute and switched to standard CSS selectors.

### Reproduction

1. Use `take_snapshot`
2. Use `evaluate_script` (for example to query innerText of an element)

### Expectation

Agent should be informed by one of the tool descriptions or by the skill that `uid`s are not part of the DOM and cannot be used in query selectors.

### MCP configuration

any

0.20.1

### Chrome version

_No response_

### Coding agent version

gemini-cli

### Model version

gemini-2.5-flash

### Chat log

_No response_

### Node version

_No response_

### Operating system

None

### Extra checklist

### LINKED ISSUE #1895: The agent uses library-specific pseudo-classes (like :contains) that are invalid in native querySelector.
### Description of the bug

Example tool (mis)use:
```
{
  "name": "mcp_chrome-devtools_evaluate_script",
  "args": { "function": "() => document.querySelector('th:contains(\"UPC\") + td').innerText" }
}
```

This happened without prior `chrome-devtools` skill activation.

### Reproduction

Prompt the agent to search for specific information on a page for which it has high confidence of knowing the exact structure (and thus attempting to make targeted read)

### Expectation

Tool description should provide more detailed description for the context scripts will be evaluated in.

### MCP configuration

any

v0.20.1

### Chrome version

_No response_

### Coding agent version

gemini-cli

### Model version

gemini-2.5-flash

### Chat log

_No response_

### Node version

_No response_

### Operating system

None

### Extra checklist

luohui1:
## Summary
- clarify that evaluate_script runs with standard browser DOM APIs and CSS selectors
- explain that snapshot uids should be passed through args rather than queried as DOM attributes
- add an eval scenario covering UPC table lookup without :contains or uid/ref attribute selectors

Fixes #1892
Fixes #1895

## Tests
- npm run build
- npm run docs:generate
- npm run cli:generate
- npm run format
- npm exec eslint -- --cache .
- npm exec prettier -- --check --cache .
- $env:PUPPETEER_EXECUTABLE_PATH='C:\Program Files\Google\Chrome\Application\chrome.exe'; npm run test tests\tools\script.test.ts

Note: npm run check-format passes ESLint and Prettier on this Windows checkout, then exits non-zero because the script's trailing `.;` is parsed as an extra Prettier path.

### DIFF
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -345,11 +345,11 @@
 ### `evaluate_script`
 
 **Description:** Evaluate a JavaScript function inside the currently selected page. Returns the response as JSON,
-so returned values have to be JSON-serializable.
+so returned values have to be JSON-serializable. Use standard browser DOM APIs. Snapshot uids are not DOM attributes; to work with a snapshot element, pass its uid through args instead of using it in document.querySelector(). querySelector only accepts standard CSS selectors, not jQuery-style pseudo-classes such as :contains().
 
 **Parameters:**
 
-- **function** (string) **(required)**: A JavaScript function declaration to be executed by the tool in the currently selected page.
+- **function** (string) **(required)**: A JavaScript function declaration to be executed by the tool in the currently selected page. Use standard browser DOM APIs and selectors.
   Example without arguments: `() => {
   return document.title
 }` or `async () => {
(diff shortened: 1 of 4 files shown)


zyzyzyryxy:
We no longer observe those issues with newer models, both issues were closed. Thank you for the effort though!

PR #2221, opened 2026-06-16
fix(network): keep redirect chain order consistent between text and JSON

nyxst4ck:
## Problem

`get_network_request` returns, from the **same call**, both a human-readable
text block (`toStringDetailed()`) and a `structuredContent.networkRequest`
object (`toJSONDetailed()`) — emitted together in `McpResponse.ts`. For a
request that went through HTTP redirects, the redirect chain comes out in
**opposite orders** in the two representations:

- `toJSONDetailed()` (`NetworkFormatter.ts`) reverses `redirectChain()` once →
  newest→oldest in the JSON.
- the text formatter then reverses it **a second time** → oldest→newest in the
  text.

Because each path calls `redirectChain()` separately and Puppeteer returns a
fresh copy on every call (`HTTPRequest#redirectChain()` does
`this._redirectChain.slice()`), the two reverses operate on different arrays and
don't cancel. So a consumer reading the text and a consumer parsing the
structured JSON from the same response see contradictory redirect orders.

## Solution

Drop the redundant `.reverse()` in the text formatter so the rendered text uses
the order already produced by `toJSONDetailed()`. Both representations are now
consistent (newest→oldest), and `structuredContent` is unchanged.

## Why the existing tests didn't catch it

- The existing "handles redirect chain" test uses a **single-element** chain,
  where reversing is a no-op.
- `getMockRequest().redirectChain()` returned the **same array reference** on
  every call, unlike real Puppeteer — so the two reverses accidentally agreed in
  tests. This PR makes the mock return a fresh copy per call (matching Puppeteer)
  and adds a regression test with a multi-element chain that asserts the text and
  JSON orders match.

## Testing

- `npm test` for the formatter suite passes. New test
  `renders the redirect chain in the same order in text and JSON` is **red**
  before the fix (text `[first, second]` vs JSON `[second, first]`) and **green**
  after, with no change to existing snapshots.
- `npm run typecheck` and Prettier/ESLint are clean.

No existing issue tracked this; found via code inspection and confirmed
empirically.

### DIFF
--- a/src/formatters/NetworkFormatter.ts
+++ b/src/formatters/NetworkFormatter.ts
@@ -314,7 +314,9 @@ function converNetworkRequestDetailedToStringDetailed(
   if (redirectChain?.length) {
     response.push(`### Redirect chain`);
     let indent = 0;
-    for (const request of redirectChain.reverse()) {
+    // `redirectChain` is already ordered by toJSONDetailed(); don't reverse it
+    // again here or the text output contradicts structuredContent (the JSON).
+    for (const request of redirectChain) {
       response.push(
         `${'  '.repeat(indent)}${convertNetworkRequestConciseToString(request)}`,
       );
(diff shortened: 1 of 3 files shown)


nroscino (approved):
LGTM, thank you!

PR #2231, opened 2026-06-19
fix: hide Windows update check consoles

### LINKED ISSUE #2230: Windows: daily update check flashes a cmd.exe console window (npm config get registry spawned from a detached process)
### Description of the bug

On Windows, the once-per-day update check briefly flashes a `cmd.exe` console window on the desktop, unrelated to any user action. It appears on the first MCP-server startup after `~/.cache/chrome-devtools-mcp/latest.json` becomes stale (>24h), so it looks like a random ~once-a-day flash.

**Root cause** — two cooperating spots, both missing `windowsHide`:

1. `src/utils/check-for-updates.ts` spawns the version checker with `{ detached: true, stdio: 'ignore' }`. On Windows `detached: true` maps to `DETACHED_PROCESS`, so the child process has **no console**.
2. `src/bin/check-latest-version.ts` → `getRegistry()` runs `execSync('npm config get registry', …)`. `npm` on Windows is a `.cmd` batch wrapper, which **requires** a console — and because the parent has none, Windows allocates a **new visible** console window for it. That window is the flash.

### Regression

Introduced by #2038 (commit `83a299fc`, 2026-05-12), which replaced the hardcoded `fetch('https://registry.npmjs.org/…')` with a shell-out to `npm config get registry`. Before that change the detached process only performed a `fetch()` and produced no console window. (The detached update check itself dates to #1783.)

The bug is present in 1.3.0 and still on `main` (`5a9d6af`) — both `check-for-updates.ts` and `check-latest-version.ts` lack `windowsHide`.

### Reproduction

1. Windows, `chrome-devtools-mcp` started via `npx` (e.g. from an MCP client).
2. Delete or age `%USERPROFILE%\.cache\chrome-devtools-mcp\latest.json` so it is >24h old (or missing).
3. Start the MCP server → a `cmd.exe` window flashes.

Direct one-liner that reproduces the exact flash on demand (run with the same Node that launches the server; replace `<pkg>` with the installed package dir):

```
node -e "const cp=require('child_process'),os=require('os'),p=require('path');cp.spawn(process.execPath,['<pkg>/build/src/bin/check-latest-version.js',p.join(os.tmpdir(),'x.json')],{detached:true,stdio:'ignore'}).unref()"
```

Captured process chain at the moment the window appears:

```
cmd.exe /d /s /c "npm config get registry"
  +- node ...\chrome-devtools-mcp\build\src\bin\check-latest-version.js ...\latest.json
     +- node ...\chrome-devtools-mcp\build\src\bin\chrome-devtools-mcp.js --browserUrl ...
```

### Suggested fix

Add `windowsHide: true` to both child-process calls:

- the `child_process.spawn(...)` in `check-for-updates.ts`, and
- the `execSync('npm config get registry', …)` in `check-latest-version.ts`.

The `execSync` one is what directly suppresses the visible `npm` console; adding it to the `spawn` as well is belt-and-suspenders.

### Workaround

Set `CHROME_DEVTOOLS_MCP_NO_UPDATE_CHECKS=1` to disable the check entirely (loses update notifications).

1.3.0 (also present on `main` @ `5a9d6af`)

### Chrome version

n/a (occurs at server startup, before any browser interaction)

### Node version

24.17.0

### Operating system

Windows 10 (19045)

cyphercodes:
## Summary

Fixes #2230.

- Sets `windowsHide: true` on the detached update-check process so Windows does not flash a console window.
- Also hides the `npm config get registry` subprocess used by the updater helper.
- Covers the update-check spawn options in the existing tests.

## Verification

- `NODE_OPTIONS=--max-old-space-size=4096 npm run build`
- `npm run test:no-build -- tests/check-for-updates.test.ts`
- `NODE_OPTIONS=--max-old-space-size=4096 npx eslint src/utils/check-for-updates.ts src/bin/check-latest-version.ts tests/check-for-updates.test.ts`
- `npx prettier --check src/utils/check-for-updates.ts src/bin/check-latest-version.ts tests/check-for-updates.test.ts`
- `git diff --check`

Note: plain `npm run build` and full-repo `npm run check-format` hit the local Node heap limit in this runner; the same build passed with the heap limit raised, and touched-file lint/format checks passed.

### DIFF
--- a/src/bin/check-latest-version.ts
+++ b/src/bin/check-latest-version.ts
@@ -21,6 +21,7 @@ function getRegistry(): string {
       encoding: 'utf8',
       stdio: ['ignore', 'pipe', 'ignore'],
       timeout: 5000,
+      windowsHide: true,
     })
       .trim()
       .replace(/\/$/, '');
--- a/src/utils/check-for-updates.ts
+++ b/src/utils/check-for-updates.ts
@@ -88,6 +88,7 @@ export async function checkForUpdates(message: string) {
       {
         detached: true,
         stdio: 'ignore',
+        windowsHide: true,
       },
     );
     child.unref();
--- a/tests/check-for-updates.test.ts
+++ b/tests/check-for-updates.test.ts
@@ -139,6 +139,7 @@ describe('checkForUpdates', () => {
     );
     assert.ok(spawnStub.firstCall.args[1][1]?.includes('latest.json'));
     assert.strictEqual(spawnStub.firstCall.args[2]?.detached, true);
+    assert.strictEqual(spawnStub.firstCall.args[2]?.windowsHide, true);
     assert.ok(unrefSpy.calledOnce);
   });
 
@@ -163,6 +164,7 @@ describe('checkForUpdates', () => {
     );
     assert.ok(spawnStub.firstCall.args[1][1]?.includes('latest.json'));
     assert.strictEqual(spawnStub.firstCall.args[2]?.detached, true);
+    assert.strictEqual(spawnStub.firstCall.args[2]?.windowsHide, true);
     assert.ok(unrefSpy.calledOnce);
   });
 });


OrKoN (approved):
Thanks!

PR #2236, opened 2026-06-20
fix(cli): apply headless default on implicit daemon auto-start

inwaar:
The `chrome-devtools start` command handler defaults `headless: true` when not explicitly provided, but the implicit auto-start path (triggered transparently when running a tool command like `new_page` with no running daemon) bypassed this logic entirely. This meant the daemon would launch in headful mode by default when auto-started from a tool command, causing failures in environments without an X server (e.g., devcontainers, CI).

This PR extracts a shared `getDefaultedStartArgs()` helper that applies the same defaults (`headless: true`, `isolated: true`) for both the explicit `start` command and the implicit auto-start path.

Related: the `start` command handler also had a `// Defaults but we do not want to affect the yargs conflict resolution.` comment explaining why defaults are applied after yargs parsing rather than via the option definitions — this behavior is preserved in the extracted helper.

### DIFF
--- a/src/bin/chrome-devtools.ts
+++ b/src/bin/chrome-devtools.ts
@@ -39,6 +39,18 @@ async function start(args: string[], sessionId: string) {
 
 const defaultArgs = ['--viaCli', '--experimentalStructuredContent'];
 
+function getDefaultedStartArgs(
+  argv: Record<string, unknown>,
+): string[] {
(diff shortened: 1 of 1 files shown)


Lightning00Blade (commented):
Looks good, can we add a test and fix the error from CI about the docs (`npm run gen` should fix them).

OrKoN:
closing as stale

Issue #1896, opened 2026-04-17
The agent fails to correctly interface with the chrome-devtools binary, inventing flags or failing to escape characters for the shell.


zyzyzyryxy:
### Description of the bug

Agent can misunderstand how to use CLI interface properly and struggle with finding a working solution.
In one session I encountered 6 consecutive failed attempts to properly call the tool:

   1. Attempt 1 (Bash Syntax Error):
       * Tool Call: chrome-devtools evaluate_script --expression '''const mainElement = document.querySelector('main'); ... return oldestDate.toLocaleDateString(...);'''
       * Error: /usr/bin/bash: line 2: span)).filter(span => { ... No such file or directory
       * Reason: The shell tried to execute the JavaScript code as a bash script because of unescaped parentheses and line breaks.

   2. Attempt 2 (Missing Positional Argument):
       * Tool Call: chrome-devtools evaluate_script --expression '''const mainElement = document.querySelector('main'); ... return oldestDate.toLocaleDateString(...);'''
       * Error: Not enough non-option arguments: got 0, need at least 1
       * Reason: The CLI parser expected the script as a positional argument; it treated the code following --expression as a value for an unrecognized flag.

   3. Attempt 3 (Missing Positional Argument):
       * Tool Call: chrome-devtools evaluate_script --expression "(() => { const dates = []; ... return oldestDate.toLocaleDateString(...); })();"
       * Error: Not enough non-option arguments: got 0, need at least 1
       * Reason: Similar to Attempt 2, but using double quotes. The parser still failed to identify the required positional argument.

   4. Attempt 4 (Unknown CLI Flag):
       * Tool Call: chrome-devtools evaluate_script --expression '''(() => { ... })();'''
       * Error: Unknown arguments: expression, found.;
       * Reason: The CLI explicitly rejected the invented --expression flag.

   5. Attempt 5 (JavaScript Syntax Error via Shell):
       * Tool Call: chrome-devtools evaluate_script "(() => { const dates = []; ... return oldestDate.getTime(); } else { return -1; } })();"
       * Error: Unexpected token ';'
       * Reason: The flag was removed, but the shell stripped or misinterpreted characters within the double-quoted string, resulting in invalid JavaScript being delivered to the browser.

   6. Attempt 6 (JavaScript Syntax Error via Shell):
       * Tool Call: chrome-devtools evaluate_script "(() => { const dateStrings = []; ... return dateStrings; })();"
[...]

PR #2244, opened 2026-06-22
fix: Evaluate script error messages

syedaly:
Add helpful error messages to evaluate_script CLI to prevent agents from repeatedly failing with wrong flags or missing arguments. 
https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1896


Lightning00Blade:
Thanks for the PR, but the real issue here is testing the new error, because of that we would prefer the team handle issues labeled with evals.

PR #2250, opened 2026-06-24
feat: expose screencast first frame timestamp

### LINKED ISSUE #2216: Expose first-frame timestamp from screencast_start

I am trying to annoate the video with specific timestamps that are accurate.

I'm building an annotation layer on top of screencast. Overlaying timed markers (pass/fail, click highlights) on the recorded video. The challenge is syncing annotation timestamps to video time.

`screencast_start` is async. By the time the MCP response comes back, the video has already been recording for 1-5s. So I can't reliably compute "second 0" of the video from the caller's side. Setting t0 before the call is too early, after the response is too late.

Looking at the source, Puppeteer's ScreenRecorder subscribes to Page.screencastFrame events which include metadata.timestamp. It uses these internally for frame duplication but never exposes them.

Ask: Could screencast_start's response include the wall-clock time (or CDP timestamp) of the first frame? Something like:

```json
{
  "content": [{ "text": "Recording started...", "firstFrameTimestamp": 1718488800.123 }]
}
```

This would let callers precisely correlate external events to video time without needing a direct CDP connection.

I built my own recorder that gets added to the toolset. Currently instead of using screencast_start and screencast_stop, I have a proxy mcp that my agent calls which then wraps the chrome devtools mcp. this proxy mcp opens a child process which is the real chrome devtools mcp. tools that are standard to the mcp are forwarded and tools that are related to recording is handled by this proxy server instead. from my own screen recorder i pass back the first frame timestamp which allows me to accurately set this beginning timestamp.

### Additional context

_No response_

Sushanth012:
## Summary
- capture the first `Page.screencastFrame` metadata timestamp while starting a screencast
- include the timestamp in the human-readable `screencast_start` response
- expose the timestamp as `structuredContent.firstFrameTimestamp` for clients that need precise video annotation alignment
- add regression coverage for the response text and structured content

## Verification
- npm run test tests/tools/screencast.test.ts
- npm run typecheck
- npm run check-format

Note: on Windows, `npm run check-format` runs ESLint and Prettier successfully, then fails because the package script's trailing semicolon is passed to Prettier as a `".;"` file pattern.

Closes #2216

### DIFF
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -213,6 +213,7 @@ export class McpResponse implements Response {
   #attachedLighthouseResult?: LighthouseData;
   #textResponseLines: string[] = [];
   #images: ImageContentData[] = [];
+  #screencastFirstFrameTimestamp: number | undefined;
   #heapSnapshotOptions?: {
     include: boolean;
     aggregates?: Record<
@@ -436,6 +437,10 @@ export class McpResponse implements Response {
 
   appendResponseLine(value: string): void {
     this.#textResponseLines.push(value);
+  }
+
+  setScreencastFirstFrameTimestamp(timestamp: number): void {
+    this.#screencastFirstFrameTimestamp = timestamp;
   }
 
   attachWaitForResult(result: WaitForEventsResult): void {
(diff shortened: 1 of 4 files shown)


Lightning00Blade:
As mention in the issue there is already a way to get the information from the file. 
Even if we decide that exposing this make sense, the implementation should happen in Puppeteer, which gets the information to generate the frames. That would provide a much cleaner API.

PR #2253, opened 2026-06-24
fix: warn evaluate_script not to navigate page because navigation destroys execution context

### LINKED ISSUE #1894: The agent attempts to use window.location.href inside evaluate_script to scrape multiple pages in one go, which kills the execution context.
### Description of the bug

Example tool (mis)use:
```
// Inside evaluate_script
async () => {
  window.location.href = "https://books.toscrape.com/catalogue/..."; 
  await new Promise(r => setTimeout(r, 1000));
  return document.querySelector('td').innerText;
}
```

This happened without prior `chrome-devtools` skill activation. Observed one instance across six runs of this specific eval (nine if counting cli runs too).

After one attempt, agent correctly pivoted to using `navigate_page` and `evaluate_script` for each page separately.

### Reproduction

Prompt agent to find the same information on multiple related pages (like scrapping information about multiple products from an online store).

### Expectation

`evaluate_script` tool description should inform the agent about its limitations.

### MCP configuration

any

v0.20.1

### Chrome version

_No response_

### Coding agent version

gemini-cli

### Model version

gemini-2.5-flash

### Chat log

_No response_

### Node version

_No response_

### Operating system

None

### Extra checklist

AhmerYarnham:
Fixes #1894

## Changes
- Added warning in evaluate_script tool description that navigation 
  (window.location.href) destroys execution context
- Updated CLI help text
- Updated tool-reference.md docs
- Added regression test coverage

## How to test
Run: npm run test:no-build -- tests/tools/script.test.ts

### DIFF
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -345,8 +345,7 @@
 
 ### `evaluate_script`
 
-**Description:** Evaluate a JavaScript function inside the currently selected page. Returns the response as JSON,
-so returned values have to be JSON-serializable.
+**Description:** Evaluate a JavaScript function inside the currently selected page. Returns the response as JSON, so returned values have to be JSON-serializable. Do not navigate the page inside this script; use `navigate_page` or `new_page` to change URLs because page navigations destroy the execution context.
 
 **Parameters:**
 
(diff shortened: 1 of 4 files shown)


Lightning00Blade:
Thanks for the PR, but the real issue here is testing the new error, because of that we would prefer the team handle issues labeled with `evals`.

PR #2225, opened 2026-06-17
fix: apply URL filters to resource loading

mturac:
Fixes #2218.

`McpContext.loadResource()` already validated `file:` resources through workspace roots, but direct `http:`/`https:` resource fetches did not apply the configured URL filters. This threads the existing blocked/allowed URL patterns into the context and checks them before fetching remote resources.

Regression coverage uses the local test server to verify blocked URLs are rejected, allowed URLs still load, and an empty allowlist does not accidentally deny all resource loads.

How I tested:
- `npm run test tests/McpContext.test.ts`
- `npm run test tests/network_blocking.test.ts`
- `npm run check-format`

I also tried `npm run test`; the build completed, but the full suite failed locally in timing-sensitive unrelated tests (`telemetry.test.ts` waiting for `server_start`, and `shutdown.test.ts` stdin-EOF shutdown budget).

PR #2254, opened 2026-06-24
fix: respecte allow/block list in loadResouce

### LINKED ISSUE #2218: Apply url allowlist / blocklist to source maps being loaded
See todo in https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/src/McpContext.ts#L954

cc @nattallius @Lightning00Blade

Lightning00Blade:
Fixes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2218
Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2225

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -64,8 +64,10 @@ interface McpContextOptions {
   experimentalIncludeAllPages?: boolean;
   // Whether CrUX data should be fetched.
   performanceCrux: boolean;
-  // Whether allowlist/blocklist is configured.
-  hasNetworkBlockOrAllowlist?: boolean;
+  // The allow list of URL patterns to allow loading resources.
+  allowList?: string[];
+  // The block list of URL patterns to block loading resources.
+  blocklist?: string[];
 }
 
 const DEFAULT_TIMEOUT = 5_000;
@@ -347,6 +349,10 @@ export class McpContext implements Context {
     await this.emulate(currentSetting, page.pptrPage);
   }
 
+  get #hasNetworkBlockOrAllowlist(): boolean {
+    return !!(this.#options.allowList || this.#options.blocklist);
+  }
+
   async emulate(
     options: {
       networkConditions?: string;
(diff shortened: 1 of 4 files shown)


src/McpContext.ts
```diff
@@ -952,13 +958,41 @@ export class McpContext implements Context {
...
+        return;
+      }
+    }
+    throw new Error(`Not allowed by allowlist: ${url}`);
+  }
+
   async loadResource(path: string): Promise<string> {
     const url = new URL(path);
 
+    this.#isUrlBlocked(url);
```
Lightning00Blade: I am debating if we should check everything agains the block list or only the `http(s)` protocols.

src/McpContext.ts
```diff
@@ -952,13 +958,41 @@ export class McpContext implements Context {
     return await this.#heapSnapshotManager.getDominatorsOf(filePath, nodeId);
   }
 
+  #isUrlBlocked(url: URL): void {
```
zyzyzyryxy: `isUrlBlocked` suggest boolean getter, but this function throws instead. `validateUrlNotBlocked` maybe? Or retain the name, change return boolean and move `throw` to the callsite.

src/McpContext.ts
```diff
@@ -952,13 +958,41 @@ export class McpContext implements Context {
...
+    }
+    for (const block of this.#options.blocklist) {
+      const pattern = new URLPattern(block);
+      if (pattern.test(url)) {
+        throw new Error(`Blocked by blocklist: ${url}`);
+      }
+    }
+  }
+
+  #isUrlAllowed(url: URL): void {
```
zyzyzyryxy: Same as with `isUrlBlocked`

src/McpContext.ts
zyzyzyryxy: As discussed offline, I think it's fine to check all.

zyzyzyryxy (approved):

src/McpContext.ts
Lightning00Blade: Done

PR #2257, opened 2026-06-25
feat: support uploading multiple files with upload_file

### LINKED ISSUE #1217: Support multi-file uploads for existing-session mode

serhiizghama:
Closes #1217. `upload_file` only took a single path, so there was no way to fill an `input[type=file][multiple]` in one call.

`filePath` now also accepts an array of paths. A plain string still works for the single-file case, so existing callers don't change — Puppeteer's `uploadFile`/`fileChooser.accept` already take multiple paths, so the handler just normalizes to an array and forwards them.

The union (`string | string[]`) needed a bit of plumbing: the CLI generator threw on non-scalar types, the docs generator rendered it as `unknown`, and the telemetry transformer didn't know `ZodUnion`. The CLI exposes the single-value form (it can't pass arrays anyway), the docs show `string or array`, and metrics count it like an array with scalars normalized to one. Regenerated artifacts are included.

Added a test that uploads two files to a multi-file input and asserts both land, plus telemetry coverage for the union. `npm run typecheck`, lint/prettier, and the upload + transformation tests all pass.


itxaiohanglover:
Good feature addition! Supporting array filePath for multiple file uploads is useful. The CLI schema update to handle string | string[] union types is a nice touch — exposing the single-value form for CLI is pragmatic.

src/tools/input.ts
```diff
@@ -464,19 +464,24 @@ export const uploadFile = definePageTool({
       .describe(
         'The uid of the file input element or an element that will open file chooser on the page from the page content snapshot',
       ),
-    filePath: zod.string().describe('The local path of the file to upload'),
+    filePath: zod
+      .union([zod.string(), zod.array(zod.string()).min(1)])
```
Lightning00Blade: I think we can just change this to `filePaths` and support `zod.array(zod.string()).min(1)`
And then update the description to reflect that. 
Usually it's easier for LLM to follow simple patterns see `scripts/eslint_rules/enforce-zod-schema-rule.js`

serhiizghama:
Done — switched to a plain `filePaths: zod.array(zod.string()).min(1)` and updated the description. Dropping the union also let me revert the union handling I'd added to the CLI/docs generators and the telemetry transform, so the diff is smaller now. Single path still works, just as `[path]`.

src/tools/input.ts
```diff
@@ -464,19 +464,24 @@ export const uploadFile = definePageTool({
       .describe(
         'The uid of the file input element or an element that will open file chooser on the page from the page content snapshot',
       ),
-    filePath: zod.string().describe('The local path of the file to upload'),
+    filePaths: zod
+      .array(zod.string())
+      .min(1)
+      .describe(
+        'The local paths of the files to upload. Pass a single path to upload one file, or several paths to upload multiple files to an input that accepts them',
```
Lightning00Blade:
```suggestion
        'One or more local paths of files to upload.',
```

Lightning00Blade: I don't think we need to be super specific an can recover some tokens from shorting the description.

serhiizghama:
Good call — shortened it, went with your suggestion exactly.

Lightning00Blade (approved):

src/tools/input.ts
```diff
@@ -443,19 +443,25 @@ export const uploadFile = definePageTool({
...
   handler: async (request, response) => {
-    const {uid, filePath} = request.params;
+    const {uid, filePaths} = request.params;
     using handle = (await request.page.getElementByUid(
       uid,
     )) as ElementHandle<HTMLInputElement>;
 
+    // Some sites use a proxy element to trigger file upload instead of
+    // a type=file element. In this case, we want to default to
+    // Page.waitForFileChooser() and upload the file this way.
```
Lightning00Blade:
```suggestion
```

PR #2262, opened 2026-06-26
fix: create PID directory with secure permissions (0o700)

### LINKED ISSUE #2258: Daemon bails with "PID directory has insecure permissions" on a 002 umask
### Description of the bug

When I run the chrome-devtools-mcp command I get hit with:

[MCP Daemon] Critical error: PID directory <dir> has insecure permissions (Mode: 40775). It should not be writable by group or others.

I worked around it by wrapping the command in a little bash script that sets umask 077 before it runs, and that does the trick, but then I went digging to understand why I had to do that and that's where I got confused.

The daemon creates its own runtime dir without specifying a mode:

// src/daemon/daemon.ts
fs.mkdirSync(pidDir, {recursive: true});

so I suppose it just picks up whatever my umask happens to be, and mine is 002. So the dir lands at 0775 which is then promptly rejected and it bails.

I see that your own test (tests/daemon/symlink.test.ts) explicitly chmods the dir to 0700, so the daemon expects a 0700 dir, but nothing actually makes that happen at runtime... I don't really see the connection between that test setup and me having to fix it on my end.

if it'd be enough to just have the daemon create its own dir as 0700 directly that would line up with what your test already assumes, but yeah I'm not sure, if that's really all it takes I'm happy to spin up a quick PR, but if I'm missing something and this is intended I'd love to understand it, because right now I'm just confused.

### Reproduction

umask 0002 && npx chrome-devtools-mcp@latest    # → insecure permissions, exits
umask 0077 && npx chrome-devtools-mcp@latest    # → works

### Expectation

I would expect the command to not exit bacause of misconfigured folder permissions

### MCP configuration

_No response_

1.4.0

### Chrome version

_No response_

### Coding agent version

_No response_

### Model version

_No response_

### Chat log

_No response_

### Node version

_No response_

### Operating system

Linux

### Extra checklist

itxaiohanglover:
## Summary

- Add `mode: 0o700` to `fs.mkdirSync` when creating the PID directory
- Ensures the directory is always created with owner-only access regardless of system umask

## Problem

The daemon creates the PID directory with `fs.mkdirSync(pidDir, {recursive: true})` without specifying a mode. On systems with `umask 002`, the directory lands at mode 0775, which the daemon's own security check rejects as "insecure permissions" and exits.

The daemon's test (`tests/daemon/symlink.test.ts`) already expects 0700 permissions, but nothing enforced it at runtime.

Fixes #2258

## Test Plan

- [ ] `umask 0002 && npx chrome-devtools-mcp@latest` no longer crashes
- [ ] `umask 0077 && npx chrome-devtools-mcp@latest` still works
- [ ] Existing daemon tests pass

### DIFF
--- a/src/daemon/daemon.ts
+++ b/src/daemon/daemon.ts
@@ -41,7 +41,7 @@ const pidDir = path.dirname(pidFilePath);
 const currentUserUid = os.userInfo().uid;
 
 try {
-  fs.mkdirSync(pidDir, {recursive: true});
+  fs.mkdirSync(pidDir, {recursive: true, mode: 0o700});
   if (os.platform() !== 'win32') {
     // POSIX specific checks
     try {


itxaiohanglover:
I've signed the Google CLA. Could you re-run the CLA check?

zyzyzyryxy (approved):

PR #2269, opened 2026-06-28
fix: validate extension-enforced output paths

huynhtrungcsc:
Fixes output path validation so tools cannot validate one path and then write to a different canonical target after extension enforcement.

Changes:
- Resolve dangling symlinks to their target path during canonicalization.
- Validate the final extension-enforced output path before writing.
- Apply the same final-path validation to heap snapshots and screencasts.
- Add regression coverage for dangling symlinks that point outside configured roots.

Validation:
- npm run format
- npm run check-format
- npm run test tests/utils/files.test.ts
- npm run test tests/roots.test.ts
- npm run test tests/tools/memory.test.ts tests/tools/screencast.test.ts

Note: I also ran the full npm test suite locally. The targeted tests above passed, but the full suite hit local WSL daemon/e2e startup timeouts while waiting for daemon.pid / server_start, which appear unrelated to this path-validation change.

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -55,7 +55,6 @@ import type {
   ExtensionServiceWorker,
 } from './types.js';
 import {
-  ensureExtension,
   getTempFilePath,
   resolveCanonicalPath,
 } from './utils/files.js';
@@ -256,6 +255,20 @@ export class McpContext implements Context {
         `Access denied: path ${filePath} (canonical: ${canonicalPath}) is not within any of the configured workspace roots.`,
       );
     }
+  }
+
+  async ensureExtension<Extension extends `.${string}`>(
+    filePath: string,
+    extension: Extension,
+  ): Promise<`${string}${Extension}`> {
+    const resolvedPath = path.resolve(filePath);
+    const currentExtension = path.extname(resolvedPath);
+    const outputPath: `${string}${Extension}` = `${resolvedPath.slice(
+      0,
+      resolvedPath.length - currentExtension.length,
+    )}${extension}`;
+    await this.validatePath(outputPath);
+    return outputPath;
   }
 
   resolveCdpRequestId(page: McpPage, cdpRequestId: string): number | undefined {
@@ -800,12 +813,11 @@ export class McpContext implements Context {
     clientProvidedFilePath: string,
     extension: SupportedExtensions,
   ): Promise<{filename: string}> {
-    await this.validatePath(clientProvidedFilePath);
+    const filePath = await this.ensureExtension(
+      clientProvidedFilePath,
+      extension,
+    );
     try {
-      const filePath = ensureExtension(
-        path.resolve(clientProvidedFilePath),
-        extension,
-      );
       await fs.mkdir(path.dirname(filePath), {recursive: true});
       await fs.writeFile(filePath, data);
       return {filename: filePath};
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -54,10 +54,7 @@ import type {
   GeolocationOptions,
   ExtensionServiceWorker,
 } from './types.js';
-import {
-  getTempFilePath,
-  resolveCanonicalPath,
-} from './utils/files.js';
+import {getTempFilePath, resolveCanonicalPath} from './utils/files.js';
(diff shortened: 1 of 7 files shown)


src/tools/memory.ts
```diff
@@ -24,16 +26,19 @@ export const takeHeapSnapshot = definePageTool({
...
   blockedByDialog: true,
   verifyFilesSchema: ['filePath'],
-  handler: async (request, response) => {
+  handler: async (request, response, context) => {
     const page = request.page;
+    const snapshotPath = ensureExtension(
+      path.resolve(request.params.filePath),
+      '.heapsnapshot',
+    );
+    await context.validatePath(snapshotPath);
```
Lightning00Blade: This looks like a easy pattern to miss/forget in the future, we can just move the function `ensureExtension` on the context and validate before returning the value.

Lightning00Blade (commented):
The PR tried to handle two separate issue, it would be nice if we can split the PR in two - one handling the issue with `ensureExtension` not being validate and the other is a missing case in the validation itself.

huynhtrungcsc:
Thanks for the review. I updated this PR to focus only on the extension-enforced output path issue and moved the extension enforcement onto the MCP context so callers receive a validated path. I also removed the separate canonical path validation changes from this PR so that can be handled independently.

src/McpContext.ts
```diff
@@ -254,6 +254,15 @@ export class McpContext implements Context {
     }
   }
 
+  async ensureExtension<Extension extends `.${string}`>(
+    filePath: string,
+    extension: Extension,
+  ): Promise<`${string}${Extension}`> {
+    const resolvedPath = ensureExtension(path.resolve(filePath), extension);
```
Lightning00Blade: Two things:
1. Let's move the helper function logic here as I don't see an usecase where we will want to skip the validatePath.
2. Second in the `return` isn't the `resolvedPath` already with the correct extension after the call to `ensureExtension`?

huynhtrungcsc: Done in 51edd0e. I moved the extension replacement logic into McpContext.ensureExtension(), removed the standalone ensureExtension helper, and now return the validated output path directly. Local build and focused roots/files tests pass; the new fork workflow runs are currently awaiting maintainer approval.

src/McpContext.ts
```diff
@@ -254,6 +253,24 @@ export class McpContext implements Context {
     }
   }
 
+  async ensureExtension<Extension extends `.${string}`>(
```
Lightning00Blade, line 257: We should just define this in the method directly and not use an type overloading.

tests/utils/files.test.ts
```diff
@@ -10,40 +10,7 @@ import os from 'node:os';
 import path from 'node:path';
 import {describe, it} from 'node:test';
 
-import {ensureExtension, resolveCanonicalPath} from '../../src/utils/files.js';
-
-describe('ensureExtension', () => {
-  it('should add an extension to a filename without one', () => {
```
Lightning00Blade, line 16: Please move the test so the now `context.ensureExtension` is tested.

src/McpContext.ts
huynhtrungcsc: Done in 6b4b5fb. I removed the overload and defined the typed output path directly inside McpContext.ensureExtension().

tests/utils/files.test.ts
huynhtrungcsc: Done in 6b4b5fb. The extension behavior cases are now covered through context.ensureExtension() in tests/roots.test.ts, and the old standalone utility tests were removed with the helper.

Lightning00Blade (approved):

huynhtrungcsc:
Thanks again. I rebased this PR onto the latest main and pushed a small formatting-only follow-up commit (588eb3f) to address the required format/docs checks after the merge-base changed.

Local validation on Node 24.17.0:
- npm ci
- npm run check-format
- npm run gen && git diff --exit-code
- npm run test tests/McpContext.test.ts tests/utils/files.test.ts tests/tools/memory.test.ts tests/tools/screencast.test.ts

All passed locally. The remaining required workflows have not started again yet after the fork push, so they may need maintainer approval/re-run; auto-merge also appears to have been reset by the force-push.

PR #2271, opened 2026-06-29
fix: add option to disable focused page emulation

### LINKED ISSUE #2290: Allow passive MCP inspection without stealing foreground window focus

I use `chrome-devtools-mcp` with a visible Chrome instance as part of a local side-agent workflow. The agent runs in opencode and uses CDP/MCP to inspect and test its work in Chrome while I continue using the same machine for other tasks, such as writing an email, working in Safari, using an IDE, or doing something else on another screen.

In this workflow, passive MCP activity such as inspecting or snapshotting pages can cause Chrome to become the active foreground window. That interrupts whatever I am doing manually. For example, I may be typing in another app, the agent uses MCP to inspect its Chrome page, Chrome becomes active, and my next keystrokes go into Chrome’s address bar or into whichever page element currently has focus.

This makes it difficult to let an agent work “on the side” using a visible browser, because ordinary inspection actions can interfere with the user’s active desktop session.

Add an option to disable focused-page emulation or other foreground-focus side effects for passive page snapshot/inspection behavior, while keeping the existing default behavior unchanged.

Explicit focus-changing actions should still be possible. If the MCP client intentionally needs to bring Chrome or a page to the front, that should remain available. The issue is passive/background inspection causing Chrome to become active when the user did not ask for that.

This would preserve the current behavior for users who rely on focused-page emulation, while allowing users running visible browser agents on their own workstation to prevent the agent from interrupting their manual work.

- Running Chrome headless avoids stealing desktop focus, but removes the ability to visually monitor the browser while the agent works.
- Running the agent/browser on a separate machine or separate user session avoids the issue, but adds unnecessary operational overhead for a local workflow.
- Avoiding the machine while the agent works defeats the purpose of a side-agent workflow.
- Telling the agent not to bring Chrome to the front helps only for explicit foreground actions. It does not address passive inspection/snapshot behavior that can still make Chrome active.

### Additional context

Related PR: #2271

This request is based on a real local workflow where an agent uses Chrome through MCP while the user continues working on the same machine. The requested behavior is an opt-out for passive inspection/snapshot focus side effects, not removal of explicit page activation.

samiralibabic:
## Summary

Fixes #2290

- Adds `--emulate-focused-pages` / `--no-emulate-focused-pages` so users can opt out of passive focused-page emulation during page discovery/snapshot setup.
- Forwards the `background` option when creating pages in isolated contexts.
- Documents the new CLI option in `README.md` and registers its generated telemetry metrics.

## Scope

This intentionally does **not** suppress explicit activation requests such as `select_page` with `bringToFront: true`.

The option only controls passive focused-page emulation applied when MCP pages are discovered and initialized.

Focused-page emulation remains enabled by default to preserve existing behavior for multi-agent workflows where pages should behave as focused/active. Disable it only when visible-browser focus behavior is problematic.

## Validation

- `npm run format`
- `npm run build`
- `npm run test:no-build -- tests/telemetry/metricsRegistry.test.ts tests/telemetry/ClearcutLogger.test.ts tests/cli.test.ts`
- `NODE_TEST_REPORTER=spec npm run test:no-build -- --test-timeout=300000 tests/McpContext.test.ts` (37/37 passed)
- `npm test -- --test-timeout=300000` reached one existing extension teardown timeout; `NODE_TEST_REPORTER=spec npm run test:no-build -- --retry --test-timeout=300000 tests/tools/extensions.test.ts` then passed 5/5.


zyzyzyryxy:
Thank you for your contribution!
We'd like to understand your case better before deciding on this PR. Please file a feature request issue and describe in detail in what scenario emulating a focused page is problematic. We'll follow up with questions there if needed.

samiralibabic:
> Thank you for your contribution! We'd like to understand your case better before deciding on this PR. Please file a feature request issue and describe in detail in what scenario emulating a focused page is problematic. We'll follow up with questions there if needed.

Thanks. I filed the feature request here: #2290

The concrete scenario is a local side-agent workflow: an agent uses Chrome through CDP/MCP in a visible browser while I continue working in another app, such as Safari, an IDE, email, or something on another screen. Passive MCP inspection/snapshot activity can make Chrome the active foreground window, so my keyboard input gets stolen from the app I am using and goes into Chrome instead.

The requested behavior is not to remove explicit page activation. If the agent intentionally needs to bring Chrome/page to front, that should still be available. The issue is passive inspection changing foreground focus during background agent work.

OrKoN:
Closing as stale.

PR #2272, opened 2026-06-30
docs: fix Antigravity config -y flag before npx positional args

Humanpanda666:
Moves the `-y` flag before the package name in the Antigravity JSON configuration example.

According to official npm documentation, `npx` requires all flags to precede positional arguments. Placing `-y` at the end caused `npx` to treat it as an argument for the MCP server, triggering an interactive installation prompt that hangs indefinitely in headless CLI environments.


Lightning00Blade (approved):
Thanks, indeed the order is wrong here.

Issue #1921, opened 2026-04-21
Browser hangs / crashes when MCP connects to Chrome with many tabs


jabagawee:
### Description of the bug

When chrome-devtools-mcp connects to a Chrome instance with a very large number of open tabs (about 2000 or so for me), the browser becomes unresponsive and then crashes on the first tool call. `list_pages` is the simplest trigger, but any tool call that constructs `McpContext` pays the same startup cost.

The MCP does a substantial amount of eager per-tab work when the context is initialized and then redoes per-tab work on every response that sets `includePages` (which is most navigation / tab-management tools). A comment in `McpContext.ts` notes that page count is "typically 2-10", which... is not the case for me and potentially other power users.

Notes on the specific places that scale with tab count are below. I wanted to file one investigation issue first rather than splitting into multiple bugs, since most of the findings could be resolved with the same design change. Happy to scope this however maintainers prefer.

### Related but distinct

I searched through related issues, and I believe that #1230 and #1918 are adjacent failure modes (frozen/discarded tabs; single-bad-tab hangs). The Chromium fix in #1230 (https://crrev.com/c/7772775) addresses one specific bad-tab failure mode but does not change the MCP's eager-load architecture. #1918 was closed as Edge-specific, but my issue reproduces with Chrome.

### Reproduction

1. Start (or have running) a Chrome instance with ~2000 tabs on one profile.
2. Enable remote debugging via `chrome://inspect/#remote-debugging`. (Note: this disables the HTTP discovery endpoint, so `/json/version` 404s. The WebSocket is available at `ws://127.0.0.1:9222/devtools/browser` with no browser-id suffix.)
3. Configure the MCP to connect to that instance.
4. Invoke any tool (`list_pages` is simplest).

Observed: Chrome UI becomes unresponsive within seconds and the browser process is killed/crashes. The tool call does not return.

Reproduced both via a normal MCP client and by piping JSON-RPC directly into the MCP server's stdin.

### Expectation

`list_pages` and other tools work against large profiles without hanging or crashing the browser.

### MCP configuration

```json
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": [
        "chrome-devtools-mcp@latest",
        "--wsEndpoint=ws://127.0.0.1:9222/devtools/browser"
      ]
    }
  }
}
```

#### Cold start (McpContext#init, once per session)

[...]

PR #2274, opened 2026-07-01
fix: cap concurrency in detectOpenDevToolsWindows to avoid CDP fan-out on many-tab profiles

mvanhorn:
## Summary

Caps the concurrency of the DevTools-window detection pass so it no longer issues one simultaneous CDP call per open tab.

## Why

`detectOpenDevToolsWindows()` in `src/McpContext.ts` fanned out an unbounded `Promise.all(pages.map(...))` over every open page, so each tab issued its own `hasDevTools()` / `openDevTools()` CDP round-trip at the same time. This method runs on init and again on every response that carries `includePages`, so a single tool call could put one concurrent CDP call in flight per tab.

On a profile with a very large number of tabs this becomes hundreds to thousands of simultaneous CDP calls on the first tool call, which is one of the paths that can make the browser hang or crash (Refs #1921).

## Change

Bound the fan-out with a small fixed-size worker pool. A new `mapWithConcurrency(items, limit, worker)` helper runs at most `limit` workers at once (default cap `10`, `DEVTOOLS_DETECTION_CONCURRENCY`), each pulling the next page from a shared index. `detectOpenDevToolsWindows()` now walks the pages through this helper with the exact same per-page body it had before.

- Profiles at or below the cap behave identically to before.
- The existing per-page `try/catch` fallback for pre-`144.0.7559.59` Chrome/Electron (where the DevTools command throws) is preserved, and one page failing does not abort detection for the rest.

This is deliberately narrow: it only caps the concurrency of the existing detection pass and does not change what gets detected or attached. It is one piece of the broader many-tab hardening tracked in #1921, not a full fix for it.

## Testing

Added unit tests for `mapWithConcurrency` in `tests/McpContext.test.ts` covering:
- every item is processed exactly once below the limit;
- the observed maximum in-flight worker count never exceeds the limit while all items still complete (50 items, limit 3);
- a rejecting worker does not prevent the remaining items from being processed.

`npm run check-format`, `npm run build`, and `npm test` all pass locally.

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -61,6 +61,53 @@ import {
 } from './utils/files.js';
 import {getNetworkMultiplierFromString} from './WaitForHelper.js';
 
+// Bounds concurrent DevTools detection CDP round-trips for large profiles.
+const DEVTOOLS_DETECTION_CONCURRENCY = 10;
+
+export async function mapWithConcurrency<T>(
+  items: readonly T[],
+  limit: number,
+  worker: (item: T) => Promise<void>,
+): Promise<void> {
+  if (limit < 1) {
+    throw new Error('Concurrency limit must be at least 1.');
+  }
+
+  let nextIndex = 0;
+  let firstError: {error: unknown} | undefined;
+
+  async function runWorker(): Promise<void> {
+    while (true) {
+      const index = nextIndex;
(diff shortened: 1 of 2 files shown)


OrKoN:
Closing as stale

PR #2277, opened 2026-07-01
chore: rename misspelled functions, remove duplicate import and dead code

fix2015:
Found a handful of identifier typos in NetworkFormatter.ts while reading through the network capture code:

- \`formatHeadlers\` → \`formatHeaders\` (function def + 2 call sites)
- \`converNetworkRequestDetailedToStringDetailed\` → \`convertNetworkRequestDetailedToStringDetailed\` (missing 't')
- \`responseAsTest\` → \`responseAsText\`

Also fixed a few other small things:
- \`navigationStated\` → \`navigationStarted\` in WaitForHelper.ts
- doubled "for for" in a PageCollector log message
- \`node:fs/promises\` was imported twice under different names in McpContext.ts (\`fs\` and \`fsPromises\`) — consolidated to a single import
- removed a dead \`if (!navigations)\` guard in NetworkCollector.splitAfterNavigation — \`navigations\` is already defaulted to \`[]\` via \`??\` so the check can never be falsy

### DIFF
--- a/src/PageCollector.ts
+++ b/src/PageCollector.ts
@@ -374,7 +374,10 @@ export class NetworkCollector extends PageCollector<HTTPRequest> {
     super(browser, listeners);
   }
   override splitAfterNavigation(page: Page) {
-    const navigations = this.storage.get(page) ?? [];
+    const navigations = this.storage.get(page);
+    if (!navigations) {
+      return;
+    }
 
     const requests = navigations[0];
 
--- a/src/PageCollector.ts
+++ b/src/PageCollector.ts
@@ -335,7 +335,7 @@ class PageEventSubscriber {
         inspectorIssue,
       )[0];
       if (!issue) {
-        logger?.('No issue mapping for for the issue: ', inspectorIssue.code);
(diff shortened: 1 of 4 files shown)


OrKoN (approved):
Thanks!

src/PageCollector.ts
```diff
@@ -375,9 +375,6 @@ export class NetworkCollector extends PageCollector<HTTPRequest> {
   }
   override splitAfterNavigation(page: Page) {
     const navigations = this.storage.get(page) ?? [];
-    if (!navigations) {
```
Lightning00Blade, line 378: I think this should not be removed but rather then `?? []` should be.

fix2015:
Good point, you're right — keeping the early return and dropping the `?? []` instead makes more sense. That way we get a clean bailout when the page has no navigations rather than silently continuing with an empty array. I'll update the PR.

Lightning00Blade (approved):

Lightning00Blade:
@fix2015 There seem to be some merge conflicts after we refactored the code, please rebase so we can merge the change.

OrKoN:
Closing as stale

PR #2281, opened 2026-07-02
refactor: Merge MCP tools for comparing heap snapshots

dinfuehr:
Instead of two separate MCP tools, we can have one instead which takes classIndex as optional argument.


Lightning00Blade (approved):
Thanks!

dinfuehr:
Thanks for the review!

PR #2285, opened 2026-07-02
chore: remove debug

Lightning00Blade:
Removes the debug lib and replaces it with custom logic for the FileLog and Node native logger for the other cases.


OrKoN (changes requested):
With this change the data logged does not seem to include logging metadata (e.g.,`2026-05-07T06:21:39.089Z mcp:log`). Only the content is logged. Btw I noticed that the file traffic was already missing CDP logs. Any thoughts on how to fix it?

Lightning00Blade:
Closing as we have a better way to do it on the latest Puppeteer version

PR #2287, opened 2026-07-03
Update Chrome DevTools MCP Tool Reference formatting

sellomasenya64-creator:
Fix formatting and correct title in tool reference.


zyzyzyryxy:
This CL actually breaks stuff and attempts to manually edit generated file. Spam.

PR #2303, opened 2026-07-06
fix: don't double multi-part extensions when saving files

serhiizghama:
Saving a performance trace to a compressed path like `trace.json.gz` (the exact example in the `performance_start_trace`/`performance_stop_trace` schema) writes it to `trace.json.json.gz` instead.

The cause is in `ensureExtension`: it strips the current extension with `path.extname` and re-appends the target one, but `path.extname("trace.json.gz")` only returns `.gz`. So it strips `.gz` → `trace.json`, then appends `.json.gz` → `trace.json.json.gz`. The compressed-trace tools pass `.json.gz` as the extension, so this hits any trace saved to a `.gz` path.

Fix is to skip the strip-and-reappend when the path already ends with the requested extension. Single-segment extensions and the replace-a-different-extension case (e.g. `result.jpg` → `.png`) are unchanged; added a couple of cases to the existing `ensureExtension` test to lock in the `.json.gz` behavior.

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -260,6 +260,13 @@ export class McpContext implements Context {
     extension: Extension,
   ): Promise<`${string}${Extension}`> {
     const resolvedPath = path.resolve(filePath);
+    // path.extname only strips the final segment, so a multi-part extension
+    // like `.json.gz` would be doubled (trace.json.gz -> trace.json.json.gz).
+    // Keep the path as-is when it already carries the requested extension.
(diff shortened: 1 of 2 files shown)


OrKoN (commented):
Thanks for the PR. It looks like some tests fail but I also wonder let's just enforce `.gz` instead of `.json.gz`?

OrKoN:
I have opened a PR for that https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2303

PR #2307, opened 2026-07-06
fix: preserve selected page when snapshot omits it

syedaly:
Fix page selection so it no longer silently jumps to the first page when a fresh snapshot temporarily omits the currently selected page.

What changed
Updated the page snapshot refresh logic to preserve the existing selection unless the previously selected page is actually closed.
Added a regression test covering the case where the selected page is missing from a snapshot.
Why
Previously, a transient snapshot omission could cause the selected page to be retargeted to an arbitrary first page. This could make subsequent tool calls operate on the wrong tab without an obvious error.

Testing
Added a regression test for selection preservation during snapshot refresh.


nattallius:
Hi, thank you for your contribution and for taking the time to address this issue.

However, this pull request currently combines several distinct modifications into a single submission.
To ensure a clean commit history and streamline the review process, we would appreciate if you could split these changes into separate, targeted pull requests for each individual fix. Consequently, we will close this pull request for now, but we look forward to reviewing your modular submissions if you would like to deliver these updates.

Issue #2304, opened 2026-07-06
Page listing silently retargets the selection to the first page, bypassing the closed page error


thomasbachem:
### Description of the bug

There are two code paths for a selected page that no longer exists, and they contradict each other.

Tools that operate on the selected page fail with a clear, actionable error:

```ts
// src/McpContext.ts – getSelectedPptrPage()
if (page.pptrPage.isClosed()) {
  throw new Error(`The selected page has been closed. Call ${listPages().name} to see open pages.`);
}
```

But any tool whose response includes the page list (`list_pages`, `select_page`, `new_page`, …) refreshes the snapshot first – and that silently re-selects the *first* page whenever the selected one is missing from `browser.pages()` at that moment:

```ts
// src/McpContext.ts – createPagesSnapshot()
if (
  (!this.#selectedPage ||
    this.#pages.indexOf(this.#selectedPage.pptrPage) === -1) &&
  this.#pages[0]
) {
  this.selectPage(this.#getMcpPage(this.#pages[0]));
}
```

So the recovery step the error message itself prescribes ("Call list_pages") quietly moves the selection to an arbitrary page, and every subsequent tool call runs there with normal-looking responses. For an AI agent this is the worst failure mode: After the one clear error (or with an intervening `list_pages`, without ever seeing an error) it gets *plausible results from the wrong page*. We have watched agents lose several turns debugging the wrong target before a listing happened to reveal the moved `[selected]` marker.

The same fallback also fires when the selected page is only *transiently* absent from a snapshot. Attached via `--browserUrl` to an Electron app, we routinely observe pages dropping out of one listing and returning in the next (tabs reloading while the app relaunches) — each occurrence silently hops the selection to the first page even though the selected page is back moments later.

### Reproduction

Verified with plain headless Chrome:

1. `chrome --headless=new --remote-debugging-port=9444 about:blank`, then connect `chrome-devtools-mcp --browserUrl http://127.0.0.1:9444`.
2. `new_page` → `https://example.com/` (becomes the selected page).
3. Close that tab externally (user closes it / `curl http://127.0.0.1:9444/json/close/<targetId>`).
4. `evaluate_script` → ✅ `Error: The selected page has been closed. Call list_pages to see open pages.`
5. `list_pages` → shows `1: about:blank [selected]` – the selection moved to the first page without any notice.
6. `evaluate_script` → runs on `about:blank` and responds normally – no error, wrong page.

### Expectation

[...]

PR #2308, opened 2026-07-06
fix: report when the selected page was auto-replaced by the fallback

thomasbachem:
Follow-up to #2304 (the "note" part discussed there).

When the selected page disappears from the page list, `createPagesSnapshot()` silently re-selects the first page. The agent gets no signal: if its next call is `list_pages` (which is what the closed-page error message recommends), the listing already shows the new selection and every subsequent tool call runs against a page the agent never picked.

This PR records the automatic fallback and surfaces it as a one-line note in the pages section of the same response:

```
## Pages
Note: the previously selected page was closed. Page 1 is now selected.
1: about:blank [selected]
```

For a selected page that is missing from the list without being closed, the note reads "is no longer listed" instead. If the expectation from #2304 holds (a page stays listed as long as it is not closed), that wording never renders; if the transient case discussed there does occur in the wild, the note will make it visible.

The fallback behavior itself is unchanged (as discussed in #2304, closed tabs keep the browser-like auto-selection). No note is emitted on first connect, when nothing was selected before.

Tests: two unit tests for the fallback bookkeeping (closed page, regular selection), one for the missing-but-open case via a stubbed page list, and the one affected snapshot updated (`close_page` now includes the note).

Refs: #2304

### DIFF
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -986,19 +986,19 @@ Call ${handleDialog.name} to handle it before continuing.`);
       );
 
       const selectionFallback = context.getSelectedPageFallback();
+      if (selectionFallback) {
+        let selectedPageId: number | undefined;
+        try {
+          selectedPageId = context.getSelectedMcpPage().id;
+        } catch {
+          selectedPageId = undefined;
+        }
+        response.push(
+          `Note: the previously selected page ${selectionFallback.wasClosed ? 'was closed' : 'is no longer listed'}.${selectedPageId !== undefined ? ` Page ${selectedPageId} is now selected.` : ''}`,
+        );
+      }
       if (regularPages.length) {
         const parts = [`## Pages`];
-        if (selectionFallback) {
-          const selectedPage = regularPages.find(page =>
-            context.isPageSelected(page),
-          );
-          const selectedPageId = selectedPage
-            ? context.getPageId(selectedPage)
-            : undefined;
-          parts.push(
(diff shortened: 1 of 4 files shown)


thomasbachem:
Just a heads-up: The macos-latest/node24 cell hit a timeout in `evaluates inside extension service worker` (passed on the other 9 cells, unrelated to the change), so I force-pushed the same commit to retrigger CI. The re-run is waiting for workflow approval.

src/McpResponse.ts
```diff
@@ -1006,8 +985,20 @@ Call ${handleDialog.name} to handle it before continuing.`);
         {regularPages: [], extensionPages: []},
       );
 
+      const selectionFallback = context.getSelectedPageFallback();
       if (regularPages.length) {
         const parts = [`## Pages`];
+        if (selectionFallback) {
+          const selectedPage = regularPages.find(page =>
```
OrKoN: I think this should use allPages and the note probably should come before `## Pages` so that it applied to the extension pages as well.

src/McpResponse.ts
```diff
@@ -1006,8 +985,20 @@ Call ${handleDialog.name} to handle it before continuing.`);
...
+            context.isPageSelected(page),
```
OrKoN: we could use `context.getSelectedMcpPage()` with `try/catch`

src/McpResponse.ts
thomasbachem: Done in b56c9d2, the note now renders before all page sections so it covers extension pages too.

src/McpResponse.ts
thomasbachem: Done, switched to `getSelectedMcpPage()` with a try/catch in the same commit.

OrKoN (approved):

PR #2311, opened 2026-07-06
fix: don't crash the tab when upload_file targets a chooser-opening element

### LINKED ISSUE #2310: upload_file on a chooser-opening element crashes the tab with RESULT_CODE_KILLED_BAD_MESSAGE
### Description of the bug

Calling `upload_file` with the `uid` of an element that **opens the file chooser** (e.g. a button that calls `input.click()`, which is how most real-world upload UIs work — the actual `<input type="file">` is hidden) crashes the Chrome tab with **`RESULT_CODE_KILLED_BAD_MESSAGE`** ("Aw, Snap!"). The browser process kills the renderer, presumably over the file-chooser interception IPC.

Calling `upload_file` with the `uid` of the `<input type="file">` element itself works correctly (file is set, `change` event fires).

The tool description explicitly documents the chooser-opening-element case as supported:

> The uid of the file input element **or an element that will open file chooser on the page** from the page content snapshot

The tool response after the crash is misleading — it reports `Error: Element with uid X no longer exists on the page.` because the whole tab has crashed, so an agent has no way to tell the page died and typically keeps flailing. This has been reproducible for months across projects.

### Reproduction

Self-contained repro (no external site needed):

1. `new_page` with this URL:
   ```
   data:text/html,<h1>upload test</h1><input type="file" id="f"><button onclick="document.getElementById('f').click()">Choose via chooser</button><div id="out"></div><script>document.getElementById('f').addEventListener('change',e=>{document.getElementById('out').textContent='GOT: '+e.target.files[0].name})</script>
   ```
2. `take_snapshot` → note the uid of the **button** ("Choose via chooser")
3. `upload_file` with that button's uid and any local file path
4. Tab crashes: "Aw, Snap! RESULT_CODE_KILLED_BAD_MESSAGE". Tool returns `Error: Element with uid <uid> no longer exists on the page.`

Control: step 3 with the uid of the `<input type="file">` element instead succeeds — snapshot afterwards shows `value="upload-test.txt"` and `StaticText "GOT: upload-test.txt"`.

### Expectation

`upload_file` on a chooser-opening element intercepts the chooser and sets the file, as documented — or at minimum fails gracefully without killing the renderer, and reports the tab crash rather than "element no longer exists".

### MCP configuration

```
claude mcp add chrome-devtools -- npx -y chrome-devtools-mcp@latest --browserUrl http://172.19.96.1:9222
```

Connecting from WSL2 to a user-launched Chrome on the Windows host started with `--remote-debugging-port=9222` (headful, real profile).

1.5.0

### Chrome version

Chrome/150.0.7871.47 (Windows x64, headful)

### Coding agent version

Claude Code 2.1.201

### Model version

claude-fable-5

### Node version

v22.22.0

### Operating system

serhiizghama:
Fixes #2310. `upload_file` calls `handle.uploadFile()` on whatever element you pass it first, and only falls back to the file chooser if that throws. But when the element is a proxy that opens the chooser (a button calling `input.click()`, which is how most real upload UIs work), `uploadFile` doesn't throw — puppeteer reads `element.multiple` (undefined on a button, no error) and then sends `DOM.setFileInputFiles` to a non-input node, which kills the renderer with `RESULT_CODE_KILLED_BAD_MESSAGE`. The try/catch can't help because the tab is already gone, and the tool then reports a misleading "element no longer exists" error.

The tool's own schema documents the chooser-opening element as supported, so the fix is to stop probing with `uploadFile`: check whether the element is actually an `<input type=file>` and only use the direct path for those. Everything else goes straight to the click + `waitForFileChooser` path it was always meant to use.

All three existing upload_file tests still pass. Added one for a non-file `<input>` to lock in that the check is `type === 'file'`, not just `instanceof HTMLInputElement` — otherwise a text input would still be routed into the crashing path.

### DIFF
--- a/src/tools/input.ts
+++ b/src/tools/input.ts
@@ -475,12 +475,18 @@ export const uploadFile = definePageTool({
       uid,
     )) as ElementHandle<HTMLInputElement>;
     try {
-      try {
+      const isFileInput = await handle.evaluate(
+        element =>
+          element instanceof HTMLInputElement && element.type === 'file',
+      );
+      if (isFileInput) {
         await handle.uploadFile(filePath);
-      } catch {
-        // Some sites use a proxy element to trigger file upload instead of
-        // a type=file element. In this case, we want to default to
-        // Page.waitForFileChooser() and upload the file this way.
+      } else {
+        // Some sites use a proxy element (e.g. a button that calls
+        // input.click()) to open the chooser instead of exposing the
+        // type=file input. Calling uploadFile on such an element sends
+        // DOM.setFileInputFiles to a non-input node, which crashes the tab, so
+        // drive the chooser explicitly.
         try {
           const [fileChooser] = await Promise.all([
             request.page.pptrPage.waitForFileChooser({timeout: 3000}),
--- a/src/tools/input.ts
+++ b/src/tools/input.ts
@@ -475,18 +475,12 @@ export const uploadFile = definePageTool({
       uid,
     )) as ElementHandle<HTMLInputElement>;
     try {
-      const isFileInput = await handle.evaluate(
-        element =>
-          element instanceof HTMLInputElement && element.type === 'file',
-      );
-      if (isFileInput) {
+      try {
         await handle.uploadFile(filePath);
-      } else {
-        // Some sites use a proxy element (e.g. a button that calls
-        // input.click()) to open the chooser instead of exposing the
-        // type=file input. Calling uploadFile on such an element sends
-        // DOM.setFileInputFiles to a non-input node, which crashes the tab, so
-        // drive the chooser explicitly.
+      } catch {
+        // Some sites use a proxy element to trigger file upload instead of
+        // a type=file element. In this case, we want to default to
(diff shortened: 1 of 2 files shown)


OrKoN:
Thanks for the PR but I think the test passes without the fix (please see https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2313)

serhiizghama:
You're right — the old test passed without the fix. The in-process test browser throws when `uploadFile` hits a non-file element, so the code recovered via the chooser and the assertion held either way. That doesn't happen in real headful Chrome, where `uploadFile` on a chooser-opening element doesn't throw — it kills the renderer and the file is never set.

Reworked the test to reproduce that: it stubs `uploadFile` on the button to a silent no-op (matching production, where probing the proxy element uploads nothing), then asserts `uploadFile` is never called on the non-file element and that the chooser path actually delivers the file. Reverting the source fix now fails it. Full suite green.

OrKoN:
closing as stale

Issue #2291, opened 2026-07-03
Growing memory usage during screencast recording if the host machine is not able to encode video using ffmpeg fast enough


jackzhouusa:
## Summary

A long-lived `chrome-devtools-mcp` server (spawned by an AI coding-agent CLI via `npm exec chrome-devtools-mcp@latest`) attached to an existing Chrome with `--browserUrl` steadily grows its JS heap until it hits the default ~4GB V8 heap limit and aborts. The `npm exec` wrapper process dies alongside it (unhandled-rejection path → `uv_kill`), so each event leaves a pair of crashed `node` processes.

This looks distinct from #1192 / #1214: those were `--autoConnect`-mode collector leaks fixed in March/April — and indeed we saw ~3 months of clean operation after those fixes. The leak reported here reappeared with 1.2.0+ in `--browserUrl` mode.

## Environment

- chrome-devtools-mcp: crashes observed on **1.2.0 and 1.4.0** (resolved fresh via `npm exec chrome-devtools-mcp@latest` on each start)
- Invocation: `npm exec chrome-devtools-mcp@latest -- --browserUrl=http://127.0.0.1:<port> --experimentalScreencast`
- node v23.6.0 (Homebrew), macOS 15.7.2 (Apple Silicon)
- Chrome: current stable, attached via CDP (`--browserUrl`), long-lived instance

## Workload profile

The server runs for hours-to-days under an agent that repeatedly: takes snapshots, runs frequent `evaluate_script` polling, and records repeated `screencast_start` / `screencast_stop` cycles against a canvas/WebGL-heavy page.

## Observed behavior

Seven crash events over 19 days (2026-06-13 → 2026-07-01), all with the identical signature. Time-to-crash shrinks as automation load increases — process lifetime went **2.7 days → 12.7 h → 3.9 h** across the last three events, consistent with a load-proportional leak rather than a fixed-rate one.

Faulting thread (named `chrome-devtools-mcp`), from the macOS crash report:

```
libsystem_c.dylib     abort
node                  node::OOMErrorHandler(char const*, v8::OOMDetails const&)
node                  v8::internal::V8::FatalProcessOutOfMemory(...)
node                  v8::internal::Heap::FatalProcessOutOfMemory(char const*)
node                  v8::internal::Heap::CollectGarbage(AllocationSpace, GarbageCollectionReason, ...)
node                  v8::internal::IncrementalMarkingJob::Task::RunInternal()
[...]

PR #2312, opened 2026-07-07
feat: add screencast fps option

mturac:
Addresses #2291.

## Summary
- add an `--experimental-screencast-fps` option for screencast recording
- validate the value as a positive integer and imply `--experimental-screencast`
- pass the configured FPS through to Puppeteer's `screencast()` call
- update generated option docs and flag telemetry metadata

## Testing
- `git diff --check origin/main...HEAD`
- `npm run check-format`
- `npm run typecheck`
- `npm test -- --test-only tests/tools/screencast.test.ts tests/cli.test.ts`

### DIFF
--- a/README.md
+++ b/README.md
@@ -664,6 +664,11 @@ The Chrome DevTools MCP server supports the following configuration option:
   - **Type:** string
   - **Default:** `false`
 
+- **`--experimentalScreencastFps`/ `--experimental-screencast-fps`**
+  Frames per second to use for screencast recording. Lower values can reduce memory pressure on pages that produce frames faster than ffmpeg can encode them.
+  - **Type:** number
+  - **Default:** Uses Puppeteer's screencast default.
+
 - **`--categoryExperimentalWebmcp`/ `--category-experimental-webmcp`**
   Set to true to enable debugging WebMCP tools. Requires Chrome 149+ with the following flags: `--enable-features=WebMCP,DevToolsWebMCPSupport`
   - **Type:** boolean
--- a/README.md
+++ b/README.md
@@ -703,6 +703,11 @@ The Chrome DevTools MCP server supports the following configuration option:
   - **Type:** string
   - **Default:** `false`
 
+- **`--experimentalScreencastFps`/ `--experimental-screencast-fps`**
+  Frames per second to use for screencast recording. Lower values can reduce memory pressure on pages that produce frames faster than ffmpeg can encode them.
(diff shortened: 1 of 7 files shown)


mturac:
Validation refreshed on the current head:

```bash
git diff --check origin/main...HEAD
npm run test -- tests/tools/screencast.test.ts tests/cli.test.ts
npm run check-format
```

All three passed locally.

mturac:
Rebased this branch onto the current `main` and re-ran validation.

Validation:

- `git diff --check origin/main...HEAD`
- `npm run test -- tests/tools/screencast.test.ts tests/cli.test.ts`
- `npm run check-format`

nattallius:
Thanks for your pr! Since the direction for the linked issue #2291  hasn't been fully confirmed yet, let's hold off on merging or closing it for now. We can keep this on hold until we get definitive confirmation on the final direction

AuraquanTech:
I ran the measurement from my [#2291 comment](https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2291#issuecomment-4963047485) against this PR's current branch directly, using the built MCP server rather than calling Puppeteer on its own.

The server was driven through the MCP path from `new_page` to the aquarium demo and then `screencast_start` and `screencast_stop`, at 1920×1080 in headless mode. Memory was sampled inside the server process every 10 seconds after forced GC. Completed recordings were checked with ffprobe to confirm that they ran at the requested frame rate.

In one sweep, the observed post-GC RSS growth was approximately:

| Requested FPS | Observed RSS growth |
| ------------: | ------------------: |
| 10 | 2.5 MB/min, with no sustained accumulation during the 10 minute window |
| 15 | 0.9 MB/min, with no sustained accumulation during the 10 minute window |
| 20 | 15.5 MB/min |
| 30 | 615.6 MB/min, linear, reaching approximately 3.96 GB when the run was stopped externally after 5.8 minutes |

The growth was again concentrated in `external` and `arrayBuffers`, while the V8 heap remained approximately flat.

Under this machine and workload, the observed crossover was between 15 and 20 FPS. More importantly for this PR, `--experimental-screencast-fps` is correctly passed through the real MCP tool path, the generated recordings use the requested rate, and selecting 10 or 15 FPS prevented the severe accumulation seen at 30 FPS during the measured window.

Happy to repeat the measurement against a revised implementation or run additional durations if useful.

Ayrton R. Goodman

OrKoN (approved):

PR #2314, opened 2026-07-07
fix: screencast memory growth by adding backpressure for frame encoding 

syedaly:
Issue #2291 
Fixed the screencast memory blow-up by adding built-in backpressure to the recording pipeline. When the browser was producing frames faster than ffmpeg could encode them, the old behavior could let memory grow without control until the process crashed. I addressed this by limiting how many frames could remain pending at once, so the recorder no longer buffers endlessly under heavy load. I also lowered the default screencast frame rate and exposed it as a configurable option, making recordings safer on WebGL/canvas-heavy pages while still preserving the ability to adjust quality when needed.


nattallius:
Closing it, as it contains lots of unnecessary code, not related to a change

PR #2315, opened 2026-07-07
fix(telemetry): resolve enum values through nested schema wrappers

Liohtml:
## What

`generateToolMetrics` reads an enum parameter's values by unwrapping **at most
one** `optional` wrapper:

```ts
if (schema._def.values?.length > 0) {
  values = schema._def.values;
} else {
  values = schema._def.innerType._def.values; // only one level
}
```

An enum parameter wrapped in `.default().optional()` (or any nested
`optional`/`default`/`nullable` combination) therefore falls through to
`schema._def.innerType._def.values === undefined`, and `npm run update-metrics`
(run as part of `npm run gen`) crashes:

```
TypeError: Cannot read properties of undefined (reading '0')
    at validateEnumHomogeneity (build/src/telemetry/metricsRegistry.js:12)
    at generateToolMetrics (build/src/telemetry/metricsRegistry.js:59)
```

## Why it matters

`npm run gen` is required whenever a tool is added or changed (per
`CONTRIBUTING.md`). Declaring an enum parameter with a default — a natural,
documented zod pattern — silently breaks docs/metrics generation.

## Fix

Add `getEnumValues()` in `transformation.ts` that recursively unwraps
`optional` / `default` / `nullable` / `effects` wrappers, mirroring the existing
`getZodType()`, and use it in `generateToolMetrics`. Behavior is unchanged for
the existing bare-enum and single-`optional` cases.

## Tests

- `getEnumValues` unit tests: bare enum, `optional`, `default`, both orders of
  `default`+`optional`, and throws for a non-enum type.
- `generateToolMetrics` regression test with `zod.enum(...).default(...).optional()`.
- `npm run typecheck`, the telemetry tests, and `npm run check-format` pass.

---
First-time contributor here — happy to sign the CLA.

### DIFF
--- a/src/telemetry/metricsRegistry.ts
+++ b/src/telemetry/metricsRegistry.ts
@@ -10,6 +10,7 @@ import {
   transformArgName,
   transformArgType,
   getZodType,
+  getEnumValues,
   PARAM_BLOCKLIST,
   stripUnderscoreBeforeNumber,
 } from './transformation.js';
@@ -103,13 +104,7 @@ export function generateToolMetrics(tools: ToolDefinition[]): ToolMetric[] {
       let argType = transformArgType(zodType);
 
       if (argType === 'enum') {
-        let values;
-        if (schema._def.values?.length > 0) {
-          values = schema._def.values;
-        } else {
-          values = schema._def.innerType._def.values;
-        }
-        argType = validateEnumHomogeneity(values);
+        argType = validateEnumHomogeneity(getEnumValues(schema));
       }
 
       args.push({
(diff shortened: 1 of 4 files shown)


nattallius:
LGTM, but @yulunz can you please check it as well?

Liohtml:
Thanks for taking a look, @nattallius! Happy to address anything @yulunz flags — the change is small and isolated (a recursive enum-value unwrap in the telemetry helper) with unit + regression tests.

yulunz (approved):
@Liohtml looks great, thanks!

please wait for approval from @nattallius too

nattallius (approved):

PR #2318, opened 2026-07-07
chore: report exact bundled package versions during release

zyzyzyryxy:
At bundle time, we know exactly which version of the package is installed, we should report exact version instead of a range from package.json.

This change parses package-lock.json to obtain this information.

This approach aligns better with upstream security analyzers relying on dependency version information.

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -50,6 +50,15 @@ const {devDependencies = {}} = JSON.parse(
 // special case for puppeteer, from which we only bundle puppeteer-core
 devDependencies['puppeteer-core'] = devDependencies['puppeteer'];
 
+const packageLock = JSON.parse(
+  fs.readFileSync(path.join(process.cwd(), 'package-lock.json'), 'utf-8'),
+);
+
+function getResolvedVersion(packageName) {
+  const packageEntry = packageLock.packages?.[`node_modules/${packageName}`];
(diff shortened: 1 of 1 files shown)


rollup.config.mjs
```diff
@@ -50,6 +50,15 @@ const {devDependencies = {}} = JSON.parse(
 // special case for puppeteer, from which we only bundle puppeteer-core
 devDependencies['puppeteer-core'] = devDependencies['puppeteer'];
 
+const packageLock = JSON.parse(
+  fs.readFileSync(path.join(process.cwd(), 'package-lock.json'), 'utf-8'),
+);
+
+function getResolvedVersion(packageName) {
+  const packageEntry = packageLock.packages?.[`node_modules/${packageName}`];
```
OrKoN, line 58: could it be that package-lock lists multiple versions and we are actually bundling a different one, e.g., a transitive dependency? Can we look at package.json next to the modulePath where we determine the packageName?

zyzyzyryxy: We are to report only direct dependencies in `bundled-packages.json` (and we do so by filtering by `Object.entries(devDependencies)`). If there is a different version of the package bundled as a transitive dependency too, we should get vulnerability reports through its dependency chain.

OrKoN (approved):

PR #2321, opened 2026-07-08
refactor: Use array instead of Map for idToClassKey

dinfuehr:
Switches idToClassKey to an array since ids are strictly incremented from 1. This also allows us to drop the id generator use.


Lightning00Blade (approved):

dinfuehr:
Thanks for the review!

PR #2322, opened 2026-07-08
fix: prevent renderer crash when upload_file targets file chooser proxy elements

syedaly:
Issue # 2310 
## Problem
Calling `upload_file` on a button or other proxy element that opens a file chooser 
crashes the Chrome renderer on Windows with `RESULT_CODE_KILLED_BAD_MESSAGE`, 
particularly in headful Chrome instances. The crash occurs intermittently due to a race 
condition in the file-chooser interception IPC.

The tool incorrectly reports success ("File uploaded") even when the crash happens 
(since the crash is asynchronous), causing subsequent tool calls to fail with misleading 
"element no longer exists" errors that mask the real problem: the page/tab has crashed.

**Related Issue:** #2310

## Root Cause
The original implementation used `Promise.all([waitForFileChooser(), click()])` as a 
fallback when `uploadFile()` failed. This pattern creates a race condition on Windows:
- `waitForFileChooser()` listener isn't fully established
- `click()` fires immediately in parallel
- File chooser dialog opens and triggers IPC before listener is ready
- Chromium kills the renderer to prevent undefined behavior

## Solution
1. **Detect element type upfront** - Check if element is an actual `<input type="file">` 
   using `evaluate()` before attempting any upload
2. **Separate code paths:**
   - **File input elements:** Direct `uploadFile()` call (existing behavior)
   - **Proxy elements (buttons, etc.):** Sequential execution:
     - Set up file chooser listener
     - Click element  
     - Accept file via listener
3. **Remove nested try-catch** - Eliminates the problematic fallback pattern that 
   attempted `uploadFile()` on non-file elements

## Impact
- ✅ Eliminates the Windows renderer crash on chooser-opening elements
- ✅ Maintains backward compatibility with direct file input elements
- ✅ Clearer error handling - actual failures now surface appropriate error messages
- ✅ Prevents misleading success messages when crashes occur

## Testing
The fix handles both documented use cases:
1. Direct file input elements: `<input type="file">`
2. Proxy elements: Buttons/links that call `input.click()` to open chooser


nattallius:
Hi, thank you for your contribution and for taking the time to address this issue.

However, this pull request currently combines several distinct modifications into a single submission. 
To ensure a clean commit history and streamline the review process, we would appreciate if you could split these changes into separate, targeted pull requests for each individual fix. Consequently, we will close this pull request for now, but we look forward to reviewing your modular submissions if you would like to deliver these updates.

PR #2323, opened 2026-07-08
feat: Support filter with heap snapshots aggregates

dinfuehr:
This exposes DevTools named filters in get_heapsnapshot_details and get_heapsnapshot_class_nodes tools. This allows the client to list e.g. all objects retained through detached DOM objects.


Lightning00Blade (approved):

dinfuehr:
Thanks for the review!

PR #2326, opened 2026-07-08
feat: restart active performance traces

### LINKED ISSUE #808: Stop running trace when starting a new trace

In an agentic LLM workflow, every step counts towards increasing the load on the AI and its context window. Performance traces are a particularly sore point for LLMs, as only one can be active at any given time. Currently, the MCP server throws an error if attempting to start a new trace when one already exists, even if that trace was from an entirely different conversation completely unrelated to the present one.

Rather than throwing an error and interrupting the thought processes of an LLM when running `performance_start_trace`, close any existing performance trace and include a notice that this action happened in the response given by the start tool.

N/A

### Additional context

_No response_

godququ5-code:
## Summary

- stop any currently running performance trace before starting a new one
- add a short notice when an existing trace is replaced
- cover the restart path with a regression test

Fixes #808.

## Validation

- npm run test tests/tools/performance.test.ts
- npm run build
- npm run check-format

### DIFF
--- a/src/tools/performance.ts
+++ b/src/tools/performance.ts
@@ -51,10 +51,7 @@ export const startTrace = definePageTool({
   verifyFilesSchema: ['filePath'],
   handler: async (request, response, context) => {
     if (context.isRunningPerformanceTrace()) {
-      response.appendResponseLine(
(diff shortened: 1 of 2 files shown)


OrKoN:
Thanks for the PR but we are still collecting feedback on the linked issue.

PR #2327, opened 2026-07-08
fix: Wait until daemon is started

dinfuehr:
startDeamon was only waiting until until daemon.pid existed. But it is possible that daemon.pid was created but the daemon wasn't ready yet. This was observable through flaky test runner failures. Therefore startDaemon now also waits until the daemon successfully replies to an initial status message.


OrKoN (approved):

OrKoN:
Thanks!

dinfuehr:
Thanks for the review!

PR #2333, opened 2026-07-09
refactor: clean up page management

OrKoN:
- remove `#pages` and Puppeteer page getters
- rename methods for clarity

### DIFF
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -994,11 +994,14 @@ Call ${handleDialog.name} to handle it before continuing.`);
       const allPages = context.getPages();
 
       const {regularPages, extensionPages} = allPages.reduce(
-        (acc: {regularPages: Page[]; extensionPages: Page[]}, page: Page) => {
-          if (page.url().startsWith('chrome-extension://')) {
-            acc.extensionPages.push(page);
+        (
+          acc: {regularPages: McpPage[]; extensionPages: McpPage[]},
+          mcpPage: McpPage,
+        ) => {
+          if (mcpPage.pptrPage.url().startsWith('chrome-extension://')) {
+            acc.extensionPages.push(mcpPage);
           } else {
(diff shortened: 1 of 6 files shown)


src/McpResponse.ts
```diff
@@ -994,11 +994,14 @@ Call ${handleDialog.name} to handle it before continuing.`);
       const allPages = context.getPages();
 
       const {regularPages, extensionPages} = allPages.reduce(
-        (acc: {regularPages: Page[]; extensionPages: Page[]}, page: Page) => {
-          if (page.url().startsWith('chrome-extension://')) {
-            acc.extensionPages.push(page);
+        (
+          acc: {regularPages: McpPage[]; extensionPages: McpPage[]},
```
Lightning00Blade, line 998: Maybe the name should be different here as well?

src/McpResponse.ts
```diff
@@ -1509,10 +1512,10 @@ function createStructuredPage(
     selected: boolean;
     isolatedContext?: string;
   } = {
-    id: context.getPageId(page),
-    url: page.url(),
+    id: mcpPage.id,
+    url: mcpPage.pptrPage.url(),
     title,
-    selected: context.isPageSelected(page),
+    selected: context.isPageSelected(mcpPage.pptrPage),
```
Lightning00Blade, line 1518: Should we change this too as the `this.#selectedPage` is already a McpPage?

src/McpResponse.ts
OrKoN: regularPages and extensionPages sounds find to me. I think eventually we should have no references to pptrPage in the McpContext.

src/McpResponse.ts
OrKoN: yes, but this is in McpResponse and probably not in this PR

Lightning00Blade (approved):

PR #2235, opened 2026-06-19
feat: add experimentalGcfFormat flag for GCF-encoded tool responses

blackwell-systems:
## Summary

Add `--experimentalGcfFormat` as a hidden experimental flag alongside the existing `--experimentalToonFormat`. When enabled, tool responses encode structured data using GCF (Graph Compact Format) instead of formatted JSON. No changes to existing behavior or default output.

### Benchmarked on Chrome DevTools data shapes

Data shapes verified from source code and test snapshots (`ConsoleMessageConcise`, `NetworkRequestConcise`, `HeapSnapshotFormatter.toJSON()`, `SnapshotFormatter.toJSON()`):

| Dataset | JSON | TOON | GCF | TOON vs JSON | GCF vs JSON | GCF vs TOON |
|---------|------|------|-----|-------------|-------------|-------------|
| Console (100) | 2,534 | 1,716 | 1,663 | 32.3% | 34.4% | +3.1% |
| Console (500) | 12,879 | 8,761 | 8,475 | 32.0% | 34.2% | +3.3% |
| Network (100) | 3,287 | 1,776 | 1,682 | 46.0% | 48.8% | +5.3% |
| Network (500) | 16,547 | 8,936 | 8,442 | 46.0% | 49.0% | +5.5% |
| Heap (100) | 2,210 | 797 | 753 | 63.9% | 65.9% | +5.5% |
| DOM (3x3) | 282 | 295 | 244 | -4.6% | 13.5% | +17.3% |
| DOM (4x3) | 396 | 464 | 407 | -17.2% | -2.8% | +12.3% |

TOON increases token count on DOM snapshots (-4.6% to -17.2% vs JSON). GCF saves tokens on every data type.

### Console message corruption

TOON's decoder crashes on console messages containing bracket-colon patterns, which are standard browser output:

```
[Error]: net::ERR_CONNECTION_REFUSED
[React DevTools]: Component rendered 3 times
[Violation]: Forced reflow while executing JavaScript took 42ms
[Performance]: Long task detected (duration: 234ms)
```

10 of 20 console messages fail TOON round-trip. GCF: zero failures. While Chrome DevTools MCP encodes only (no decode), the structurally ambiguous output can affect downstream consumers.

### LLM comprehension

GCF scores 100% on general structured data and 90.7% on adversarial payloads across GPT-4o, GPT-5.5, Claude, and Gemini. JSON scores 53.6%. TOON scores 68.5%. (1,700+ evaluations.)

[...]

PR #2343, opened 2026-07-10
test: add unit tests for the GCF data format (follow-up to #2235)

blackwell-systems:
## Summary

Follow-up to #2235, which added `--experimentalDataFormat=gcf`. During that review the request was to add unit tests as the prerequisite for GCF being considered for the default/non-experimental path:

> "Please also consider adding some unit tests - they are not strictly required for experimental feature, but we'll need them in case we do decide to make GCF the default. This might be a separate PR I think."

This PR adds that coverage. It introduces one new test file, `tests/McpResponseGcf.test.ts` (8 tests), with no changes to `src/`.

## Scope

Covers every `McpResponse`-level surface that `compactEncode` touches:

**Encode surfaces — each round-trips losslessly back to `structuredContent`:**
- Page snapshot
- Network requests
- Console messages (including the bracket-colon shapes, e.g. `[Error]: net::ERR_CONNECTION_REFUSED`)
- Heap snapshot class diffs (array of flat objects)
- Detailed heap class diff (object with nested numeric arrays)

**Safety / robustness:**
- `default` format emits no GCF header and still renders the built-in formatter output (guards against a vacuous pass)
- `structuredContent` is byte-identical with and without GCF — clients consume `structuredContent`, so enabling GCF must only change the text representation
- Values containing delimiters and GCF sentinels (`|` `,` `"` `~` `-` `@` `##`, newlines, tabs, unicode, empty strings, typed-literal lookalikes like `true`) round-trip losslessly

## Rigor

- Tests decode the emitted block with `@blackwell-systems/gcf` and deep-equal it against `structuredContent`, so they assert real losslessness rather than snapshotting opaque bytes — a regression in the encode path fails them.
- One test additionally pins the **exact wire bytes** (golden assertion), so a symmetric encode+decode bug can't round-trip its way past it.
- All 8 pass; verified stable across repeated runs. Lint, type-check, and build are clean. The rest of the suite is unaffected (the only failures locally are the pre-existing e2e tests that require a system Chrome `stable` channel).

## Notes

- No `src/` changes; no new runtime dependencies (`@blackwell-systems/gcf` is already the optional peer dep added in #2235).
- Tests use the existing `withMcpContext` harness and mock helpers, matching the conventions in `tests/McpResponse.test.ts`.

cc @zyzyzyryxy — follow-up to the unit-test request in #2235.

### DIFF
--- a/tests/McpResponseGcf.test.ts
+++ b/tests/McpResponseGcf.test.ts
@@ -1,0 +1,345 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'node:assert';
+import {describe, it} from 'node:test';
+
+import {decodeGeneric} from '@blackwell-systems/gcf';
+
+import {getMockRequest, getTextContent, html, withMcpContext} from './utils.js';
+
+/**
+ * The GCF-encoded block is pushed last in the relevant response branch, so it
+ * runs from the `GCF profile=` header to the end of the joined text. Slice it
+ * out so it can be decoded back to the structured value it was produced from.
+ */
+function extractGcfBlock(text: string): string {
+  const start = text.indexOf('GCF profile=');
+  assert.notStrictEqual(
+    start,
+    -1,
+    `expected a GCF header in response text, got:\n${text}`,
+  );
+  return text.slice(start).trim();
+}
+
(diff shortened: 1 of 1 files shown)


zyzyzyryxy:
Thanks for your contribution! There is a decision to keep GCF support as experimental for now, but I'll keep this PR in mind as we might revisit that decision later this year.

Lightning00Blade:
I will close this for now while we evaluate things.

PR #2344, opened 2026-07-10
feat: let evaluate_script skip stable DOM wait

### LINKED ISSUE #2338: Allow read-only `evaluate_script` calls to skip the stable DOM wait

Currently, every `evaluate_script` runs through `waitForEventsAfterAction()`: A ~100ms navigation probe plus a stable DOM wait (no mutations for 100ms, capped at 3s – `src/WaitForHelper.ts`). For interaction tools that's the right default, but scripts that only read state pay it too.

On a quiet page that's a fixed ~200ms per eval; on a page that never goes quiet (JS-driven animations, live UI updates) every eval sits out the full 3s cap.

And since the stable DOM timeout is swallowed (`logger?.(error)`), nothing points to where the time goes – we spent a while believing the tool was inherently slow, and ended up cramming all reads into one giant eval per round-trip to work around it.

Repro: Run `setInterval(() => document.body.setAttribute('x', Date.now()), 30)` on any page – every `evaluate_script` now takes ~3.1s.

A `waitForStableDom` parameter on `evaluate_script`, defaulting to `true` – `false` skips just the stable DOM wait and keeps the navigation probe and dialog handling. The parameter description can tell the model when to pass `false`.

Getting the flag wrong is then mostly harmless: A script that unexpectedly navigates is still awaited (and the navigation reported), an `alert()` is still auto-handled.

I prototyped this (~15 lines through `script.ts` -> `McpPage` -> `WaitForHelper`) and verified the misuse cases above:

| read-only eval on | default | `waitForStableDom: false` |
| --- | --- | --- |
| quiet page | ~205ms | ~103ms |
| page mutating every 30ms | ~3107ms | ~104ms |

Happy to send the PR.

- Auto-detecting read-only scripts – doesn't work, as mutations can trail the eval by a paint or microtask, which is exactly what the quiet window is guarding
- Making the wait timings configurable via CLI flag – global, and still pays the wait on every call

### Additional context

Measured on 1.5.0 (macOS, Chrome 150 headless) – the wait mechanics are unchanged on current `main`.

mturac:
## Summary
- add a `waitForStableDom` option to `evaluate_script` so read-only scripts can skip the stable DOM wait
- keep navigation and dialog handling on the existing post-action wait path
- update the generated tool reference

Fixes #2338

## Testing
- `npm run typecheck`
- `npm run test -- tests/tools/script.test.ts`
- `npm run check-format`
- `git diff --check`


thomasbachem:
Heads-up: `npm run gen` also regenerates `src/telemetry/tool_call_metrics.json` and `src/bin/chrome-devtools-cli-options.ts` for the new param – as-is check-docs will fail.

mturac:
Thanks for the heads-up. I re-ran the generation path on the current head `4e790e92a834` and confirmed the generated docs, CLI options, and metrics files are up to date.

Validation:

```sh
npm run gen
git diff --exit-code docs/tool-reference.md src/bin/chrome-devtools-cli-options.ts src/telemetry/tool_call_metrics.json
npm run check-format
npm run test:no-build -- tests/tools/script.test.ts
```

All pass locally.

Lightning00Blade:
Thanks for the PR. 
The PR's https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2346 has 1 more test and the test is a little bit better as It verifies that the underlying wait code is not called.

PR #2345, opened 2026-07-11
fix: keep page ids unique across browser reconnects

### LINKED ISSUE #2339: Page ids silently rebind to different pages after a browser restart
### Description of the bug

In `--browserUrl` mode the server survives a browser restart nicely – tools error cleanly while the browser is down, and the next call reconnects. But the reconnect builds a fresh `McpContext`, so page ids start over at 1: Whatever pages the restarted browser opens get the same ids the old pages had. A client that re-uses an id from before the restart doesn't get an error – it gets a different page.

The selection fallback note from #2308 doesn't fire either, since the fresh context treats the reconnect as a first connect – so there is no signal at all. And when the browser bounces between two tool calls (a crash, a "Relaunch to update", an app relaunch), the client never even sees a failing call in between. It's the same failure mode as #2304 one level up – for an agent, plausible results from the wrong page are worse than an error.

We happen to hit this daily driving an Electron app that relaunches during development, but there is nothing Electron-specific about it – any browser restart under a long-lived session does the same, and the reproduction below is plain headless Chrome.

### Reproduction

Verified with plain headless Chrome:

1. Connect via `--browserUrl`, `new_page` a second page – the listing shows `1: pageA` / `2: pageB [selected]`
2. Kill the browser and start a new one on the same port with two different pages (C, D)
3. `select_page` with the stale id 2 – succeeds silently, `[selected]` lands on pageD
4. `evaluate_script` with `() => document.title` – returns "pageD", no error, no note

### Expectation

Page ids were introduced as stable handles (#688) – stale ones should fail loudly rather than resolve to unrelated pages, consistent with the "No page found" contract:

- Seed the new context's page id counter from the previous one on reconnect – an id from before the restart is then simply not found, and the agent re-lists and recovers in one turn
- Ideally also surface a one-time note in the first response after a reconnect ("reconnected to the browser, page ids have changed"), like the #2308 fallback note

### MCP configuration

```json
{
	"command": "npx",
	"args": ["chrome-devtools-mcp@1.5.0", "--browserUrl", "http://127.0.0.1:9339"]
}
```

1.5.0 – reproduced identically on current `main`

### Chrome version

Chrome 150 (headless) – originally observed against Electron 43 (Chromium 150)

### Coding agent version

n/a – reproduces deterministically with a scripted MCP stdio client

### Model version

_No response_

### Chat log

_No response_

### Node version

v22.22.2

### Operating system

macOS

### Extra checklist

thomasbachem:
Closes #2339

Reconnecting after a browser restart builds a fresh `McpContext`, which restarted the page id counter at 1 – ids from before the restart silently resolved to unrelated pages of the new browser (repro in #2339).

Two changes:

- The new context continues the id counter where the previous one left off (`startingPageId` option), so a stale id now fails with the existing "No page found" error and the agent re-lists
- The first response after a reconnect carries a one-time note ("the browser was restarted or reconnected since the last call. Page ids have changed..."), like the #2308 fallback note. Since tool errors run through the same response formatting, the note shows up together with the very error the stale id produces

Verified against the #2339 repro: the stale `select_page` now returns the note plus "No page found", and the next listing shows the new browser's pages under fresh ids.

One thing I noticed but left alone: The replaced context isn't `dispose()`d on reconnect – pre-existing, and everything it holds is tied to the dead browser anyway. Happy to add that here if you'd like.

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -56,7 +56,16 @@ interface McpContextOptions {
   // capability. When false (default), file-writing tools are restricted to the
   // OS temp directory. When true, the previous permissive behavior is restored.
   allowUnrestrictedPaths?: boolean;
+  // Whether this context replaces a previous one after a browser reconnect.
+  // Surfaces a one-time note in the next response.
+  reconnected?: boolean;
 }
+
+// Page ids are handed out from a process-wide counter so they stay unique
+// across all contexts, in particular across browser reconnects. An id issued
+// before a reconnect then fails to resolve instead of hitting an unrelated
+// page of the reconnected browser.
+let nextPageId = 1;
 
 export class McpContext implements Context {
   browser: Browser;
@@ -79,7 +88,7 @@ export class McpContext implements Context {
   #screenRecorderData: {recorder: ScreenRecorder; filePath: string} | null =
     null;
 
-  #nextPageId = 1;
+  #reconnectNotice = false;
   #extensionPages = new WeakMap<Target, Page>();
 
   #extensionServiceWorkerMap = new WeakMap<Target, string>();
@@ -110,6 +119,7 @@ export class McpContext implements Context {
     this.#locatorClass = locatorClass;
     this.#options = options;
     this.#allowUnrestrictedPaths = options.allowUnrestrictedPaths ?? false;
+    this.#reconnectNotice = options.reconnected ?? false;
 
     this.#serviceWorkerConsoleCollector = new ServiceWorkerConsoleCollector(
       this.browser,
@@ -184,6 +194,10 @@ export class McpContext implements Context {
     const context = new McpContext(browser, logger, opts, locatorClass);
     await context.#init();
     return context;
+  }
+
+  static resetPageIdsForTesting(): void {
+    nextPageId = 1;
   }
 
   roots(): Root[] {
@@ -359,6 +373,17 @@ export class McpContext implements Context {
     return page;
   }
 
+  /**
+   * Returns true once if this context was created by reconnecting after the
+   * previous browser connection was lost, so the next response can surface a
+   * note. Cleared on first call.
+   */
+  consumeReconnectNotice(): boolean {
+    const notice = this.#reconnectNotice;
+    this.#reconnectNotice = false;
+    return notice;
+  }
+
   getPageById(pageId: number): McpPage {
     // Resolve only among listed pages (`getPages()`), so an id that
     // `list_pages` never showed cannot be targeted (e.g. a `devtools://`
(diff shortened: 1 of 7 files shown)


src/McpResponse.ts
```diff
@@ -917,6 +917,11 @@ export class McpResponse implements Response {
     }
 
     const response = [];
+    if (context.consumeReconnectNotice()) {
```
OrKoN: Instead of consuming the notice in the McpResponse, let's have a setter on McpResponse that turns on the notice and consume the notice in ToolHandler::handle.

src/McpContext.ts
```diff
@@ -56,6 +56,13 @@ interface McpContextOptions {
   // capability. When false (default), file-writing tools are restricted to the
   // OS temp directory. When true, the previous permissive behavior is restored.
   allowUnrestrictedPaths?: boolean;
+  // The first page id this context assigns. Used to keep ids unique across
+  // browser reconnects, so ids from before a reconnect don't resolve to
+  // unrelated pages of the new browser.
+  startingPageId?: number;
```
OrKoN: Instead of requiring to pass the startingPageId, let's move nextPageId to the module level variable so that all contexts within the process for sure re-use the same counter?

tests/McpContext.test.ts
```diff
@@ -222,6 +227,49 @@ describe('McpContext', () => {
...
+        // The reconnect note is surfaced once, then cleared.
+        const first = await new McpResponse({} as ParsedArguments).handle(
+          'test',
+          context,
+        );
+        assert.ok(
+          JSON.stringify(first.content).includes('Page ids have changed'),
+          'first response should carry the reconnect note',
+        );
+        const second = await new McpResponse({} as ParsedArguments).handle(
```
OrKoN: let's test McpResponse producing the notice in the McpResponse.test.ts and only verifies that the McpResponse is configured to produce such as notice.

tests/McpContext.test.ts
```diff
@@ -222,6 +227,49 @@ describe('McpContext', () => {
...
+          startingPageId: 5,
+          reconnected: true,
+        },
+        Locator,
+      );
+      try {
+        // Ids continue where the previous context left off, so an id from
+        // before the reconnect does not resolve to an unrelated page.
+        const page = await context.newPage();
+        assert.ok(page.id >= 5, 'ids should continue from startingPageId');
```
OrKoN: instead of using specific ids, let's verify that the same page in two different contexts has the same id.

OrKoN (changes requested):
Thanks, I think it is a good idea to add this. I have left a few comments.

src/McpResponse.ts
```diff
@@ -917,6 +917,11 @@ export class McpResponse implements Response {
...
+      response.push(
```
OrKoN, line 932: we also need to include this into structuredContent

thomasbachem:
Thanks – all five applied. Two notes:

- `nextPageId` never resets now, so ids stay unique across all contexts in the process. Tests keep numbering from 1 via a `resetPageIdsForTesting()` called in the `withMcpContext` fixture.
- On the id test: The page-to-id map is per-context, so the same page gets a fresh id in a new context. The test therefore asserts a stale id no longer resolves and new ids continue past it – asserting "same page, same id" literally would need the map at module level too. Happy to do that if that's the semantics you want.

OrKoN (approved):

OrKoN:
Thanks!

PR #2346, opened 2026-07-11
feat: allow evaluate_script to skip the stable DOM wait

### LINKED ISSUE #2338: Allow read-only `evaluate_script` calls to skip the stable DOM wait

thomasbachem:
Closes #2338

Adds a `waitForStableDom` parameter to `evaluate_script` (default `true`). Passing `false` skips just the stable DOM wait in `waitForEventsAfterAction()` – the navigation probe and dialog handling stay, so a script that unexpectedly navigates is still awaited (and the navigation reported), and an `alert()` is still auto-handled. The page path passes the flag through, and service worker evals now always skip the stable DOM wait (workers can't interact with the DOM – thanks @Lightning00Blade).

Read-only evals drop from ~205ms to ~103ms on quiet pages, and from ~3.1s to ~103ms on pages that never go quiet – numbers and repro in #2338.

This is the branch the issue promised – I see #2344 beat me to posting, so take whichever is easier to review. Differences here: complete `npm run gen` artifacts (`tool_call_metrics.json` + `chrome-devtools-cli-options.ts`, so check-docs passes), a skip test that spies `WaitForHelper.prototype.waitForStableDom` (fails if the helper ignores the flag, not only if the option isn't passed through), and a test pinning the safety property – a script that navigates with the flag off still gets its navigation awaited and reported.


hotrod6531-sudo (approved):

src/tools/script.ts
```diff
@@ -95,7 +102,7 @@ Example with arguments: \`(el) => el.innerText\`
                 context,
               });
             },
-            {handleDialog: dialogAction ?? 'accept'},
+            {handleDialog: dialogAction ?? 'accept', waitForStableDom},
```
Lightning00Blade: We should be able to pass `false` here without any changes as workers can't interact with the DOM.

src/tools/script.ts
```diff
@@ -52,6 +52,12 @@ Example with arguments: \`(el) => el.innerText\`
         .describe(
           'Handle dialogs while execution. "accept", "dismiss", or string for response of window.prompt. Defaults to accept.',
         ),
+      waitForStableDom: zod
+        .boolean()
+        .optional()
+        .describe(
+          'Whether to wait for the DOM to settle after the script ran. Pass false when the script only reads state and does not change the page, for a faster response. Navigations triggered by the script are still awaited. Defaults to true.',
```
Lightning00Blade: I think this is a little bit verbose. Maybe this will be sufficient:
```suggestion
          'Whether to wait for the DOM to settle. Pass false if the script only read data. Defaults to true.',
```

Lightning00Blade (commented):
Thanks for the PR, I left some small comments.

src/tools/script.ts
thomasbachem: Done – the worker path now always passes `false`.

src/tools/script.ts
thomasbachem: Applied – took your wording with one small tweak ("reads").

Lightning00Blade: Yeah thats a type on my end, thanks.

Lightning00Blade (approved):

OrKoN (approved):

thomasbachem:
Sorry, that dequeue was mine. The branch still called `response.handle('test', context)`, and #2392 dropped the leading `toolName` parameter after this PR last ran CI, so the merge commit failed to compile.

I've rebased onto `main` and updated the call to `response.handle(context)` – still a single commit, and the only change from what you reviewed is that one line. `npm run build`, `npm run test tests/tools/script.test.ts` and `npm run format` all pass locally against current `main`.

The workflow runs need approval whenever one of you gets a moment, then it should be good for the queue.

src/tools/script.ts
hotrod6531-sudo: I'm kind of just started so if you're talking to me I appreciate if you're trying to help me I appreciate it I apologize for the ignorance but I'm learning

src/tools/script.ts
hotrod6531-sudo: Thank you again

Issue #2309, opened 2026-07-06
Windows: system-wide repeating/stuck key (space) after chrome-devtools-mcp usage, persists after keyboard disconnect, requires OS restart


Fakiros:
## Summary

After periods of using `chrome-devtools-mcp` (connected as an MCP server to Claude Code, running headful on a Windows client machine), the Windows machine occasionally enters a state where a keystroke (space bar) starts **auto-repeating system-wide** — in every application that has focus, not just the automated browser tab. Clicking into a different window/app does not stop it; the repeated space keeps appearing wherever focus goes next.

Critically: **physically disconnecting the keyboard does not stop the repeat**, which suggests the "held key" state is being tracked at the OS input level as a synthetic/injected event, not coming from real hardware. The only remedy found so far is a full OS restart. This has now recurred 3 separate times over a period of working with the MCP server.

## Environment

- `chrome-devtools-mcp`, connected over HTTP transport (`http://<windows-host>:8932/mcp`) from a remote Claude Code session (Linux host) via SSH/Remote-SSH
- Windows client machine, real (headful) Chrome window driven by the server
- `@playwright/mcp` (`microsoft/playwright-mcp`) is also configured on the same Windows machine on a separate port (8931), so I cannot 100% rule out an interaction between the two, but chrome-devtools-mcp is the one with active tool usage in this session's history around the time of at least one occurrence.

## Hypothesis (unconfirmed)

If a keyboard tool (e.g. a key-press action) dispatches a `keyDown` for Space but, due to an exception, timeout, or lost connection, the matching `keyUp` is never sent, the OS itself may continue to treat the key as physically held and apply its own key-repeat — which would explain why the repeat:
- targets whatever window currently has OS focus (not just the browser), and
- continues even with the physical keyboard unplugged.

This would align with the input-reliability issues already reported in #2199 (injected input silently stops reaching a tab) and #2290 (server can steal foreground window focus) — both suggest the input/focus state tracking between the MCP server and the OS/browser isn't always consistent.

## Reproduction

Not yet reliably reproducible on demand — it happens "after some time" / under conditions I haven't fully isolated. I don't have a minimal repro script yet. Filing this now because of severity (requires a full OS restart each time) in case others have hit the same thing or maintainers have insight into whether any code path sends a keyDown without a guaranteed matching keyUp (e.g., on an error/timeout branch).

## Impact

High — the affected machine becomes effectively unusable (phantom input in every focused window) until a full restart.

[...]

PR #2347, opened 2026-07-11
fix: release held modifiers when press_key key event fails

Liohtml:
## What

`press_key` currently presses each modifier down, presses the main key, then releases the modifiers — in three sequential steps with no `try/finally`:

```ts
for (const modifier of modifiers) {
  await page.pptrPage.keyboard.down(modifier);
}
await page.pptrPage.keyboard.press(key);      // if this rejects…
for (const modifier of modifiers.toReversed()) {
  await page.pptrPage.keyboard.up(modifier);  // …this never runs
}
```

If `keyboard.press(key)` rejects — a CDP hiccup, a target crash, or a dropped connection — the release loop is skipped and the modifier keys are left **logically held down in the browser**. `waitForEventsAfterAction` re-throws the action error, so nothing downstream releases them either.

This is the unpaired-`keyDown` class of defect asked about in #2309 (*"whether any code path sends a keyDown without a guaranteed matching keyUp on an error/timeout branch"*). This PR fixes the one concrete instance of it in this repo.

## Fix

Wrap the down/press sequence in `try/finally` and track which modifiers were actually pressed, releasing each held modifier even when the main press throws. Only modifiers whose `keyboard.down()` succeeded are released, so a failure *while* pressing a modifier doesn't emit a spurious `keyUp`.

## Test

Adds a regression test (real browser, keydown/keyup logging) that injects a `press()` failure mid-sequence and asserts both modifiers are still released. It fails on `main` (`['dControl','dShift']` — no keyups) and passes with the fix (`['dControl','dShift','uShift','uControl']`).

Verified locally: full `tests/tools/input.test.ts` suite passes, `npm run typecheck`, eslint, and prettier all clean.

## Scope note re: #2309

I want to be precise about what this does and does not address. This closes a **browser-level** stuck-key path: leaked keys here live in Chromium's input state (CDP `Input.dispatchKeyEvent` is injected into the renderer), so the observable effect is a modifier stuck **within the driven page**. The report in #2309 is a bare `Space` that repeats **system-wide** and survives physically unplugging the keyboard — that symptom is at the OS input layer, which CDP-injected input doesn't route through, so I don't claim this fully explains that case (details and a non-reboot workaround are in a comment on the issue). Still, an unpaired keyDown on an error branch is a real defect worth closing on its own, and it's exactly the code path the issue asked to audit.

Prepared with AI assistance (Claude Code) and verified against a local build before submission.

### DIFF
--- a/src/tools/input.ts
+++ b/src/tools/input.ts
@@ -526,12 +526,20 @@ export const pressKey = definePageTool({
     const [key, ...modifiers] = tokens;
 
     const result = await page.waitForEventsAfterAction(async () => {
-      for (const modifier of modifiers) {
-        await page.pptrPage.keyboard.down(modifier);
-      }
-      await page.pptrPage.keyboard.press(key);
-      for (const modifier of modifiers.toReversed()) {
-        await page.pptrPage.keyboard.up(modifier);
+      const heldModifiers: KeyInput[] = [];
+      try {
+        for (const modifier of modifiers) {
+          await page.pptrPage.keyboard.down(modifier);
+          heldModifiers.push(modifier);
+        }
+        await page.pptrPage.keyboard.press(key);
+      } finally {
+        // Release every modifier that was successfully pressed, even if a
+        // later key event throws. Otherwise a failed press leaves modifiers
+        // logically held down in the browser (see #2309).
+        for (const modifier of heldModifiers.toReversed()) {
+          await page.pptrPage.keyboard.up(modifier);
+        }
       }
     });
 
--- a/tests/tools/input.test.ts
+++ b/tests/tools/input.test.ts
@@ -9,6 +9,8 @@ import fs from 'node:fs/promises';
 import path from 'node:path';
 import {describe, it} from 'node:test';
 
+import sinon from 'sinon';
+
 import type {ParsedArguments} from '../../src/bin/chrome-devtools-mcp-cli-options.js';
 import {McpResponse} from '../../src/McpResponse.js';
 import {TextSnapshot} from '../../src/TextSnapshot.js';
@@ -1362,7 +1364,7 @@ describe('input', () => {
 
     it('releases held modifiers when the main key press fails', async () => {
       await withMcpContext(async (response, context) => {
-        const page = context.getSelectedPptrPage();
+        const page = context.getSelectedMcpPage().pptrPage;
         await page.setContent(
           html`<script>
             logs = [];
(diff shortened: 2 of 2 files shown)


tests/tools/input.test.ts
```diff
@@ -1359,5 +1359,55 @@ describe('input', () => {
...
+            document.addEventListener('keyup', e => logs.push('u' + e.key));
+          </script>`,
+        );
+        context.getSelectedMcpPage().textSnapshot = await TextSnapshot.create(
+          context.getSelectedMcpPage(),
+        );
+
+        // Simulate the main key press failing mid-sequence (e.g. a CDP
+        // hiccup) after the modifiers have already been pressed down.
+        const originalPress = page.keyboard.press.bind(page.keyboard);
```
OrKoN: could we use sinon for stubbing out methods like this?

Liohtml: Yep, done. Using sinon.stub(page.keyboard, 'press') + sinon.restore() now, same as check-for-updates.test.ts. Pushed.

Liohtml:
Rebased on main. The CI red was the McpContext getter refactor (#2342 and friends) landing after I branched — `getSelectedPptrPage()` is gone, so my new test failed to compile against the merged tree. Updated it to `getSelectedMcpPage().pptrPage`. Green locally across the full input suite, typecheck + lint clean.

OrKoN (approved):

OrKoN:
Thanks!

src/tools/input.ts
```diff
@@ -526,12 +526,20 @@ export const pressKey = definePageTool({
...
-      await page.pptrPage.keyboard.press(key);
-      for (const modifier of modifiers.toReversed()) {
-        await page.pptrPage.keyboard.up(modifier);
+      const heldModifiers: KeyInput[] = [];
+      try {
+        for (const modifier of modifiers) {
+          await page.pptrPage.keyboard.down(modifier);
+          heldModifiers.push(modifier);
+        }
+        await page.pptrPage.keyboard.press(key);
```
OrKoN, line 535: @Liohtml I have seen https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2353 and I just want to clarify something here: this code is an improvement but generally if press fails also keyboard.ups would fail because the usual cause is something like the page is closed/crashed. Do you actually have realistic scenarios when this would fail on healthy pages?

Issue #317, opened 2025-10-08
Allow switching execution context to Shared Workers


tobiu:
Hi again,

This is a follow-up to our feature request for dedicated workers (https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/316).

Our AI development agent, Gemini 2.5 Pro, which helped draft the previous request, also identified the need for similar support for Shared Workers.

In the [Neo.mjs framework](https://github.com/neomjs/neo), our entire architecture is based on the "Off the Main Thread" (OMT) concept, which you can read more about [here](https://github.com/neomjs/neo/blob/dev/learn/benefits/OffTheMainThread.md). As part of this, we use a **Shared Worker** to manage state and communication across multiple browser windows for true multi-window application support. While our cornerstone is the `app` worker, our architecture allows for multiple shared workers (e.g., for `data`, `vdom`, `canvas`, `tasks`), each with a distinct responsibility.

**The Problem:**
Currently, we have no debugging visibility into our shared workers whatsoever. Unlike dedicated workers where some errors might bubble up, the shared worker is a complete black box when using `mcp-server`. We cannot see console output, and more importantly, we cannot interact with its global scope.

**Proposed Feature:**
Could the context-switching mechanism proposed in issue #316 be extended to also support Shared Workers?

We understand this may be significantly more complex, as shared workers have a different lifecycle and connection model (e.g., connecting via a `MessagePort`). However, having this capability would be invaluable for debugging multi-window applications and would make `mcp-server` a comprehensive tool for modern, multi-threaded web development.

This would be a critical feature for us and, we believe, for anyone building complex applications that share logic across multiple tabs or windows.

As before, we are happy to provide any details about our use case or assist in testing. Thank you!

natorion:
Let's merge it in the other issue

Issue #316, opened 2025-10-08
Allow switching execution context to Dedicated Workers


tobiu:
Hi there,

First off, thank you for the `mcp-server` project. It's a fantastic tool and a great starting point for programmatic browser interaction.

We are integrating `mcp-server` into the [Neo.mjs framework](https://github.com/neomjs/neo) workflow to give our AI development agent, Gemini 2.5 Pro, the ability to 'see' and interact with the UI. While working on this integration, the agent itself identified this limitation and helped draft this feature request.

Our framework is built on an "Off the Main Thread" (OMT) architecture, where nearly all application logic runs inside workers. This creates a challenge for debugging that we believe `mcp-server` is uniquely positioned to solve.

**The Problem:**
The standard Chrome DevTools UI provides a dropdown that allows developers to switch the console's execution context between the main thread and any active workers. This is essential for debugging, as it allows for direct interaction with the global scope of each thread.

Currently, the `mcp-server` appears to be bound to the main thread's execution context. While some error messages from dedicated workers might bubble up, it's impossible to interact with the worker's scope. For example, we can't run a command like `Neo.get('my-component-id')` to inspect a live object that exists only within the worker.

**Proposed Feature:**
Could a mechanism be introduced to replicate the context-switching functionality of the real DevTools?

Ideally, this would involve:
1.  An API to list all available execution contexts (e.g., main thread, dedicated worker 1, dedicated worker 2).
2.  An API to set the active execution context for the `mcp-server` connection.

Once the context is switched to a specific worker, all subsequent `Runtime.evaluate` calls would execute within that worker's global scope. This would provide the powerful, interactive debugging experience needed for modern, multi-threaded web applications.

We believe this feature would unlock a new level of debugging and automation for any project that leverages dedicated workers.

We'd be happy to provide more details on our use case or assist in testing. Thanks for considering this!

sebastianbenz:
This is related to #96 for which this is required as well.

natorion:
I think this makes sense. Similar to the extension support, we should make sure that selecting and working on other execution context does not increase the baseline token consumption of the MCP server e.g. we should not return all exection context on default, but it should be something explicit.

tobiu:
Hi @natorion,

thanks for confirming the ticket, appreciated. I did split the topic into dedicated workers (should be available from the main thread devtools) and shared workers #317 (what we get inside `chrome://inspect/#workers`) to keep the scope reasonable.

[...]

PR #2352, opened 2026-07-13
feat(script): evaluate scripts in dedicated worker contexts

gunjanjaswal:
Adds opt-in support for evaluating JavaScript inside a page's dedicated Web Workers, so agents can reach off-main-thread application logic (e.g. Neo.mjs).

Everything is gated behind a new `--experimentalWorkers` flag and off by default. Worker contexts are never enumerated unless explicitly requested, so baseline token usage is unchanged.

- `list_dedicated_workers` lists the selected page's dedicated workers and assigns each a stable worker id.
- `evaluate_script` gains an optional `workerId` argument that resolves the target through the existing `Page | Frame | WebWorker` union, mirroring the existing `serviceWorkerId` path (same mutual-exclusivity and no-element-args guards).

Shared workers (#317) are intentionally left out of scope.

Tests cover both tools and the new evaluate path. `build`, `typecheck`, `eslint`, `prettier`, and `npm run gen` (docs + CLI reference) all run clean.

Refs #316


OrKoN:
Thanks for the PR! I think it looks good but I will try to do a review next week.

gunjanjaswal:
Thanks @OrKoN! I've just rebased it on main to clear the conflict (the tool-metrics file needed my `list_dedicated_workers` entry re-appended after the new heap-snapshot tool), so it's mergeable again and ready whenever you get to it. No rush.

tests/tools/script.test.ts
```diff
@@ -469,5 +470,150 @@ describe('script', () => {
...
+      );
+    });
   });
 });
+
+/**
+ * Spawns a dedicated worker in the page and resolves once Puppeteer has
+ * attached to it.
+ */
+async function spawnDedicatedWorker(page: Page): Promise<void> {
```
OrKoN, line 605: Could we move it to test utils?

src/tools/script.ts
```diff
@@ -146,6 +186,34 @@ Example with arguments: \`(el) => el.innerText\`
   };
 });
 
+export const listDedicatedWorkers = defineTool({
+  name: 'list_dedicated_workers',
```
OrKoN, line 190: I have been thinking if we need a dedicated tool to list workers or if we can just list the workers in the pages response similar to service workers are handled `includeExtensionServiceWorkers`? I think ideally we could land https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2434 first to streamline the interactions with workers.

gunjanjaswal:
Thanks @OrKoN. On the test helper, yes, I'll move it to test utils.

On the bigger question: listing workers in the pages response the way `includeExtensionServiceWorkers` does makes sense to me. It keeps the worker surface consistent with how service workers already appear rather than adding a separate listing path. And I agree #2434 (extracting worker logic into `McpWorker`) landing first gives this something clean to build on instead of threading worker handling through `script.ts`.

Happy to sequence it that way. Would you prefer I pick up #2434 first and then rework this on top of `McpWorker` (folding the worker listing into the pages response), or is it already spoken for? Either way I'll hold the rebase here until the shape is settled, so I'm not churning the branch against a design that's about to change.

OrKoN:
hi @gunjanjaswal, thanks, if you are still interested, let's work on https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2434 first. I will close this PR for now to reduce the amount of open PRs but let's re-open it after refactoring is done.

PR #2354, opened 2026-07-13
fix(input): prevent upload_file crash on chooser-opening elements

### LINKED ISSUE #2310: upload_file on a chooser-opening element crashes the tab with RESULT_CODE_KILLED_BAD_MESSAGE

ashutoshsinghpr7:
## Summary

Fixes #2310 — `upload_file` crashes the Chrome tab with `RESULT_CODE_KILLED_BAD_MESSAGE` when used on a chooser-opening proxy element (e.g. a button that calls `input.click()`).

## Root Cause

`upload_file` called `handle.uploadFile()` on **every** element. This method is only valid for `<input type="file">` elements. Calling it on a button or other element sends an invalid IPC message to the renderer, causing Chrome to kill the tab.

The crash happened *before* reaching the existing error-handling catch block, so the fallback `waitForFileChooser()` path was never reached.

## Fix

Check the element type before choosing the upload path:

- `<input type="file">` → use `uploadFile()` directly (fast, safe, unchanged behavior)
- Everything else → use `waitForFileChooser()` + `click()` (handles proxy elements correctly)

This matches the documented behavior:
> `uid`: The uid of the file input element **or an element that will open file chooser** on the page from the page content snapshot

## Testing

Self-contained repro from the issue:
1. `new_page` with a page containing a hidden `<input type="file">` + button that triggers `input.click()`
2. `take_snapshot` → get button uid
3. `upload_file` with button uid → no longer crashes tab, file upload succeeds via chooser

Existing file input direct-upload path is unchanged and continues to work.


OrKoN (changes requested):
Thanks for a PR. Is there a test that reproduces the crash? It sounds like the best place to fix the crash would be in Chromium instead of working around it here.

OrKoN:
Closing as stale.

PR #2358, opened 2026-07-13
perf: defer focus emulation until page selection

dexhunter:
## Summary

- Defer `emulateFocusedPage(true)` until an `McpPage` is first used instead of applying it to every discovered page during initialization.
- Cover both explicit selection and direct `--experimental-page-id-routing` targets, awaiting the first enable attempt before the page-scoped handler runs.
- Share one in-flight enable promise per page, remember successful enables, and allow a later operation to retry after a protocol failure.
- Keep previously used pages emulated and leave eager target attachment and console/network collection unchanged.

Related to #1921.

## Why

`McpPage.init()` currently enables focused-page emulation for every discovered tab. On a 35-tab profile, that sends 35 focus-emulation commands even though only one page is initially selected. The other 34 calls disable normal background-page throttling for pages the MCP has never used.

This change applies the existing behavior on first use. The initially selected page is still emulated, a page selected later is enabled once, and a page targeted directly by `pageId` is enabled before its first page-scoped operation. A previously used page is not disabled. This preserves shared-browser and multi-agent behavior without eagerly changing never-used tabs.

This is deliberately one narrow part of #1921. It does not change Puppeteer's page attachment, DevTools universe creation, collectors, or page enumeration. PR #2271 proposes a global opt-out flag; this patch instead retains emulation automatically for every used page while avoiding eager calls for unused pages.

## Measurement

The strict evaluator launches real Chrome, connects the built production MCP over CDP, and attributes focus-emulation protocol calls to selected and non-selected targets.

| Revision | Initialization calls, 35 tabs | Calls for non-selected tabs |
| --- | ---: | ---: |
| Current `main` | 35 | 34 |
| This PR | 1 | 0 |

With `--experimental-page-id-routing`, the evaluator also targets an unselected page directly: the first page-scoped call produces exactly one focus-emulation command, and a repeated call produces zero additional commands.

The selection-gated direction was identified while running autoresearch with Weco. The 14-step trajectory, baseline, and evaluated candidates are available in this [public autoresearch run](https://dashboard.weco.ai/share/VX3kq2eqGPwHmq6I4G_srwdTqeSIfZ).

## Validation

- `npm run build`
- `node scripts/test.mjs tests/ToolHandler.test.ts tests/McpContext.test.ts`
- `npm run check-format`
- `npm run typecheck`
- strict real-Chrome evaluator: `focused_page_emulation_nonselected_calls: 34 -> 0`
- strict page-ID gates: first direct use `1` call, repeated use `0` calls
- `git diff --check`

## Page-ID routing compatibility

This remains compatible with the experimental page-ID routing option: a page-scoped request resolves the requested pageId and awaits enableFocusEmulation before the handler runs. select_page follows the same path. The first request for a page sends one enable command and later requests send none. PR #2271's opt-out flag is a separate product choice; this PR keeps emulation for every page that is actually used while avoiding calls for unused pages.

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -397,6 +397,7 @@ export class McpContext implements Context {
 
   selectPage(newPage: McpPage): void {
     this.#selectedPage = newPage;
+    void newPage.enableFocusEmulation();
     newPage.updateTimeouts();
   }
 
--- a/src/McpPage.ts
+++ b/src/McpPage.ts
@@ -73,6 +73,8 @@ export class McpPage implements ContextPage {
   // Metadata
   isolatedContextName?: string;
   #devtoolsUniverse?: TargetUniverse;
+  #focusEmulationEnabled = false;
+  #focusEmulationPromise?: Promise<void>;
 
   // Dialog
   #dialog?: Dialog;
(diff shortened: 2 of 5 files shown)


OrKoN (changes requested):
This is going to affect clients using `--experimental-pageid-routing` that we aim to make the default soon. This allows targeting pages by ID in most of the commands requiring no use of select_page. I think it might better to support https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2271 instead. WDYT?

dexhunter:
Thanks for calling out the interaction with page-ID routing. I agree that #2271 is the right direction for making focused-page emulation configurable; this PR instead changes the default initialization timing, so rebasing it blindly would compete with that API change.

I will treat #2358 as blocked/superseded by the #2271 direction rather than push an incompatible follow-up. The current branch is also behind `main`, so I will not claim the existing performance result as merge-ready.

dexhunter:
Closing this in favor of #2271, which is the maintainer-requested direction for configurable focused-page emulation. Rebasing this default-timing change would still conflict with the planned page-ID routing behavior. Thanks for the review.

PR #2367, opened 2026-07-14
feat: include the CDP targetId in list_pages output

### LINKED ISSUE #2366: Include the CDP `targetId` in `list_pages` output

I run a thin stdio proxy in front of `chrome-devtools-mcp` that keeps page ids stable across restarts of the browser it drives (a dev app that gets re-exec'd at the same CDP port). The server reconnects fine on 1.6.0 and #2345 keeps ids unique, but a tab still gets a new id after a restart, so the proxy rewrites `list_pages` to hold one id per tab.

To do that it has to map each `list_pages` entry back to its CDP target, and there's no shared id to join on – `list_pages` gives `title` and `url` but not the CDP `targetId`. So the proxy fetches `GET /json` on the side and matches each entry by title, then url (consume-once). It's the most fragile thing the proxy does. Tabs that share a url are separable only by title – several of my tabs render the identical `data:` loader – so when titles also collide the match comes down to positional luck. And it's a second, independent read of browser state, so a tab opening or navigating between the `/json` fetch and the server's own view throws the listing off.

Put each page's CDP `targetId` in `list_pages` output – ideally a field on the `structuredContent.pages[]` entries (and/or the text line). Then the mapping is just `entry.targetId`, with no `/json` round-trip to match up.

It's already on hand server-side (`PageCollector` reads it per page today), so it's a one-field addition to `createStructuredPage()` that leaves the existing fields alone – same shape as #2156, which added `title`. Happy to send the PR.

- Keeping the `GET /json` + title/url heuristic – works today, but is the fragile part above
- Reading the id client-side some other way – nothing in the MCP output keys on the CDP target, so it still means re-reading `/json` (or opening a separate CDP session) and re-matching, with the same fragility and a second source of truth to keep in sync

### Additional context

`targetId` is stable per tab within a session and renews when the browser process restarts, same lifecycle as the underlying CDP id – I'm only after exposing it, not changing that.

Related: #2156 (added `title`), #2345 (unique ids across reconnects), #2304 (stale-listing selection retarget).

thomasbachem:
Closes #2366

Adds each page's CDP `targetId` to `list_pages` output, on the `structuredContent.pages[]` entries. `createStructuredPage()` reads it via the same internal-Puppeteer accessor `PageCollector` already uses (`page.target()._targetId`), so it's additive – `id`/`url`/`title`/`selected` are untouched.

The use case is in #2366 – correlating an MCP page to its CDP target without the fragile `GET /json` + title/url matching a client has to do today. With the id present it's a direct `entry.targetId` lookup.

The test lists pages and checks each `targetId` is distinct, non-empty, and among the real page targets from `browser.targets()` – a separate path than the code under test, so it fails if the value is wrong, not just if the field went missing. `stabilizeStructuredContent` normalizes the non-deterministic id so snapshots stay stable, and the affected snapshots are updated.

### DIFF
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -1374,12 +1374,17 @@ function createStructuredPage(
     url: string;
     title: string;
     selected: boolean;
+    targetId: string;
     isolatedContext?: string;
   } = {
     id: mcpPage.id,
     url: mcpPage.pptrPage.url(),
     title,
     selected: context.isPageSelected(mcpPage),
+    // CDP target id, stable per tab within a browser session, so clients can
+    // correlate an MCP page to its underlying CDP target.
+    // @ts-expect-error use internal Puppeteer API to get target ID
+    targetId: mcpPage.pptrPage.target()._targetId,
   };
   if (isolatedContextName) {
     entry.isolatedContext = isolatedContextName;
--- a/tests/McpResponse.test.js.snapshot
+++ b/tests/McpResponse.test.js.snapshot
@@ -325,7 +325,8 @@ exports[`McpResponse > list pages 2`] = `
       "id": 1,
       "url": "about:blank",
       "title": "",
-      "selected": true
+      "selected": true,
+      "targetId": "<targetId>"
     }
   ]
 }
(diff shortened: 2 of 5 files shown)


thomasbachem:
Closing this – exposing `targetId` on `list_pages` was declined in #2366, and I'll stick with `GET /json` matching on our side. Thanks for the look!

PR #2205, opened 2026-06-13
fix: limit windows test concurrency

pupuking723:
Fixes #2137.

This updates the test workflow so Windows jobs run the Node test runner with `--test-concurrency=1`.

The issue points to intermittent Windows failures where the browser target closes during screenshot tests. Those tests start real browser sessions, and limiting test concurrency on Windows avoids overlapping screenshot/browser work there while keeping the existing parallel behavior on Linux and macOS.

Verification:

- `git diff --check`
- Inspected the workflow command expansion: only `windows-latest` receives `--test-concurrency=1`; merge queue retries still pass `--retry` as before.

Issue #2137, opened 2026-05-26
[Task]: Flaky tests on windows


OrKoN:
### Task to do:

Windows test often fail with something like:

```
✖ failing tests:

test at build\tests\tools\screenshot.test.js:72:9
✖ with full page resulting in a large screenshot (3124.219ms)
  TargetCloseError: Protocol error (Page.captureScreenshot): Target closed
      at CallbackRegistry.clear (file:///D:/a/chrome-devtools-mcp/chrome-devtools-mcp/node_modules/puppeteer-core/lib/puppeteer/common/CallbackRegistry.js:82:36)
      at CdpCDPSession.onClosed (file:///D:/a/chrome-devtools-mcp/chrome-devtools-mcp/node_modules/puppeteer-core/lib/puppeteer/cdp/CdpSession.js:121:25)
      at #onClose (file:///D:/a/chrome-devtools-mcp/chrome-devtools-mcp/node_modules/puppeteer-core/lib/puppeteer/cdp/Connection.js:210:21)
      at Socket.<anonymous> (file:///D:/a/chrome-devtools-mcp/chrome-devtools-mcp/node_modules/puppeteer-core/lib/puppeteer/node/PipeTransport.js:26:30)
      at Socket.emit (node:events:521:24)
      at Pipe.<anonymous> (node:net:350:12) {
    [cause]: ProtocolError
        at <instance_members_initializer> (file:///D:/a/chrome-devtools-mcp/chrome-devtools-mcp/node_modules/puppeteer-core/lib/puppeteer/common/CallbackRegistry.js:102:14)
        at new Callback (file:///D:/a/chrome-devtools-mcp/chrome-devtools-mcp/node_modules/puppeteer-core/lib/puppeteer/common/CallbackRegistry.js:106:16)
        at CallbackRegistry.create (file:///D:/a/chrome-devtools-mcp/chrome-devtools-mcp/node_modules/puppeteer-core/lib/puppeteer/common/CallbackRegistry.js:25:26)
        at Connection._rawSend (file:///D:/a/chrome-devtools-mcp/chrome-devtools-mcp/node_modules/puppeteer-core/lib/puppeteer/cdp/Connection.js:118:26)
        at CdpCDPSession.send (file:///D:/a/chrome-devtools-mcp/chrome-devtools-mcp/node_modules/puppeteer-core/lib/puppeteer/cdp/CdpSession.js:72:14)
[...]

PR #2368, opened 2026-07-15
test: relaunch a disconnected cached browser in tests

thomasbachem:
The Windows screenshot flakes in #2137 are `TargetCloseError`s (`Target closed`), which is what you get from reusing a browser that already died.

`withBrowser` (`tests/utils.ts`) caches one browser per `launchOptions` key, but the cache-hit path only checks that an entry exists, not that it's still connected – and nothing evicts a dead one. So a single browser death mid-run (a pipe drop, an OOM, a teardown race) leaves every later same-key test reusing the dead handle, turning one failure into a run of them.

#2205's concurrency limit lowers how often a browser dies; a liveness check on cache hit bounds the blast radius when one still does – a disconnected browser is dropped and relaunched, so a death fails only its own test. Inert on a healthy run, since `connected` only goes false once the connection is actually gone.

The test closes a cached browser and asserts the next `withBrowser` call gets a fresh, connected one – it fails on `main` and passes with the check.

### DIFF
--- a/tests/utils.test.ts
+++ b/tests/utils.test.ts
@@ -1,0 +1,34 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'node:assert';
+import {describe, it} from 'node:test';
+
+import type {Browser} from 'puppeteer';
+
+import {withBrowser} from './utils.js';
+
+describe('withBrowser', () => {
+  it('relaunches a cached browser that has disconnected', async () => {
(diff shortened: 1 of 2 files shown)


OrKoN (approved):
Thanks!

thomasbachem:
#2396 covers this – the retry loop drops the cached browser and relaunches when it disconnects, so the original failure is gone.

The one thing it doesn't do is check `browser.connected` before using a cached browser, so a browser that's already dead still costs an attempt out of the three. Happy to send that as a small follow-up if you think it's worth it, but closing this one as it stands.

PR #2371, opened 2026-07-15
feat: adding network request overrides

### LINKED ISSUE #848: Network Request Interception and URL Replacement

Currently there is no automated way through the Chrome DevTools MCP server to intercept and modify network requests. Developers can manually override or redirect resources using the Chrome DevTools UI (e.g., Local Overrides or Response Overrides), but these actions must be done interactively and cannot be scripted or integrated into automated measurement workflows. This becomes a blocker when trying to test development versions of scripts against real production pages.
This become a blocker also to automate experiments in CI or lab environments by using built scripts before doing a production release.
In this issue I'm referring to script but it could be useful also for others resources as is done in the Chrome DevTools UI.

Example real-world impact: if a production page loads a script at https://bar.foo.com/scriptWeAlreadyHaveInWebsite.js, there is no way via MCP today to replace that with a development version and collect metrics without doing it manually in the browser.

Add support in the Chrome DevTools MCP server for programmatic network request interception and modification, including:

- Redirect requests to a different URL
- Override responses with custom content (remote or local file)

Specifically, the ability to:

- Match a network request (by URL or pattern)
- Redirect it to a different URL (e.g., dev/newVersion.js)
- Optionally serve a custom or local file as the response

This would expose the same functionality found in the DevTools UI (“Override content” / “Local Overrides”) through MCP, enabling automation and tooling scenarios.

The only alternative currently available is to manually use Chrome DevTools UI (https://developer.chrome.com/docs/devtools/overrides) or by developing a custom Chrome Extension.

### Additional context

_No response_

marco-prontera:
closes #848 

Adds page-scoped network request overrides using Chrome DevTools Protocol request interception.

This introduces three new tools:
  - add_network_override — redirects matching requests to another URL or fulfills them using a local file.
  - list_network_overrides — lists active overrides for the selected page.
  - remove_network_override — removes an override by ID.

Overrides support CDP-style URL patterns, optional resource-type filtering, persistence across navigation/reloads, local-file rereading after rebuilds, URL allow/block policies, and safe cross-origin header handling.

Essentially, imagine being a developer working on a third-party script that runs across different websites. You want to enable the agent to test what will happen in a simulated end-user environment using your local changes before shipping them to production. This feature enables you to do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,8 +15,9 @@ A [CLI](docs/cli.md) is also provided for use without MCP.
 - **Get performance insights**: Uses [Chrome
   DevTools](https://github.com/ChromeDevTools/devtools-frontend) to record
   traces and extract actionable performance insights.
-- **Advanced browser debugging**: Analyze network requests, take screenshots and
-  check browser console messages (with source-mapped stack traces).
+- **Advanced browser debugging**: Analyze and override network requests, take
+  screenshots and check browser console messages (with source-mapped stack
+  traces).
 - **Reliable automation**. Uses
   [puppeteer](https://github.com/puppeteer/puppeteer) to automate actions in
   Chrome and automatically wait for action results.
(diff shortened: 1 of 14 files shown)


OrKoN:
Thanks for the PR but we are still collecting feedback on the feature request so we will not be reviewing the implementation right now.

PR #2384, opened 2026-07-17
feat: update memory leak debugging skill to use MCP tools instead of memlab scripts

### LINKED ISSUE #2329: Update memory leak debugging skill to use own tools
Currently, the skill is using memlab scripts but we can replace that with our MCP tools https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/skills/memory-leak-debugging/SKILL.md

cc @dinfuehr

syedaly:
## Summary
Updates the memory leak debugging skill to use chrome-devtools-mcp's own tools instead of relying on memlab scripts, so the workflow is self-contained and doesn't require an external dependency.

## Motivation
The skill currently documents a memlab-based workflow for memory leak debugging. Since chrome-devtools-mcp already exposes tools that can cover this functionality, the skill should be updated to use those instead, simplifying setup and keeping the debugging flow consistent with the rest of the MCP tool suite.

Fixes #2329

## Changes
- Replaced memlab script references in `skills/memory-leak-debugging/SKILL.md` with equivalent chrome-devtools-mcp tool usage
- Updated step-by-step instructions/examples to reflect the new tool calls
- (add/remove as applicable based on your actual diff)

## Testing
- [ ] Verified the updated skill steps work end-to-end for a sample memory leak scenario
- [ ] Confirmed documentation renders correctly and links/tool names are accurate


Lightning00Blade:
PR has unrelated changes.

PR #2389, opened 2026-07-19
feat: persist CLI startup defaults

### LINKED ISSUE #2388: Persist CLI defaults for automatic daemon startup

The experimental `chrome-devtools` CLI automatically starts its background daemon when the first tool command is invoked. However, the automatic startup path currently calls `start([], sessionId)`, so it cannot reuse user-specific startup options.

For example, a user may always need:

```sh
chrome-devtools start --executablePath /usr/bin/chromium --headless=false
```

If the daemon is not running after a reboot, crash, or explicit stop, invoking a normal command such as `chrome-devtools navigate_page --url https://example.com` starts it with the built-in defaults instead. The user must remember to run `start` manually before every new daemon lifecycle, which defeats part of the automatic-start behavior.

The standalone CLI currently has no persistent configuration file or environment-based defaults for options such as `executablePath`, `headless`, `userDataDir`, or other supported `start` options.

### Describe the solution you would like

Add persistent startup defaults for the standalone `chrome-devtools` CLI.

One possible design is an XDG-compatible configuration file:

```text
$XDG_CONFIG_HOME/chrome-devtools/config.json
```

with `~/.config/chrome-devtools/config.json` as the fallback location. For example:

```json
{
  "executablePath": "/usr/bin/chromium",
  "headless": false
}
```

When a tool command automatically starts the daemon, these configured values should be passed to the existing startup path instead of using only an empty argument list.

Suggested precedence:

```text
explicit command-line options > configuration file > built-in defaults
```

The same defaults could also apply to an explicit `chrome-devtools start` command when an option is omitted. Existing option parsing and validation should be reused, and `chrome-devtools status` should continue to show the effective daemon arguments.

An equivalent persistent mechanism would also solve the problem if a configuration file is not considered appropriate.

### Describe alternatives you have considered

- Run `chrome-devtools start ...` manually before every daemon lifecycle. This is easy to forget and bypasses automatic startup
- Define a shell wrapper or function that checks `status` and conditionally runs `start`. This is shell-specific, duplicates daemon management logic, and is harder for agents and scripts to use consistently
- Run a user-level service that always keeps the daemon alive. This is platform-specific and adds lifecycle management outside the CLI
- Configure startup arguments in an MCP client. This configures an MCP server launched by that client, not the standalone `chrome-devtools` CLI daemon

### Additional context

The current automatic-start implementation is here:

https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/src/bin/chrome-devtools.ts#L262-L267

Related but distinct issues include:

- #1232, which concerns passing options through a Claude Code plugin configuration
- #1184, which added `--autoConnect` support to the CLI
- #889, which concerns discovering Chrome installed in non-standard locations

This request is specifically about persisting defaults for the standalone CLI automatic daemon startup path.

hebo6:
## Summary

- load persistent CLI startup defaults from an XDG-compatible configuration file
- apply configured options to automatic daemon startup and explicit `start` commands
- preserve command-line precedence over configuration values
- document the configuration path and format

## Testing

- `npm run format`
- `npm run build`
- `npm run test tests/cli-config.test.ts`
- `npm run test tests/e2e/chrome-devtools-start-stop.test.ts`
- `npm run test tests/e2e/chrome-devtools-commands.test.ts`
- full test suite with local Chromium, with one unrelated request-header ordering snapshot mismatch

Fixes #2388

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -33,6 +33,21 @@ chrome-devtools take_screenshot --filePath screenshot.png
 chrome-devtools stop
 ```
 
+## Startup defaults
+
+You can persist options used to start the daemon in an XDG-compatible configuration file at `$XDG_CONFIG_HOME/chrome-devtools/config.json`, or `~/.config/chrome-devtools/config.json` when `XDG_CONFIG_HOME` is not set
+
+The file contains a JSON object whose keys match the options accepted by `chrome-devtools start`
+
+```json
+{
+  "executablePath": "/usr/bin/chromium",
+  "headless": false
+}
+```
+
+These defaults apply both when a tool command automatically starts the daemon and when `chrome-devtools start` is invoked explicitly. Command-line options override configuration values, which override the built-in defaults
+
 ## Command Usage
 
 The CLI only supports tools available in the MCP server without additional arguments (see [Tool reference](./tool-reference.md)).
--- a/src/bin/chrome-devtools-cli-config.ts
+++ b/src/bin/chrome-devtools-cli-config.ts
@@ -1,0 +1,61 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import process from 'node:process';
+
+export function getCliConfigPath(
+  env: NodeJS.ProcessEnv = process.env,
+  homeDirectory = os.homedir(),
+): string {
+  const configHome =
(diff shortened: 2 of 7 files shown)


nattallius:
Hi! Please, sign the cla before we can proceed

hebo6:
Hi @nattallius, I have signed the CLA and resolved the merge conflicts with the latest main branch. The PR is now mergeable. Could you please review it when you have a chance? Thank you!

OrKoN:
Thanks for the PR but we are not going to land it for now since we are working on a general configuration solution for the MCP server + CLI. Let's revisit after that? I will close the PR for now.

PR #2391, opened 2026-07-20
refactor: split response formatting out of McpResponse

OrKoN:
This is a large change but I want to separate the data gathering from formatting before the next iteration. The change should be as-is and the tests should be refactored according to the layering described in AGENTS.md.


OrKoN:
@Dearit please stop approving PRs (this causes unnecessary notifications)

PR #2393, opened 2026-07-21
chore(deps-dev): update eslint to v10

### LINKED ISSUE #955: Update eslint major version to v10

New major version 10 of eslint was released. But currently it is not yet supported by **typescript-eslint**. [Support](https://github.com/typescript-eslint/typescript-eslint/pull/12057) for this version was added on Feb 13, but it's not yet released.

Bump @typescript-eslint/eslint-plugin when the release is ready.

This issue is to track the dependency update once the release is available.

### Additional context

_No response_

mturac:
Fixes #955

## Summary

- update ESLint and `@eslint/js` to v10
- replace the import lint plugin with an ESLint 10-compatible implementation while preserving the existing import and Node protocol checks
- update the local license rule for ESLint 10's rule-context properties
- align the supported Node.js versions with ESLint 10's engine requirements
- preserve the current recommended-rule baseline during the major upgrade

## Testing

- `npm ci`
- `npm run check-format`
- `npm run typecheck`
- `npm run test`

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     }
   },
   "engines": {
-    "node": "^20.19.0 || ^22.13.0 || >=24"
+    "node": "^20.19.0 || ^22.12.0 || >=23"
   },
   "overrides": {
     "eslint-plugin-import": {
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
   "mcpName": "io.github.ChromeDevTools/chrome-devtools-mcp",
   "devDependencies": {
     "@blackwell-systems/gcf": "^2.2.2",
-    "@eslint/js": "^9.35.0",
+    "@eslint/js": "^10.0.1",
     "@google/genai": "^2.0.1",
     "@modelcontextprotocol/sdk": "1.29.0",
     "@rollup/plugin-commonjs": "^29.0.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -99,7 +100,7 @@
     }
   },
   "engines": {
-    "node": "^20.19.0 || ^22.12.0 || >=23"
+    "node": "^20.19.0 || ^22.13.0 || >=24"
```
Lightning00Blade: Revert this change. The field say what our package supports, local development can require a different ones.

Lightning00Blade:
Let's not change the deps for now, we can simply add the following code to the `package.json`:
```
  "overrides": {
    "eslint-plugin-import": {
      "eslint": "$eslint"
    }
  },
 ```
Which will force use EsLint v10 which is only incompatible because of `eslint-plugin-import` does not declare it an acceptable peer dependency, but code wise it works.

Lightning00Blade (changes requested):

mturac:
Updated this to keep the existing eslint-plugin-import setup, add the ESLint peer override, and restore the package engine range.

Validation:
- npm run check-format
- npm run build
- npm run test -- --test-concurrency=1

OrKoN:
Thanks for the PR! @Lightning00Blade could you please re-gen package-lock locally for this fix?

mturac:
Rebased on latest main and regenerated the lockfile locally with `npm install --package-lock-only`; it produced no tracked changes.

Validation:
- `npm run check-format`
- `npm run build`
- `npm run test -- --test-concurrency=1`

package.json
Lightning00Blade: This showed up again.

mturac: Restored the top-level package engine range and lockfile mirror to `^20.19.0 || ^22.12.0 || >=23`, while keeping the `eslint-plugin-import` override.

Validation:
- `npm run check-format`
- `npm run build`
- `npm run test -- --test-concurrency=1`

Lightning00Blade (approved):

PR #2400, opened 2026-07-22
feat: add --visual-cursor to show an animated cursor during input automation

ccl125:
## Motivation

When an agent drives a real (headed) Chrome instance, all input is injected at the protocol level, so the OS cursor never moves and a human watching the browser cannot tell where the agent is clicking. Other agent harnesses solve this with a visible cursor overlay; chrome-devtools-mcp currently has nothing similar. This makes supervised/demo sessions hard to follow.

## What this PR does

Adds an opt-in `--visual-cursor` flag. When enabled, mouse-based input tools (`click`, `dblClick`, `click_at`, `drag`) animate a ghost cursor on the page before acting:

- The cursor smoothly slides from its current position to the target point (default 800 ms, configurable via `--visual-cursor-duration <ms>`).
- An expanding ripple marks the click point.
- The cursor is injected with `Page.addScriptToEvaluateOnNewDocument`, so it survives navigations.
- Everything is wrapped in silent fallbacks: if injection or animation fails (CSP, navigation in flight, closed page), the real input action is unaffected.
- Flag off (default): zero behavioral or latency change.

## Implementation

- `src/visualCursor.ts`: injection script builder (parameterized by duration), idempotent `ensureVisualCursor()`, and `animateCursorTo()`.
- `src/tools/input.ts`: `click`/`clickAt`/`drag` resolve the target point (`clickablePoint()` with a `boundingBox()` center fallback) and call `animateCursorTo()` before the real action when the flag is on. Keyboard-only tools (`fill`, `press_key`, `type_text`) are untouched, as is the native `<select>` option path.
- Docs regenerated with `npm run gen`.

## Testing

- New `tests/visualCursor.test.ts` (12+ tests): flag parsing, injection script content (default and custom duration), idempotent ensure, re-injection after navigation, move + ripple behavior, silent degradation on closed pages, end-to-end click behavior with the flag on/off.
- `npm run build`, `npm run typecheck`, `npm run format` and `npm test` pass (one pre-existing environment-dependent failure in `tests/tools/screenshot.test.ts` "Page is too large" also fails on a clean `main`).
- Manually verified on headed Chrome 150 (`--autoConnect --visual-cursor`): cursor glides to targets across clicks and persists through navigations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -614,6 +614,16 @@ The Chrome DevTools MCP server supports the following configuration option:
   - **Type:** string
   - **Default:** `false`
 
+- **`--visualCursor`/ `--visual-cursor`**
+  If specified, shows an animated ghost cursor on the page before mouse actions (click, double click, click at coordinates, drag) so a human observer can follow where the agent is interacting. Defaults to false.
+  - **Type:** boolean
+  - **Default:** `false`
+
+- **`--visualCursorDuration`/ `--visual-cursor-duration`**
+  Duration in milliseconds of the ghost cursor slide animation shown with --visual-cursor. Defaults to 800.
+  - **Type:** number
+  - **Default:** `800`
+
 - **`--proxyServer`/ `--proxy-server`**
   Proxy server configuration for Chrome passed as --proxy-server when launching the browser. See https://www.chromium.org/developers/design-documents/network-settings/ for details.
   - **Type:** string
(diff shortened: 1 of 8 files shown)


ccl125:
@googlebot I signed it!

OrKoN:
Please file a feature request for initial discussions. For now we do not plan to inject scripts into inspected pages to support something like a visual cursor.

OrKoN:
Thanks for filing the feature request!

PR #2402, opened 2026-07-22
fix: rename maxRetainedSize column

dinfuehr:
maxRetainedSize confused the LLM. We actually report the retained size of all objects in that aggregate, not the maximum value.


Lightning00Blade (approved):
Thanks!

PR #2403, opened 2026-07-22
fix: preserve browserUrl path prefix

### LINKED ISSUE #2394: --browserUrl drops path prefix when resolving /json/version
## Describe the bug

When `--browserUrl` includes a path prefix (e.g. Chrome DevTools behind a reverse proxy or gateway), Puppeteer strips the path prefix when constructing the `/json/version` discovery URL.

## Steps to reproduce

1. Set up CDP behind a reverse proxy with path prefix: `http://127.0.0.1:54704/t/abc123`
2. Run: `npx chrome-devtools-mcp --browserUrl http://127.0.0.1:54704/t/abc123`
3. Error: `Failed to fetch browser webSocket URL from http://127.0.0.1:54704/json/version` (path prefix `/t/abc123` dropped)

## Root cause

In bundled Puppeteer's `getWSEndpoint()`:

```js
const endpointURL = new URL('/json/version', browserURL);
```

The JS `URL` constructor with an absolute path replaces the base URL's entire pathname. `http://127.0.0.1:54704/t/abc123` + `/json/version` → `http://127.0.0.1:54704/json/version` (path prefix lost).

## Suggested fix

```js
const endpointURL = new URL(browserURL);
endpointURL.pathname = endpointURL.pathname.replace(/\/?$/, '/') + 'json/version';
```

## Workaround

Use `--wsEndpoint` with the full WebSocket URL to bypass HTTP discovery.

## Environment

- chrome-devtools-mcp 0.16.0 (same code path in latest)
- macOS, INA app CDP broker with token-based path prefix
BODY 2>&1

mturac:
## Summary

- Resolve `--browserUrl` through its `/json/version` discovery endpoint before connecting
- Preserve path prefixes when constructing the discovery URL and strip query strings/fragments
- Validate discovery payloads before using the WebSocket endpoint

## Tests

- `npm run test tests/browser.test.ts`
- `npm run check-format`
- `npm run test`
- `git diff --check`

Fixes #2394

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -44,6 +44,40 @@ function makeTargetFilter(enableExtensions = false) {
   };
 }
 
+function browserJsonVersionEndpoint(browserURL: string): string {
+  const endpointURL = new URL(browserURL);
+  const pathname = endpointURL.pathname.endsWith('/')
+    ? endpointURL.pathname
+    : `${endpointURL.pathname}/`;
+  endpointURL.pathname = `${pathname}json/version`;
+  endpointURL.search = '';
+  endpointURL.hash = '';
+  return endpointURL.toString();
+}
+
+function isObject(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
+
+export async function resolveBrowserWSEndpoint(
+  browserURL: string,
+): Promise<string> {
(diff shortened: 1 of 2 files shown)


OrKoN:
Thanks for the PR but I think this need to be addressed in Puppeteer instead.

PR #2407, opened 2026-07-23
feat(skills): add modern-web-guidance skill

mathiasbynens:
b/538102133


OrKoN:
@mathiasbynens are we moving forward with this or can we close the PR?

mathiasbynens:
We’re not moving forward with this as-is.

PR #2408, opened 2026-07-23
refactor: migrate the MCP SDK to v2

Lightning00Blade:
This was generated based on the documents provided in the SDK v2 migration for both the SDK and the MCP version.
It is to serve as a template for pre-factoring and other changes.

### DIFF
--- a/scripts/generate-docs.ts
+++ b/scripts/generate-docs.ts
@@ -6,7 +6,7 @@
 
 import fs from 'node:fs';
 
-import type {Tool} from '@modelcontextprotocol/sdk/types.js';
+import type {Tool} from '@modelcontextprotocol/server';
 
 import {
   mcpOptions,
@@ -38,13 +38,15 @@ interface ZodCheck {
 }
 
 interface ZodDef {
-  typeName: string;
+  typeName?: string;
   checks?: ZodCheck[];
   values?: string[];
-  type?: ZodSchema;
+  entries?: Record<string, string>;
+  type?: string | ZodSchema;
   innerType?: ZodSchema;
   schema?: ZodSchema;
-  defaultValue?: () => unknown;
+  in?: ZodSchema;
+  defaultValue?: (() => unknown) | unknown;
 }
 
 interface ZodSchema {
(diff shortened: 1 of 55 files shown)


src/bin/chrome-devtools-mcp-main.ts
```diff
@@ -55,29 +55,24 @@ async function shutdown(reason: string): Promise<void> {
...
 logger?.(`Starting Chrome DevTools MCP Server v${VERSION}`);
 const {server} = await createMcpServer(args, {
   logFile,
 });
-const transport = new StdioServerTransport();
-await server.connect(transport);
-logger?.('Chrome DevTools MCP Server connected');
 logDisclaimers(args);
 void ClearcutLogger.get()?.logDailyActiveIfNeeded();
 void ClearcutLogger.get()?.logServerStart(computeFlagUsage(args, cliOptions));
```
OrKoN: we should check if this should be logged after the server is started

src/telemetry/transformation.ts
```diff
@@ -162,10 +199,10 @@ function hasEquivalentType(zodType: ZodType, value: unknown): boolean {
 }
 
 export function sanitizeParams(
-  params: ShapeOutput<zod.ZodRawShape>,
-  schema: zod.ZodRawShape,
-): ShapeOutput<zod.ZodRawShape> {
-  const transformed: ShapeOutput<zod.ZodRawShape> = {};
+  params: Record<string, unknown>,
+  schema: Record<string, unknown>,
```
OrKoN, line 208: can we keep more specific types here?

src/tools/console.ts
```diff
@@ -39,6 +39,44 @@ const FILTERABLE_MESSAGE_TYPES: [
 
 const LIST_CONSOLE_MESSAGES_TOOL_NAME = 'list_console_messages';
 
+const consoleSchema = {
```
OrKoN: I think we should keep this inline and figure out inference of types

upgrade-to-v2.md
```diff
@@ -0,0 +1,1880 @@
+---
```
OrKoN, line 1: TODO: remove tempoary markdown docs

tests/third_party_notices.test.js.snapshot
```diff
@@ -430,116 +789,212 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
...
+Documentation in this project (excluding specifications) is licensed under
+CC-BY-4.0. See https://creativecommons.org/licenses/by/4.0/legalcode for
+the full license text.
 
 
 -------------------- DEPENDENCY DIVIDER --------------------
 
-Name: fast-deep-equal
-URL: https://github.com/epoberezkin/fast-deep-equal#readme
+Name: @modelcontextprotocol/client
```
OrKoN, line 803: do we bundle the client? why do we need to bundle it?

OrKoN (commented):
Have you tested manually with some clients like Antigravity?

scripts/generate-docs.ts
```diff
@@ -28,7 +28,12 @@ const OUTPUT_PATH = './docs/tool-reference.md';
 const SLIM_OUTPUT_PATH = './docs/slim-tool-reference.md';
 
 // Extend the MCP Tool type to include our annotations
-interface ToolWithAnnotations extends Tool {
+interface ToolWithAnnotations extends Omit<Tool, 'inputSchema'> {
+  inputSchema: {
```
OrKoN, line 32: why do we need to omit and re-add inputSchema? is there anything non-compatible?

tests/third_party_notices.test.js.snapshot
Lightning00Blade: We use it in `src/daemon/daemon.ts`.

scripts/generate-docs.ts
Lightning00Blade: Internal type and our type conflict so we need to override them.

OrKoN (approved):

PR #2420, opened 2026-07-27
fix(performance): reset trace-running flag when start_trace setup fails

### LINKED ISSUE #2419: performance_start_trace permanently reports 'a performance trace is already running' after a setup step throws
### Description of the bug

The `performance_start_trace` handler sets the context "trace is running" flag **before** several fallible `await`s, but that flag is only ever cleared inside `stopTracingAndAppendOutput`'s `finally`. If any `await` between setting the flag and reaching that function rejects, the flag stays `true` for the remainder of the session, and every subsequent `performance_start_trace` short-circuits with:

> Error: a performance trace is already running. Use performance_stop_trace to stop it. Only one trace can be running at any given time.

Both `reload` and `autoStop` default to `true`, so with default parameters the very first thing the handler does after setting the flag is a navigation (`goto('about:blank')`) that can reject.

**Affected code** (`src/tools/performance.ts`, `startTrace` handler at `main` @ e2c19fe):

- `context.setIsRunningPerformanceTrace(true)` is called at the top of the handler.
- After it, these can all throw before any reset is reached:
  - `page.pptrPage.goto('about:blank', ...)` (reload path)
  - `page.pptrPage.tracing.start({categories})`
  - `page.pptrPage.goto(pageUrlForTracing, ...)` (reload path)
- The only reset is `context.setIsRunningPerformanceTrace(false)` inside `stopTracingAndAppendOutput`'s `finally`, which is never reached on a setup failure.

Recovery is non-obvious and noisy: calling `performance_stop_trace` passes its guard, then Puppeteer's `tracing.stop()` throws "Cannot stop recording: tracing was not started". Its `finally` does eventually clear the flag, but the recovery call surfaces an error too — an agent driving the tool typically just sees "already running" and abandons tracing for the session.

Note: this is distinct from #808. There, a trace is genuinely running and the request is to auto-stop it. Here, **no trace is running at all** — the flag is stale because an exception during setup skipped the reset. It is also a counterexample to "just use the stop trace tool" as a recovery path, since that path errors as well.

### Reproduction

1. Get the selected page into a state where the initial navigation fails (e.g. the target is closing/crashed, or `about:blank` navigation rejects), or otherwise cause `goto('about:blank')` / `tracing.start()` to throw.
2. Call `performance_start_trace` with default params (`reload: true`, `autoStop: true`). The call throws.
3. Call `performance_start_trace` again on a healthy page → "a performance trace is already running", although none is.

Deterministic reproduction in the repo's own test harness (fails on current `main`):

```ts
it('resets the running flag if a setup step throws', async () => {
  await withMcpContext(async (response, context) => {
    const selectedPage = context.getSelectedMcpPage().pptrPage;
    sinon.stub(selectedPage, 'url').callsFake(() => 'https://www.test.com');
    sinon.stub(selectedPage, 'goto').rejects(new Error('Navigation failed'));
    const startTracingStub = sinon.stub(selectedPage.tracing, 'start');
    await assert.rejects(
      startTrace.handler(
        {params: {reload: true, autoStop: true}, page: context.getSelectedMcpPage()},
        response,
        context,
      ),
      /Navigation failed/,
    );
    sinon.assert.notCalled(startTracingStub);
    // Fails on main: the flag is stuck true.
    assert.strictEqual(context.isRunningPerformanceTrace(), false);
  });
});
```

### Expectation

After a failed start, the tool returns to a clean state and a subsequent `performance_start_trace` starts a new trace normally.

Proposed fix: wrap the start body in `try/catch`; on failure, stop tracing if it was started (ignoring "not started" errors) and clear the flag before rethrowing. When `autoStop` already ran `stopTracingAndAppendOutput`, the flag is already false and the catch is a no-op. I have this fix plus the regression test above passing locally and can open a PR.

main @ e2c19fe (1.6.0)

### Chrome version

n/a (reproducible with any; verified with Chrome 150)

ch-bas:
Fixes #2419

## Problem

`performance_start_trace` sets the context "trace is running" flag before several fallible `await`s (the `about:blank` navigation, `tracing.start`, and the navigation back to the page URL). The flag is only cleared inside `stopTracingAndAppendOutput`'s `finally`, which is never reached if any of those awaits reject. With `reload`/`autoStop` defaulting to `true`, a failed initial navigation leaves the flag stuck `true` for the rest of the session, so every later `performance_start_trace` short-circuits with "a performance trace is already running" — although no trace is running.

## Fix

Wrap the start body in `try/catch`. On failure, stop tracing if it was started (ignoring "not started" errors) and clear the flag before rethrowing, so the original error still reaches the client. When `autoStop` already ran `stopTracingAndAppendOutput`, the flag is already `false` and the catch is a no-op — no double-stop, no dangling tracer.

## Testing

- Added a regression test (`resets the running flag if a setup step throws`) that stubs the initial navigation to reject and asserts the flag is reset and that a follow-up `performance_start_trace` proceeds. It fails on `main` (`true !== false`) and passes with this change.
- `npm run test tests/tools/performance.test.ts` — all 19 tests pass.

### DIFF
--- a/src/tools/performance.ts
+++ b/src/tools/performance.ts
@@ -61,57 +61,75 @@ export const startTrace = definePageTool({
     const page = request.page;
     const pageUrlForTracing = page.pptrPage.url();
 
-    if (request.params.reload) {
-      // Before starting the recording, navigate to about:blank to clear out any state.
-      // We use `load` because `networkidle0` is known to be flaky for about:blank in Puppeteer.
-      await page.pptrPage.goto('about:blank', {
-        waitUntil: 'load',
+    try {
+      if (request.params.reload) {
+        // Before starting the recording, navigate to about:blank to clear out any state.
+        // We use `load` because `networkidle0` is known to be flaky for about:blank in Puppeteer.
+        await page.pptrPage.goto('about:blank', {
+          waitUntil: 'load',
+        });
+      }
+
+      // Keep in sync with the categories arrays in:
+      // https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/panels/timeline/TimelineController.ts
+      // https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/gather/gatherers/trace.js
+      const categories = [
+        '-*',
+        'blink.console',
+        'blink.user_timing',
+        'devtools.timeline',
+        'disabled-by-default-devtools.screenshot',
+        'disabled-by-default-devtools.timeline',
+        'disabled-by-default-devtools.timeline.invalidationTracking',
+        'disabled-by-default-devtools.timeline.frame',
+        'disabled-by-default-devtools.timeline.stack',
+        'disabled-by-default-v8.cpu_profiler',
+        'disabled-by-default-v8.cpu_profiler.hires',
+        'latencyInfo',
+        'loading',
+        'disabled-by-default-lighthouse',
+        'v8.execute',
+        'v8',
+      ];
+      await page.pptrPage.tracing.start({
+        categories,
       });
-    }
 
-    // Keep in sync with the categories arrays in:
-    // https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/panels/timeline/TimelineController.ts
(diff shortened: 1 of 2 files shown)


OrKoN (approved):
LGTM thanks!

PR #2422, opened 2026-07-27
fix(screenshot): dispose element handle after take_screenshot

### LINKED ISSUE #2421: take_screenshot leaks an ElementHandle when uid is provided (never disposed)
### Description of the bug

When `take_screenshot` is called with a `uid`, the handler resolves an element handle via `getElementByUid` but never disposes it. `getElementByUid` mints a **fresh** Puppeteer `ElementHandle` on every call (`McpPage.getElementByUid` → `#resolveElementHandle` → `node.elementHandle()`), so each element screenshot leaks a remote-object reference that lives until the page's execution context is destroyed. Over a long-running session with repeated element screenshots, these accumulate.

This diverges from every uid-based handler in `src/tools/input.ts` (`click`, `hover`, `drag`, `fill`, `upload_file`, …), which all dispose their handles in a `finally`.

**Affected code** (`src/tools/screenshot.ts`, `screenshot` handler at `main` @ e2c19fe):

```ts
const element = request.params.uid
  ? await request.page.getElementByUid(request.params.uid) // fresh handle, never disposed
  : undefined;
```

There is no `dispose()` anywhere in `screenshot.ts`. Note the leak also occurs on error paths: `getSourceBox(page, element, fullPage)` (clip computation) and `element.screenshot()` can both throw after the handle was created.

### Reproduction

1. Use `take_snapshot` to get element uids.
2. Call `take_screenshot` with a `uid` repeatedly (e.g. in a loop over many elements, or across many turns of a session).
3. Observe that retained remote objects for the page grow and are not released until navigation/context teardown.

### Expectation

The element handle is released after the screenshot is captured (or after a failed capture), matching the `input.ts` handlers.

Proposed fix — wrap clip computation + capture in `try/finally`:

```ts
const element = request.params.uid
  ? await request.page.getElementByUid(request.params.uid)
  : undefined;
try {
  // ...clip computation (getSourceBox) + the three capture branches...
} finally {
  void element?.dispose();
}
```

Impact is modest: no incorrect output, a slow resource leak only — relevant for long-lived MCP sessions taking many element screenshots. Happy to open a PR with the fix and a regression test.

main @ e2c19fe (1.6.0)

### Chrome version

ch-bas:
Fixes #2421

## Problem

When `take_screenshot` is called with a `uid`, the handler resolves an element handle via `getElementByUid` but never disposes it. `getElementByUid` mints a fresh `ElementHandle` per call (`McpPage.getElementByUid` → `node.elementHandle()`), so each element screenshot leaks a remote-object reference for the life of the page's execution context. This diverges from every uid-based handler in `input.ts`, which all dispose their handles in a `finally`.

## Fix

Wrap the clip computation and the capture in `try/finally` and dispose the handle in the `finally`, so it is also released when `getSourceBox` or the capture throws. The capture branches themselves are unchanged — this is a pure control-flow restructuring plus the disposal.

## Testing

- Added two regression tests: `disposes the element handle after an element screenshot` (success path) and `disposes the element handle when the capture fails` (capture stubbed to reject). Both fail on `main` (`dispose` called 0 times) and pass with this change.
- `npm run test tests/tools/screenshot.test.ts` — all 21 tests pass.

### DIFF
--- a/src/tools/screenshot.ts
+++ b/src/tools/screenshot.ts
@@ -156,50 +156,57 @@ export const screenshot = definePageTool(args => {
           : (request.params.quality ?? screenshotQuality);
       const fullPage = request.params.fullPage ?? false;
 
-      // Compute a downscale clip when --screenshot-max-width or
-      // --screenshot-max-height is set and the source exceeds either bound.
-      // The smaller scale factor wins so both bounds are respected while
-      // preserving aspect ratio.
-      let clip: ScreenshotClip | undefined;
-      if (
-        screenshotMaxWidth !== undefined ||
-        screenshotMaxHeight !== undefined
-      ) {
-        const box = await getSourceBox(page, element, fullPage);
-        if (box) {
-          clip = computeDownscaleClip(
-            box,
-            screenshotMaxWidth,
-            screenshotMaxHeight,
-          );
-        }
-      }
-
+      // `getElementByUid` mints a fresh ElementHandle per call, so dispose it
+      // once the capture is done to avoid leaking a remote object for the life
+      // of the page's execution context.
       let screenshot: Uint8Array;
-      if (clip) {
-        // page.screenshot with clip lets the CDP scale param downscale the
-        // capture for viewport, full-page and element shots alike. We rely on
-        // Puppeteer's default of captureBeyondViewport=true when a clip is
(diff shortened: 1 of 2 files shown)


OrKoN (approved):
LGTM thanks!

PR #2425, opened 2026-07-27
fix(script): clear dialog state after evaluate_script auto-handles a dialog

### LINKED ISSUE #2424: evaluate_script wedges all dialog-blocked tools: stale dialog state not cleared after auto-handling
### Description of the bug

After `evaluate_script` runs a script that opens a JS dialog (`alert`/`confirm`/`prompt`), every subsequent dialog-blocked tool fails with a misleading `A dialog is open (...)` error, even though the dialog was already auto-handled and is gone. The session stays wedged until `handle_dialog` is called or the page is navigated/closed.

**Root cause** (`src/tools/script.ts`, `main` @ e2c19fe):

`McpPage` attaches a persistent `'dialog'` listener that stores every dialog in `#dialog` (`McpPage.ts:155-158`). The page-evaluation path of `evaluate_script` passes `handleDialog: dialogAction ?? 'accept'` to `waitForEventsAfterAction`, so a dialog opened by the script is auto-accepted — but the page path never clears `#dialog` afterward:

```ts
// src/tools/script.ts (page path)
const result = await mcpPage.waitForEventsAfterAction(
  async () => { await performEvaluation(evaluatable, fnString, args, response, {filePath, context}); },
  {handleDialog: dialogAction ?? 'accept'},
);
response.attachWaitForResult(result);   // <-- no clearDialog()
```

This is asymmetric with the two other `handleDialog` callers, which both clear it:
- the sibling **service-worker path** 25 lines above: `if (result.dialogHandled) { getSelectedMcpPage().clearDialog(); }` (`script.ts:100-102`)
- `navigate_page`: `if (result.dialogHandled) page.clearDialog();` (`pages.ts:274-278`)

Because `#dialog` stays set to the already-accepted dialog, and `evaluate_script` plus most tools are `blockedByDialog: true`, the next such tool hits `throwIfDialogOpen()` (`ToolHandler.ts:242-244` → `McpPage.ts:214-220`) and throws.

**Blast radius:** `take_screenshot`, `take_snapshot`, all input tools (`click`, `hover`, `fill`, `type_text`, `drag`, …), `evaluate_script` itself, `performance_start_trace`/`stop`, `lighthouse_audit`, `emulate`, `take_heapsnapshot` — all `blockedByDialog: true`. Triggered by a normal script with default `dialogAction`.

### Reproduction

1. `evaluate_script` with a function that opens a dialog, e.g. `() => { confirm('x'); return 'done'; }` (default `dialogAction` = `accept`).
2. The call succeeds and the dialog is auto-accepted.
3. Call any dialog-blocked tool, e.g. `take_screenshot` → fails with `A dialog is open (confirm: x)`, although no dialog is open.

Deterministic test (fails on `main`, passes with the fix):

```ts
it('clears the dialog state after auto-handling a dialog', async () => {
  await withMcpContext(async (response, context) => {
    const mcpPage = context.getSelectedMcpPage();
    await mcpPage.pptrPage.setContent(html`<button id="test">test</button>`);
    await evaluateScript().handler(
      {params: {function: String(() => { confirm('hello'); return 'done'; })}},
      response,
      context,
    );
    // On main: getDialog() returns the already-accepted dialog (handled: true).
    assert.strictEqual(mcpPage.getDialog(), undefined);
    assert.doesNotThrow(() => mcpPage.throwIfDialogOpen());
  });
});
```

### Expectation

After `evaluate_script` auto-handles a dialog, the stored dialog reference is cleared (as the worker path and `navigate_page` already do), so subsequent tools are not blocked by a phantom dialog.

Fix: on the page path, `if (result.dialogHandled) mcpPage.clearDialog();` before attaching the result. I have this plus the regression test passing locally and can open a PR.

main @ e2c19fe (1.6.0)

### Chrome version

ch-bas:
Fixes #2424

## Problem

After `evaluate_script` runs a script that opens a JS dialog (`alert`/`confirm`/`prompt`), every subsequent dialog-blocked tool fails with a misleading `A dialog is open (...)` error — even though the dialog was already auto-handled and is gone. The session stays wedged until `handle_dialog` is called or the page is navigated/closed.

`McpPage` stores every dialog in `#dialog` via a persistent `'dialog'` listener. The page-evaluation path of `evaluate_script` passes `handleDialog` (default `'accept'`) so the dialog is auto-accepted, but — unlike the sibling service-worker path (`script.ts:100-102`) and `navigate_page` (`pages.ts:274-278`) — it never clears `#dialog`. Since `evaluate_script` and most tools are `blockedByDialog: true`, the next one throws on `throwIfDialogOpen()`.

## Fix

On the page path, clear the dialog state when a dialog was handled, mirroring the worker path:

```ts
if (result.dialogHandled) {
  mcpPage.clearDialog();
}
```

## Testing

- Added `clears the dialog state after auto-handling a dialog`: evaluates `() => { confirm('hello'); return 'done'; }` and asserts `getDialog()` is `undefined` and `throwIfDialogOpen()` doesn't throw. Fails on `main` (`getDialog()` returns the already-accepted dialog) and passes with this change.
- `npm run test tests/tools/script.test.ts` — all tests pass.

### DIFF
--- a/src/tools/script.ts
+++ b/src/tools/script.ts
@@ -129,6 +129,9 @@ Example with arguments: \`(el) => el.innerText\`
           },
           {handleDialog: dialogAction ?? 'accept'},
         );
+        if (result.dialogHandled) {
+          mcpPage.clearDialog();
+        }
         response.attachWaitForResult(result);
       } finally {
         void Promise.allSettled(args.map(arg => arg.dispose()));
--- a/tests/tools/script.test.ts
+++ b/tests/tools/script.test.ts
@@ -150,6 +150,34 @@ describe('script', () => {
       });
     });
 
+    it('clears the dialog state after auto-handling a dialog', async () => {
+      await withMcpContext(async (response, context) => {
+        const mcpPage = context.getSelectedMcpPage();
+        await mcpPage.pptrPage.setContent(
+          html`<button id="test">test</button>`,
+        );
+
+        await evaluateScript().handler(
+          {
+            params: {
+              function: String(() => {
+                confirm('hello');
+                return 'done';
+              }),
+            },
+          },
+          response,
+          context,
+        );
+
+        // After the dialog is auto-accepted, the stored dialog reference must
+        // be cleared. Otherwise throwIfDialogOpen() wedges every subsequent
+        // blockedByDialog tool with a false "A dialog is open" error.
+        assert.strictEqual(mcpPage.getDialog(), undefined);
+        assert.doesNotThrow(() => mcpPage.throwIfDialogOpen());
+      });
+    });
+
     it('work for scripts that trigger prompts and fill them', async () => {
       await withMcpContext(async (response, context) => {
         const page = context.getSelectedMcpPage().pptrPage;


Lightning00Blade (changes requested):
Thank you for the PR,
I recently added a change to Puppeteer, it will allows to track the dialog state regardless where it got handled.
https://github.com/puppeteer/puppeteer/pull/15225
Then we can handle this inside `throwIfDialogOpen`, which would mean we don't have to track each individual place this could happen.
Because of that I will close this PR.

PR #2427, opened 2026-07-27
fix(wait): avoid 180s mutex stall when a dialog opens during an action

### LINKED ISSUE #2426: A dialog opened during a no-handleDialog action freezes the whole session for ~180s (mutex stall in waitForStableDom)
### Description of the bug

When a tool that passes **no** `handleDialog` (`click`, `hover`, `fill`, `type_text`, `drag`, `new_page`, …) runs an action that opens a JS dialog (`alert`/`confirm`/`prompt`) — e.g. `onclick="setTimeout(() => confirm('save?'), 0)"`, an alert-on-load page, or a confirm-on-submit — the whole MCP session freezes for ~180 seconds. Every queued tool call, **including `handle_dialog` (the only remedy), is blocked** for the duration.

**Root cause** (`main` @ e2c19fe):

1. `McpPage` attaches a persistent `'dialog'` listener that only **stores** the dialog (`McpPage.ts:155-158`); it does not accept/dismiss it. So a dialog opened by a no-`handleDialog` action stays open and pauses the renderer.
2. `WaitForHelper.waitForEventsAfterAction` only registers its dialog handler when `handleDialog` is passed, so `#dialogOpened` stays false and it proceeds to `waitForStableDom()` (`WaitForHelper.ts:196-202`).
3. `waitForStableDom`'s setup evaluation is **unguarded**: `const stableDomObserver = await this.#page.evaluateHandle(...)` (`WaitForHelper.ts:43`). The `Promise.race` with the stable-DOM timeout wraps only the *second* evaluate. An open dialog pauses the renderer, so this `evaluateHandle` hangs.
4. No `protocolTimeout` is configured in `browser.ts` (launch/connect), so Puppeteer's default **180000ms** applies before the hung call finally rejects.
5. This await runs while `ToolHandler` holds the single tool `Mutex` (acquired at `ToolHandler.ts:208`, released only in the `finally` at `:313`). `Mutex` is strict FIFO with **no timeout** (`utils/Mutex.ts`), so all queued calls — including `handle_dialog` — wait ~180s.

I verified step 3/4 empirically: with a dialog left open (stored, not handled), `page.evaluateHandle()` hangs until `protocolTimeout` and then throws `ProtocolError`.

This is distinct from the `evaluate_script` dialog-state issue: that one is about tools that *do* pass `handleDialog` not clearing state; this is about tools that *don't* pass it stalling on the DOM-stability wait.

### Reproduction

1. Snapshot a page with `<button onclick="setTimeout(() => confirm('x'), 0)">go</button>`.
2. `click` the button (no `handleDialog` is passed by the click tool).
3. The dialog opens after the click; the click call hangs ~180s holding the mutex.
4. Any other tool call, including `handle_dialog`, is queued and blocked until it unwinds.

A `navigate_page` to a page that calls `alert()` on load hits the same stall (its `handleDialog` only matches `beforeunload`, so an `alert` is unhandled).

### Expectation

An open dialog during an action should not stall the session. The action should return promptly; the stored dialog then makes the next dialog-blocked tool fail fast with "A dialog is open", prompting `handle_dialog`.

Proposed fix (two parts):
- **(A)** Always register the dialog listener in `WaitForHelper` (not only when `handleDialog` is set) and record any observed dialog in a separate flag, then skip `waitForStableDom` when it is set. The new flag does not affect `dialogHandled`, so callers that `clearDialog()` on `dialogHandled` never clear a genuinely-open dialog.
- **(B)** Defense-in-depth: bound the `waitForStableDom` setup evaluation with the abort signal and the stable-DOM timeout so any renderer pause can no longer exceed the intended cap.

I have both plus fast, deterministic regression tests passing locally and can open a PR.

main @ e2c19fe (1.6.0)

### Chrome version

ch-bas:
Fixes #2426

## Problem

When a tool that passes no `handleDialog` (`click`, `hover`, `fill`, `type_text`, `drag`, `new_page`) runs an action that opens a JS dialog, `McpPage`'s persistent listener stores the dialog but never dismisses it, so the renderer stays paused. `WaitForHelper` then ran `waitForStableDom`, whose setup `evaluateHandle` (`WaitForHelper.ts:43`) hung until Puppeteer's default `protocolTimeout` (**180s**, none is configured). That await happens while `ToolHandler` holds the single, timeout-less tool mutex, so every queued call — including `handle_dialog`, the only remedy — was blocked ~180s, freezing the session.

Verified empirically: with a dialog left open, `page.evaluateHandle()` hangs until `protocolTimeout` then throws `ProtocolError`.

## Fix

- **(A) Correctness:** always register the dialog listener (not only when `handleDialog` is set) and record any observed dialog in a separate `#dialogDetected` flag, then skip `waitForStableDom` when it is set. `#dialogDetected` does **not** affect `dialogHandled`, so callers that `clearDialog()` on `dialogHandled` (`navigate_page`, the worker/page paths of `evaluate_script`) are unaffected and never clear a genuinely-open dialog. The action returns promptly and the next dialog-blocked tool fails fast with "A dialog is open", prompting `handle_dialog`.
- **(B) Defense-in-depth:** bound the `waitForStableDom` setup evaluation with the abort signal and the stable-DOM timeout, so any renderer pause can no longer exceed the intended cap.

## Testing

- New `tests/WaitForHelper.test.ts` with a hanging `evaluateHandle` stub (fast + deterministic, no 180s wait):
  - `does not hang when the DOM-stability setup never resolves` — covers (B).
  - `skips the DOM-stability wait when a dialog opens (no handleDialog)` — covers (A); also asserts `dialogHandled` stays `false`.
  - Both fail on `main` (hang until the test's bound) and pass with this change.
- Re-ran the suites that exercise `waitForEventsAfterAction` (`input`, `pages`, `script`, `snapshot`) — all pass.
- ESLint + Prettier clean.

### DIFF
--- a/src/WaitForHelper.ts
+++ b/src/WaitForHelper.ts
@@ -19,6 +19,10 @@ export class WaitForHelper {
   #navigationTimeout: number;
 
   #dialogOpened = false;
+  // Set when any dialog is observed during the action, whether or not this
+  // helper handled it. A dialog pauses the renderer, so DOM-stability
+  // evaluation would hang until protocolTimeout; we use this to skip it.
+  #dialogDetected = false;
   #initialUrl: string;
 
   constructor(
@@ -40,31 +44,45 @@ export class WaitForHelper {
    * for the DOM to be stable before returning.
    */
   async waitForStableDom(): Promise<void> {
-    const stableDomObserver = await this.#page.evaluateHandle(timeout => {
-      let timeoutId: ReturnType<typeof setTimeout>;
-      function callback() {
-        clearTimeout(timeoutId);
-        timeoutId = setTimeout(() => {
-          domObserver.resolver.resolve();
-          domObserver.observer.disconnect();
-        }, timeout);
-      }
-      const domObserver = {
-        resolver: Promise.withResolvers<void>(),
-        observer: new MutationObserver(callback),
-      };
-      // It's possible that the DOM is not gonna change so we
-      // need to start the timeout initially.
-      callback();
+    // Bound the setup evaluation against the abort signal and the stable-DOM
+    // timeout. Without this cap a paused renderer (e.g. an open dialog) would
+    // make evaluateHandle hang until protocolTimeout (default 180s) while the
+    // tool mutex is held. The `.catch` swallows the eventual ProtocolError if
+    // the timeout wins first, so it does not surface as an unhandled rejection.
+    const stableDomObserver = await Promise.race([
+      this.#page
+        .evaluateHandle(timeout => {
+          let timeoutId: ReturnType<typeof setTimeout>;
+          function callback() {
+            clearTimeout(timeoutId);
+            timeoutId = setTimeout(() => {
+              domObserver.resolver.resolve();
+              domObserver.observer.disconnect();
+            }, timeout);
+          }
+          const domObserver = {
+            resolver: Promise.withResolvers<void>(),
+            observer: new MutationObserver(callback),
+          };
+          // It's possible that the DOM is not gonna change so we
+          // need to start the timeout initially.
+          callback();
 
-      domObserver.observer.observe(document.body, {
-        childList: true,
-        subtree: true,
-        attributes: true,
-      });
+          domObserver.observer.observe(document.body, {
+            childList: true,
+            subtree: true,
+            attributes: true,
+          });
 
-      return domObserver;
-    }, this.#stableDomFor);
+          return domObserver;
+        }, this.#stableDomFor)
+        .catch(() => undefined),
+      this.timeout(this.#stableDomTimeout).then(() => undefined),
+    ]);
+
+    if (!stableDomObserver) {
+      return;
+    }
 
     this.#abortController.signal.addEventListener('abort', async () => {
       try {
@@ -141,34 +159,42 @@ export class WaitForHelper {
     if (this.#abortController.signal.aborted) {
       throw new Error("Can't re-use a WaitForHelper");
     }
-    if (options?.handleDialog) {
-      const dialogHandler = (
-        dialog: Pick<Dialog, 'accept' | 'dismiss' | 'type'>,
-      ) => {
-        let actionToTake: DialogAction | undefined;
+    const dialogHandler = (
+      dialog: Pick<Dialog, 'accept' | 'dismiss' | 'type'>,
+    ) => {
+      // Record every dialog so waitForStableDom is skipped: an open dialog
+      // pauses the renderer, so the DOM-stability evaluation would otherwise
+      // hang until protocolTimeout (default 180s) while holding the tool
+      // mutex, wedging the whole session.
+      this.#dialogDetected = true;
 
-        if (typeof options.handleDialog === 'object') {
-          actionToTake = options.handleDialog[dialog.type()];
+      if (!options?.handleDialog) {
+        return;
+      }
+
+      let actionToTake: DialogAction | undefined;
+
+      if (typeof options.handleDialog === 'object') {
+        actionToTake = options.handleDialog[dialog.type()];
+      } else {
+        actionToTake = options.handleDialog;
+      }
+
+      if (actionToTake) {
+        this.#dialogOpened = true;
+        if (actionToTake === 'dismiss') {
+          void dialog.dismiss();
+        } else if (actionToTake === 'accept') {
+          void dialog.accept();
         } else {
-          actionToTake = options.handleDialog;
+          void dialog.accept(actionToTake);
         }
-
-        if (actionToTake) {
-          this.#dialogOpened = true;
-          if (actionToTake === 'dismiss') {
-            void dialog.dismiss();
-          } else if (actionToTake === 'accept') {
-            void dialog.accept();
-          } else {
-            void dialog.accept(actionToTake);
-          }
-        }
-      };
-      this.#page.on('dialog', dialogHandler);
-      this.#abortController.signal.addEventListener('abort', () => {
-        this.#page.off('dialog', dialogHandler);
-      });
-    }
+      }
+    };
+    this.#page.on('dialog', dialogHandler);
+    this.#abortController.signal.addEventListener('abort', () => {
+      this.#page.off('dialog', dialogHandler);
+    });
 
     const navigationFinished = this.waitForNavigationStarted()
       .then(navigationStated => {
@@ -193,7 +219,7 @@ export class WaitForHelper {
     try {
       await navigationFinished;
 
-      if (this.#dialogOpened) {
+      if (this.#dialogOpened || this.#dialogDetected) {
         return this.#getResult();
       }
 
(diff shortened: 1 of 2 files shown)


src/WaitForHelper.ts
```diff
@@ -40,31 +44,45 @@ export class WaitForHelper {
...
+            childList: true,
+            subtree: true,
+            attributes: true,
+          });
+
+          return domObserver;
+        }, this.#stableDomFor)
+        .catch(() => undefined),
+      this.timeout(this.#stableDomTimeout).then(() => undefined),
+    ]);
```
Lightning00Blade:
```suggestion
      this.timeout(this.#stableDomTimeout),
    ]).catch(() => {});
```

src/WaitForHelper.ts
```diff
@@ -193,7 +219,7 @@ export class WaitForHelper {
     try {
       await navigationFinished;
 
-      if (this.#dialogOpened) {
+      if (this.#dialogOpened || this.#dialogDetected) {
```
Lightning00Blade: As we always want to exit early, and in the case of open this will always be true.
```suggestion
      if (this.#dialogDetected) {
```

tests/WaitForHelper.test.ts
```diff
@@ -0,0 +1,102 @@
...
+import type {Page} from '../src/third_party/index.js';
+import {WaitForHelper} from '../src/WaitForHelper.js';
+
+/**
+ * Builds a minimal fake Puppeteer page that lets tests control the
+ * DOM-stability setup evaluation and emit dialog events. The real thing would
+ * pause its renderer while a dialog is open, so `evaluateHandle` never
+ * resolves; a never-resolving stub models that without a live browser.
+ */
+function createFakePage(evaluateHandle: () => Promise<unknown>): {
```
Lightning00Blade: Let's try to make a real test, rather then mocking things.

src/WaitForHelper.ts
```diff
@@ -19,6 +19,10 @@ export class WaitForHelper {
   #navigationTimeout: number;
 
   #dialogOpened = false;
+  // Set when any dialog is observed during the action, whether or not this
+  // helper handled it. A dialog pauses the renderer, so DOM-stability
+  // evaluation would hang until protocolTimeout; we use this to skip it.
+  #dialogDetected = false;
```
Lightning00Blade, line 23: Let's rename to  #dialogOpened -> #dialogHandled.
And remove the comment.

Lightning00Blade (commented):
Code wide looks good, lets change the test to reuse a real page.

ch-bas:
Thanks for the review! Addressed in the follow-up commit:

- Renamed `#dialogOpened` → `#dialogHandled` and dropped the comment.
- Skip the DOM-stability wait on `#dialogDetected` alone — it's set for every observed dialog, so it already covers the handled case.
- Simplified the `waitForStableDom` setup race to a single trailing `.catch(...)`.
- Rewrote the regression test to drive a **real page** via `withMcpContext` (opens a dialog from the action with no `handleDialog`, asserts the call returns promptly instead of stalling and that the dialog is still recorded).

One small note: I used `.catch(() => undefined)` instead of the suggested `.catch(() => {})` because the repo's `@typescript-eslint/no-empty-function` rule rejects the empty body — same behavior, lint-clean.

tests/WaitForHelper.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        new Promise<'stalled'>(resolve =>
+          setTimeout(() => resolve('stalled'), 15_000),
+        ),
+      ]);
+
+      if (result === 'stalled') {
+        assert.fail(
+          'waitForEventsAfterAction stalled while a dialog was open; the tool mutex would be held ~180s',
+        );
+      }
```
Lightning00Blade:
```suggestion
        assert(result !== 'stalled',
          'stalled because dialog was shown. Would timeout with ProtocolError.
        );
```

tests/WaitForHelper.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      // The dialog was detected but not handled (no handleDialog was passed).
+      assert.strictEqual(result.dialogHandled, false);
+      // The dialog is still open and recorded, so the next blockedByDialog tool
+      // correctly refuses to run.
+      assert.throws(() => mcpPage.throwIfDialogOpen());
+
+      // Clean up so the page can be torn down.
```
Lightning00Blade: Were there any issue with that, I think Puppeteer can pages with open dialog?

tests/WaitForHelper.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      // (~180s) while the tool mutex is held, freezing the session. Racing the
+      // call against a short timeout proves it returns promptly instead.
+      const result = await Promise.race([
+        mcpPage.waitForEventsAfterAction(async () => {
+          await mcpPage.pptrPage.evaluate(() => {
+            setTimeout(() => confirm('blocked?'), 0);
+          });
+        }),
+        new Promise<'stalled'>(resolve =>
+          setTimeout(() => resolve('stalled'), 15_000),
```
Lightning00Blade: This can be dropped to `5_000` and let's add a comment that this is relate to the `WaitForHelper.#stableDomTimeout`

tests/WaitForHelper.test.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
Lightning00Blade:
```suggestion
 * Copyright 2026 Google LLC
```

src/WaitForHelper.ts
```diff
@@ -141,34 +152,42 @@ export class WaitForHelper {
...
-        } else {
-          actionToTake = options.handleDialog;
-        }
+    const dialogHandler = (
+      dialog: Pick<Dialog, 'accept' | 'dismiss' | 'type'>,
+    ) => {
+      // Record every dialog so waitForStableDom is skipped: an open dialog
+      // pauses the renderer, so the DOM-stability evaluation would otherwise
+      // hang until protocolTimeout (default 180s) while holding the tool
+      // mutex, wedging the whole session.
```
Lightning00Blade: Let's move this a JSDoc on the property and just say:
`Track all dialogs as they pause the renderer`

Lightning00Blade (approved):

ch-bas:
Thanks for the approval! Addressed the nits in the follow-up commit:

- Moved the dialog-tracking note to a JSDoc on `#dialogDetected` (`Track all dialogs as they pause the renderer`).
- Test: fixed the copyright year, dropped the stall timeout to `5_000` with a comment tying it to `WaitForHelper.#stableDomTimeout`, and switched the stall check to `assert(result !== 'stalled', ...)`.
- Re the cleanup question — you're right: I verified the page tears down fine with an open dialog, so I removed the manual `dismiss()`/`clearDialog()`.

Verified against the freshly-merged base (puppeteer 25.4.0): the WaitForHelper test plus the input/script suites pass, and lint is clean.

Lightning00Blade (approved):

PR #2428, opened 2026-07-27
fix: dispose heap snapshot workers on context teardown

gunjanjaswal:
Heap snapshot worker threads (Node worker threads, independent of the browser) and their parsed snapshots are only freed per file, through the `close_heapsnapshot` tool. On context teardown and on browser reconnect they leak.

`McpContext.dispose()` cleans up the service-worker console collector and each `McpPage`, but never touches the `HeapSnapshotManager`, so any snapshot loaded during the session keeps its worker alive. And `getContext()` in `index.ts` replaces the previous context on reconnect (`if (context?.browser !== browser)`) without ever calling `dispose()` on the old one, so every reconnect strands the previous context's workers and retained snapshots. Repeated reconnects accumulate the leak.

This adds `HeapSnapshotManager.disposeAll()` (terminates every cached worker and clears the map, mirroring the existing per-file `dispose`), calls it from `McpContext.dispose()`, and disposes the outgoing context before it is replaced on reconnect. A regression test in `McpContext.test.ts` loads a fixture snapshot and asserts it is freed after `dispose()`.

Verified on Node 24 (the `.nvmrc` version): typecheck, eslint, and the affected test files pass, and `npm run gen` leaves the tree clean, so there is no generated-files drift. Reverting the source makes the new test fail.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -147,8 +147,6 @@ export async function createMcpServer(
           });
 
     if (context?.browser !== browser) {
-      // Free resources tied to the previous context (heap snapshot workers,
-      // console collectors, page listeners) before replacing it on reconnect.
       context?.dispose();
       context = await McpContext.from(browser, logger, {
         experimentalDevToolsDebugging: devtools,
--- a/src/index.ts
+++ b/src/index.ts
@@ -147,6 +147,9 @@ export async function createMcpServer(
           });
 
     if (context?.browser !== browser) {
+      // Free resources tied to the previous context (heap snapshot workers,
+      // console collectors, page listeners) before replacing it on reconnect.
+      context?.dispose();
       context = await McpContext.from(browser, logger, {
         experimentalDevToolsDebugging: devtools,
         experimentalIncludeAllPages: serverArgs.experimentalIncludeAllPages,
--- a/tests/McpContext.test.ts
+++ b/tests/McpContext.test.ts
@@ -285,6 +285,24 @@ describe('McpContext', () => {
     });
   });
 
+  it('disposes loaded heap snapshots on teardown', async () => {
+    await withMcpContext(async (_response, context) => {
+      const filePath = path.join(
+        process.cwd(),
+        'tests/fixtures/example.heapsnapshot',
+      );
+      await context.getHeapSnapshotStats(filePath);
+      assert.ok(context.hasHeapSnapshots(), 'snapshot loaded before teardown');
+
+      context.dispose();
+
+      assert.ok(
+        !context.hasHeapSnapshots(),
+        'heap snapshots freed on teardown',
+      );
+    });
+  });
+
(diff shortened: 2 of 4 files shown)


Lightning00Blade (commented):
Thanks for the PR!

I think it make sense adding the dispose on context re-recreation.
I think we should also add test for that if possible as it's the more important part of the code change.

gunjanjaswal:
Good call, added a test that covers the reconnect path directly. It loads a snapshot, disposes the context the way `getContext()` does before replacing it on reconnect, then asserts the cached worker thread actually gets terminated (it spies on the worker proxy `dispose` and calls through, so it is the real teardown rather than a mock). Reverting the `disposeAll()` line makes it fail, so it is pinned to the fix.

tests/McpContext.test.ts
```diff
@@ -285,6 +285,84 @@ describe('McpContext', () => {
...
+      context.dispose();
+
+      assert.ok(
+        !context.hasHeapSnapshots(),
+        'heap snapshots freed on teardown',
+      );
+    });
+  });
+
+  it('terminates heap snapshot workers when the context is recreated on reconnect', async () => {
```
Lightning00Blade: This test does not actually test the mechanism but rather mimics it. 
Given that we don't have the infra setup for that, let's just remove the test 
`terminates heap snapshot workers when the context is recreated on reconnect`
And leave the rest as is.

src/index.ts
```diff
@@ -147,6 +147,9 @@ export async function createMcpServer(
           });
 
     if (context?.browser !== browser) {
+      // Free resources tied to the previous context (heap snapshot workers,
+      // console collectors, page listeners) before replacing it on reconnect.
```
Lightning00Blade: We can remove it `dispose` is a commonly known operation.

Lightning00Blade (commented):
LGTM, will approve after the comments are fixed.

gunjanjaswal:
Both done. Dropped the reconnect worker test (fair point, without the infra it was mirroring the code rather than exercising it) and removed the comment on the dispose call. The teardown test that checks snapshots are freed stays. typecheck, eslint, and gen are clean.

Lightning00Blade (approved):
LGTM, thanks!

gunjanjaswal:
Thanks for the review. Quick note on the red: the macOS / node 22 job failed on one test, `pages.test.js` "list pages for extension service workers", hitting the 30s Puppeteer timeout. It is unrelated to this change (which only dropped a test and a comment), it is green on the other OS/node jobs here, and that macOS job has been flaky across other PRs too. Should clear on a re-run.

gunjanjaswal:
Thanks for the review and merge, @Lightning00Blade. Good call trimming that reconnect test down to what the infra can actually exercise.

Issue #2270, opened 2026-06-29
Expose the experimental CDP PWA domain (install/launch/uninstall) as tools


LuHuangMSFT:
### Is your feature request related to a problem? Please describe.

There is currently no way for an MCP client/agent to drive **Progressive Web App lifecycle actions** through `chrome-devtools-mcp`. The available tools reach page content (DOM, console, network, input), but installing/launching/uninstalling a PWA happens in the **browser chrome** (the omnibox install button, the app window), which is out of reach for the existing tools.

This blocks an entire class of automation and debugging use cases:

- Agents/tests that need to **install a web app and verify OS integration** (manifest id, display mode, file/protocol handlers, badging).
- Reproducing and debugging **install-state-dependent behavior** (e.g. APIs that only change in a standalone app window vs. a tab: `display-mode` media queries, `windowControlsOverlay`, launch handlers).
- End-to-end PWA flows without a manual user gesture on the native install prompt.

Chromium already implements the **`PWA` CDP domain** (`install`, `uninstall`, `launch`, `launchFilesInApp`, `openCurrentPageInApp`, `getOsAppState`, `changeAppUserSettings`) in `chrome/browser/devtools/protocol/pwa_handler.cc`. The capability exists at the protocol layer — it just isn't surfaced as MCP tools.

### Describe the solution you'd like

Add a new **opt-in tool category** (e.g. `--category-pwa`, off by default like `extensions`/`webmcp`) that wraps the `PWA` CDP domain. Proposed tools:

- `install_pwa` — `PWA.install` (manifestId, optional installUrlOrBundleUrl)
- `uninstall_pwa` — `PWA.uninstall`
- `launch_pwa` — `PWA.launch` (opens the app window)
- `open_current_page_in_app` — `PWA.openCurrentPageInApp`
- `get_os_app_state` — `PWA.getOsAppState` (badge count, file handlers)
- `install_current_page_as_pwa` — convenience: reads the page manifest, derives the manifest id, installs (no manifestId argument needed)

Because `PWA.*` is **browser-scoped**, the tools open a browser-level CDP session (`browser.target().createCDPSession()`) rather than a page session.

**Note on display mode:** CDP installs default `user_display_mode` to `kBrowser` (they don't simulate the install dialog), so launched apps open as tabs. The tools expose an optional `displayMode: "standalone" | "browser"` that chains `PWA.changeAppUserSettings` after install to produce a real standalone app window.

[...]

PR #2430, opened 2026-07-27
feat: add PWA automation tools

LuHuangMSFT:
## What / Why

Adds an opt-in Progressive Web Apps tool category that exposes the installed web app lifecycle to MCP clients and agents.

The browser-level PWA APIs were introduced in Puppeteer 25.4.0 (puppeteer/puppeteer#15235). The current base uses Puppeteer 25.5.0, which also enforces URL restrictions for browser-level PWA operations (puppeteer/puppeteer#15271).

## Tools

- `install_pwa`
- `launch_pwa`
- `get_os_app_state`
- `uninstall_pwa`

The category is disabled by default and enabled with `--category-pwa`.

These are browser-scoped operations and do not require a selected page. Callers provide the resolved manifest ID and, for installation, an explicit page or bundle URL. `open_current_page_in_app` and automatic current-page installation are intentionally deferred until Puppeteer exposes page-native APIs with reliable loaded-document and browser-context semantics.

## Compatibility and safety

- Requires a pipe-launched browser. The CLI rejects `--category-pwa` when combined with `--auto-connect`, `--browser-url`, or `--ws-endpoint`.
- Puppeteer rejects PWA operations when URL allow/block restrictions are configured, preventing browser-internal PWA fetch and redirect flows from bypassing the configured policy.
- File-based install bundles are validated against the MCP filesystem roots.
- DevTools installation defaults to browser display mode unless `displayMode: "standalone"` is supplied. Launch uses the app's saved display preference.

## Testing

- PWA tests cover lifecycle/state, standalone launch, successful explicit-URL launch, use without a selected page, file-bundle path validation, OS-state output, and selected-page fallback after uninstall.
- Existing off-by-default category tests cover PWA tool registration and opt-in exposure.
- Generated README, tool reference, CLI metadata, and telemetry metrics are updated.
- Type checking, formatting, targeted tests, broad supporting repository tests, third-party notice validation, and diff checks pass locally.
- Validated against Chrome for Testing 150 and a local Chromium/Edge build.
- A real stdio MCP lifecycle passed end to end with the final four-tool design: category gating, explicit install, OS state, standalone launch, selecting and scripting the returned page, uninstall, selection fallback, and post-uninstall state rejection.

Refs #2270


LuHuangMSFT:
@OrKoN Puppeteer 25.4.0 now contains the merged browser-level PWA APIs, and the MCP integration is ready for review.

src/tools/pwa.ts
```diff
@@ -0,0 +1,166 @@
...
+import {fileURLToPath} from 'node:url';
+
+import {zod} from '../third_party/index.js';
+
+import {ToolCategory} from './categories.js';
+import type {Context} from './ToolDefinition.js';
+import {defineTool} from './ToolDefinition.js';
+
+function ensureNoNetworkRestrictions(context: Context): void {
+  if (context.hasNetworkRestrictions()) {
```
OrKoN: Any reason for this?

src/tools/ToolDefinition.ts
```diff
@@ -193,7 +194,9 @@ export type SupportedExtensions =
  * Only add methods used by tools/*.
  */
 export type Context = Readonly<{
+  readonly browser: Browser;
```
OrKoN: Let's avoid exposing the entire browser to all tools. Instead let's expose methods tools would need.

src/bin/chrome-devtools-mcp-cli-options.ts
```diff
@@ -394,6 +401,14 @@ export function parseArguments(
         );
         args.usageStatistics = false;
       }
+      if (
```
OrKoN: could we use the `conflicts` configuration for the `--category-pwa` option instead of validating separately?

tests/cli.test.ts
```diff
@@ -278,6 +282,28 @@ describe('cli args parsing', () => {
     });
   });
 
+  it('rejects PWA tools with connected browser modes', () => {
```
OrKoN: this file should not spawn a subprocess but instead test by calling `parseArguments`

package.json
```diff
@@ -76,7 +76,7 @@
     "globals": "^17.0.0",
     "lighthouse": "13.4.0",
     "prettier": "^3.6.2",
-    "puppeteer": "25.3.0",
+    "puppeteer": "25.4.0",
```
OrKoN: let's update Puppeteer in a separate PR https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2432.

OrKoN (changes requested):
Thanks for the PR. A few comments below:

src/telemetry/tool_call_metrics.json
```diff
@@ -905,5 +905,63 @@
...
+    "name": "get_os_app_state",
+    "args": [
+      {
+        "name": "manifest_id_length",
+        "argType": "number"
+      }
+    ]
+  },
+  {
+    "name": "install_current_page_as_pwa",
```
OrKoN: where is the install_current_page_as_pwa tool defined?

src/tools/pwa.ts
LuHuangMSFT: installPWA and launchPWA are browser-level operations, while our URL allow/block rules protect page-target requests. Chrome can perform PWA fetches and redirects outside that enforcement path, so allowing these tools could bypass the configured URL policy. Until those requests can be restricted reliably, we reject install and launch whenever URL restrictions are enabled.

src/tools/ToolDefinition.ts
LuHuangMSFT: Replaced `Context.browser` with four narrow PWA methods implemented by `McpContext`, so other tools no longer receive the full Puppeteer browser object.

> [Commit 41f4051](https://github.com/ChromeDevTools/chrome-devtools-mcp/commit/41f4051)

src/bin/chrome-devtools-mcp-cli-options.ts
LuHuangMSFT: Moved the incompatible connection modes into `categoryPwa.conflicts` and removed the separate post-parse validation.

> [Commit 6a1f501](https://github.com/ChromeDevTools/chrome-devtools-mcp/commit/6a1f501)

tests/cli.test.ts
LuHuangMSFT: The test now calls `parseArguments` directly. I also removed the false default from `categoryPwa` so yargs applies the conflicts only when the category is enabled.

> [Commit 24c8d10](https://github.com/ChromeDevTools/chrome-devtools-mcp/commit/24c8d10)

package.json
LuHuangMSFT: Rebased the branch onto `main`, which now contains #2432. The Puppeteer dependency and lockfile are no longer changed by this PR.

src/telemetry/tool_call_metrics.json
LuHuangMSFT: I decided to remove that function but missed this. Removed.

> [Commit 5054315](https://github.com/ChromeDevTools/chrome-devtools-mcp/commit/5054315f344ba67b55dfcc4dc1086da8881ca47d)

LuHuangMSFT:
@OrKoN Addressed all six review threads and force-pushed the rebased branch at f648aad. The dependency bump is now supplied by merged #2432, the PR body reflects the final four-tool design.

src/tools/pwa.ts
OrKoN: @LuHuangMSFT Puppeteer configures the network restrictions for all CDP targets. Do you have a test that verifies that PWA can bypass those permissions?

LuHuangMSFT: Yes. I ran two temporary end-to-end probes against Puppeteer’s configured `blockedUrlPattern`.

**`launchPWA` redirect probe**

```ts
const launchUrl = `${server.baseUrl}/pwa/launch-redirect.html`;
const blockedUrl = `${server.baseUrl}/pwa/blocked.html`;
let blockedRequests = 0;

server.addRoute('/pwa/launch-redirect.html', (_req, res) => {
  res.statusCode = 302;
  res.setHeader('Location', blockedUrl);
  res.end();
});
server.addRoute('/pwa/blocked.html', (_req, res) => {
  blockedRequests++;
  res.end('blocked');
});

await withMcpContext(
  async (_response, context) => {
    await context.browser.launchPWA({manifestId, url: launchUrl});
    await new Promise(resolve => setTimeout(resolve, 1000));
    assert.strictEqual(blockedRequests, 0);
  },
  {blockedUrlPattern: [blockedUrl]},
);
```

The allowed, in-scope launch URL redirected to the blocked URL. The assertion failed with:

```text
Expected: 0
Actual:   1
```

**`installPWA` probe**

I served a valid installable PWA at `blockedUrl`, configured that exact URL in `blockedUrlPattern`, and called:

```ts
await withMcpContext(
  async (_response, context) => {
    await context.browser.installPWA({
      manifestId,
      installUrlOrBundleUrl: blockedUrl,
      displayMode: 'standalone',
    });
    assert.strictEqual(blockedRequests, 0);
  },
  {blockedUrlPattern: [blockedUrl]},
);
```

Installation completed, but the assertion again failed with:

```text
Expected: 0
Actual:   1
```

I also verified the install fixture without the blocklist first. These results directly demonstrate that both `PWA.launch` navigation and `PWA.install` fetching can reach URLs blocked by Puppeteer’s target-level restrictions.

OrKoN: Let me double check this. I think we would need to go back to Puppeteer and fix it this in there instead since the restrictions are enforced by Puppeteer.

src/bin/chrome-devtools-mcp-cli-options.ts
```diff
@@ -373,8 +380,10 @@ export function parseArguments(
   version: string,
   argv = process.argv,
   env = process.env,
+  exitProcess = true,
 ) {
   const yargsInstance = yargs(hideBin(argv))
+    .exitProcess(exitProcess)
```
OrKoN: why is this change needed?

src/tools/pwa.ts
OrKoN: https://github.com/puppeteer/puppeteer/pull/15271

LuHuangMSFT: Thanks. Will wait for deps bump to rebase.

src/bin/chrome-devtools-mcp-cli-options.ts
LuHuangMSFT: This was added only to support the new direct conflict test. Looking at existing repository precedent, we do not separately test Yargs conflicts declarations, and the existing parser tests for the affected connection modes already catch unintended regressions such as the default: false issue. I’ll remove both this parameter and the redundant conflict test.

LuHuangMSFT: Removed in [13a2cc0](https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2430/commits/13a2cc0b8e76f18d7795ab27880986d988bccde7)

LuHuangMSFT:
@OrKoN Puppeteer 25.5.0 is now on main, so I rebased and removed the duplicate MCP network-restriction guard. Ready for another look. Thanks!

OrKoN (approved):

Issue #2431, opened 2026-07-28
Memory leak on 1.6.0 (no autoConnect): main-isolate old space grows ~13 MiB/min per active session, plateaus when idle, cyclic heap-limit OOMs


milordy-coutinho:
### Environment

- chrome-devtools-mcp **1.6.0** (global npm install), Node v24.16.0, Debian 13
- Chrome stable 150.0.7871.186, launched via `--executablePath` with a persistent `--userDataDir`, headful on Xvfb
- One long-lived server process shared by multiple MCP clients through a local stdio proxy; typically 10–30 concurrent MCP sessions
- Target pages are heavy internal SPAs (dashboards); tools used most: `evaluate_script`, `take_screenshot`, `navigate_page`, `take_snapshot`
- `--max-old-space-size=1536` on the server process

### Symptom

Recurring fatal OOMs of the MCP process under sustained use:

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
... v8::internal::Runtime_AllocateInOldGeneration
```

Under heavy multi-session activity the process dies roughly every ~3 hours (6 OOMs in a single day at peak usage).

### Relationship to #1192

I am aware #1192 was closed as completed. I verified the devtools-frontend fix is present in the 1.6.0 bundle (`isPrimaryFrame()` in `build/src/third_party/index.js` has the corrected form). The behavior below persists on 1.6.0, so either this is a different retention path of the same class, or the fix does not cover this usage pattern (no `--autoConnect` involved).

### Instrumentation and data

I injected a sampler into the process via `NODE_OPTIONS --import` that logs `v8.getHeapStatistics()` + old_space every 30 s per isolate (main and worker threads tagged separately), plus a per-tool-call log for correlation. Main-isolate old_space over one afternoon cycle (UTC):

| time | old_space (MiB) | note |
|---|---|---|
| 18:48 | 27 | fresh start |
| 19:18 | 209 | active load |
| 19:48 | 370 | active load |
| 20:18 | 573 | active load |
| 20:48–21:18 | ~565 | **idle period — no growth** |
| 21:48 | 727 | load resumes |
| 22:07 | **1296** | +569 MiB in 22 min |

Key observations:

1. Growth is **strictly correlated with activity** and fully plateaus when idle — consistent with per-event/per-request data being retained, not a timer-based leak.
[...]

PR #2435, opened 2026-07-28
fix: bound per-navigation network request retention

mvanhorn:
Add an explicit per-navigation request limit to `NetworkCollector` and evict the oldest entries as new requests arrive, using a production default of 1,000 retained requests while preserving request order, stable IDs for retained entries, and the existing three-navigation policy. `NetworkCollector` limits preserved history to three navigation buckets, but each bucket—especially the current navigation—can retain an unlimited number of Puppeteer `HTTPRequest` objects.

Emit more than 1,000 requests on one page without navigating and verify that only the newest 1,000 remain, in arrival order, with retained stable IDs still resolving; Cross several main-frame navigations, including redirect navigation requests and subframe events, and verify that each of the three retained navigation buckets is independently bounded without losing the current navigation request.

Refs #2431

### DIFF
--- a/src/McpPage.ts
+++ b/src/McpPage.ts
@@ -159,10 +159,7 @@ export class McpPage implements ContextPage {
     };
     page.on('dialog', this.#dialogHandler);
 
-    this.networkCollector = new NetworkCollector(
-      page,
-      NetworkCollector.DEFAULT_MAX_REQUESTS_PER_NAVIGATION,
-    );
+    this.networkCollector = new NetworkCollector(page);
     this.consoleCollector = new ConsoleCollector(page, collect => {
       return {
         console: event => {
@@ -853,7 +850,7 @@ export class McpPage implements ContextPage {
     this.networkCollector.dispose();
     this.networkCollector = new NetworkCollector(
       this.pptrPage,
-      NetworkCollector.DEFAULT_MAX_REQUESTS_PER_NAVIGATION,
+      undefined,
       collect => {
         return {
           request: req => {
--- a/src/McpPage.ts
+++ b/src/McpPage.ts
@@ -159,7 +159,10 @@ export class McpPage implements ContextPage {
     };
     page.on('dialog', this.#dialogHandler);
 
-    this.networkCollector = new NetworkCollector(page);
+    this.networkCollector = new NetworkCollector(
+      page,
+      NetworkCollector.DEFAULT_MAX_REQUESTS_PER_NAVIGATION,
+    );
     this.consoleCollector = new ConsoleCollector(page, collect => {
       return {
         console: event => {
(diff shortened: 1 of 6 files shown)


src/tools/network.ts
```diff
@@ -34,7 +34,7 @@ const FILTERABLE_RESOURCE_TYPES: readonly [ResourceType, ...ResourceType[]] = [
 
 export const listNetworkRequests = definePageTool({
   name: 'list_network_requests',
-  description: `List all requests for the currently selected page since the last navigation.`,
+  description: `List the most recent requests for the currently selected page since the last navigation. Up to 1,000 requests are retained per navigation.`,
```
Lightning00Blade: Nit: We don't n
```suggestion
  description: `Lists the most recent requests for the currently selected page since the last navigation.`,
```

src/McpPage.ts
```diff
@@ -157,7 +157,10 @@ export class McpPage implements ContextPage {
     };
     page.on('dialog', this.#dialogHandler);
 
-    this.networkCollector = new NetworkCollector(page);
+    this.networkCollector = new NetworkCollector(
+      page,
+      NetworkCollector.DEFAULT_MAX_REQUESTS_PER_NAVIGATION,
```
Lightning00Blade: This is handled by the default, so we can remove.

src/McpPage.ts
```diff
@@ -849,15 +852,19 @@ export class McpPage implements ContextPage {
...
-          if (req.url().includes('favicon.ico')) {
-            return;
-          }
-          collect(req);
-        },
-      } as ListenerMap;
-    });
+    this.networkCollector = new NetworkCollector(
+      this.pptrPage,
+      NetworkCollector.DEFAULT_MAX_REQUESTS_PER_NAVIGATION,
```
Lightning00Blade: We can pass `undefined` to make sure the default it handled correctly.

src/PageCollector.ts
```diff
@@ -283,8 +293,11 @@ class PageEventSubscriber {
 }
 
 export class NetworkCollector extends PageCollector<HTTPRequest> {
+  static readonly DEFAULT_MAX_REQUESTS_PER_NAVIGATION = 1_000;
```
Lightning00Blade: Nit: Given that it is const we can omit the default.
```suggestion
  static readonly MAX_REQUESTS_PER_NAVIGATION = 1_000;
```

Lightning00Blade (approved):
LGTM, with some small comments.

mvanhorn:
All four are in 4067b70. Took both suggestions as written, removed the explicit default in the constructor call, and passed `undefined` at the reassignment site so the default path gets exercised. The rename meant touching the two references in `tests/PageCollector.test.ts` as well.

Lightning00Blade:
@mvanhorn Please rebase an fix the issue (`npm run gen`).

PR #2443, opened 2026-07-30
refactor: introduce Explicit resouce managent (using)

### LINKED ISSUE #2440: drag leaks the source ElementHandle when the target uid lookup fails
### Description of the bug

The `drag` tool resolves both element handles *before* its `try/finally`:

```ts
// src/tools/input.ts, drag handler
const fromHandle = await request.page.getElementByUid(request.params.from_uid);
const toHandle   = await request.page.getElementByUid(request.params.to_uid); // can throw
try {
  ...drag...
} finally {
  void fromHandle.dispose();
  void toHandle.dispose();
}
```

If `getElementByUid(to_uid)` throws — which it does for a stale or invalid `to_uid` (`Element uid "…" no longer exists on the page.`, `Element uid "…" not found`, or `No snapshot found`) — the `try` block is never entered, so the already-acquired `fromHandle` is never disposed. `getElementByUid` mints a fresh `ElementHandle` per call, so each failed `drag` leaks one remote object for the life of the page's execution context.

This is the same class of leak as the recently-fixed #2422 (`take_screenshot`). Among the multi-handle input tools, `fill_form` (via `fillFormElement`) already wraps each handle in `try/finally`, so `drag` is the only one affected.

### Reproduction

1. Take a snapshot to get element uids.
2. Call `drag` with a valid `from_uid` and a `to_uid` that no longer resolves (e.g. the DOM changed since the snapshot, or a bad uid).
3. `getElementByUid(to_uid)` throws; the call fails and `fromHandle` is leaked. Repeat → remote objects accumulate until page teardown.

### Expectation

The source handle is disposed even when the target lookup fails, matching the other input handlers.

Fix: acquire `toHandle` inside a nested `try` so `fromHandle` is disposed on every path. I have this plus a regression test (spy `fromHandle.dispose`, force the target lookup to throw, assert dispose was still called) passing locally and can open a PR.

main @ 574c320 (1.6.0)

### Chrome version

Lightning00Blade:
Re-export the Puppeteer helpers as we have tested those.
This should also be a small performance improvement as we don't await any of the disposals.
Will also fix memory leaking for some methods.
With the added benefit of make the code look cleaner (no try/catch.)

Also Fixes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2440, Closes https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2441

### DIFF
--- a/src/third_party/index.ts
+++ b/src/third_party/index.ts
@@ -51,6 +51,10 @@ export type {Realm} from 'puppeteer-core/internal/api/Realm.js';
 export type {JSONSchema7, JSONSchema7Definition} from 'json-schema';
 export {Mutex} from 'puppeteer-core/internal/util/Mutex.js';
 export {
+  DisposableStack,
+  AsyncDisposableStack,
+} from 'puppeteer-core/internal/util/disposable.js';
+export {
   resolveDefaultUserDataDir,
   detectBrowserPlatform,
   Browser as BrowserEnum,
--- a/src/tools/input.ts
+++ b/src/tools/input.ts
@@ -44,7 +44,7 @@ function handleActionError(error: unknown, uid: string) {
 }
 
 async function selectNativeSelectOption(handle: ElementHandle<Element>) {
-  const selectHandle = await handle.evaluateHandle(node => {
+  using selectHandle = await handle.evaluateHandle(node => {
     if (!(node instanceof HTMLOptionElement)) {
       return null;
     }
@@ -64,26 +64,22 @@ async function selectNativeSelectOption(handle: ElementHandle<Element>) {
 
     return select;
   });
-  try {
-    const select = selectHandle.asElement() as ElementHandle<Element> | null;
+
+    using select = selectHandle.asElement() as ElementHandle<Element> | null;
     if (!select) {
       return false;
     }
 
-    const valueHandle = await handle.getProperty('value');
-    try {
+    using valueHandle = await handle.getProperty('value');
+
       const value = await valueHandle.jsonValue();
       if (typeof value !== 'string') {
         return false;
       }
       await select.asLocator().fill(value);
-    } finally {
-      void valueHandle.dispose();
-    }
+
     return true;
-  } finally {
-    void selectHandle.dispose();
-  }
+
 }
 
 export const click = definePageTool({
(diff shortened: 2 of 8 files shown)


src/tools/input.ts
```diff
@@ -225,22 +216,16 @@ async function selectOption(
...
-            if (childValue) {
-              await handle.asLocator().fill(childValue.toString());
-            }
-          } finally {
-            void childValueHandle.dispose();
-          }
-          break;
-        } finally {
-          void childHandle.dispose();
+        const childValueHandle = await childHandle.getProperty('value');
```
Lightning00Blade: Should be using.

OrKoN:
Let's add a dummy test that the behavior of using is as expected in this repo (just in case something does not work in some versions)?

Lightning00Blade:
@OrKoN Can you expand on what the test should do? I think the Stacks are extensively tested in Puppeteer.
And the using part handled by TypeScript.

Lightning00Blade:
Note, if you forget to import the Stacks from `third_party` the test will fail for NodeJS v22 which does not support it out of the box.

OrKoN:
> @OrKoN Can you expand on what the test should do? I think the Stacks are extensively tested in Puppeteer. And the using part handled by TypeScript.

Test that it actually works as expected (e.g., disposes things when going out of scope). Similar to what we do in Puppeteer but just for this repo setup (node matrix / compilation config).

src/third_party/index.ts
```diff
@@ -49,6 +49,10 @@ export type {CdpPage} from 'puppeteer-core/internal/cdp/Page.js';
 export type {CdpWebWorker} from 'puppeteer-core/internal/cdp/WebWorker.js';
 export type {Realm} from 'puppeteer-core/internal/api/Realm.js';
 export type {JSONSchema7, JSONSchema7Definition} from 'json-schema';
+export {
+  DisposableStack,
+  AsyncDisposableStack,
+} from 'puppeteer-core/internal/util/disposable.js';
```
OrKoN, line 56: should we import it via polyfills.ts?

OrKoN: never mind, we actually need it, I misread.

Lightning00Blade: Yeah Puppeteer does not pollute the global scope with this two only with the Global symbols.

eslint.config.mjs
```diff
@@ -103,17 +103,18 @@ export default defineConfig([
...
-          'newlines-between': 'always',
-
-          alphabetize: {
-            order: 'asc',
-            caseInsensitive: true,
-          },
-        },
-      ],
+      // Incompatible with ESLint 10
+      // 'import/order': [
```
OrKoN, line 107: is it from some other PR?

Lightning00Blade: I will split it but this is broken now with the move to EsLint v10

OrKoN (approved):

PR #2445, opened 2026-07-30
feat(telemetry): log is_devtools_open with each tool call.

yulunz:


src/ToolHandler.ts
```diff
@@ -303,12 +326,17 @@ export class ToolHandler {
         isError: true,
       };
     } finally {
+      const devToolsData = await getDevToolsStatus(page, context);
```
OrKoN: I think we should fetch DevTools data in the try block where tool calls are handled as it still might throw. Also since we are putting it on the hot path, let's add a timer in case getting data takes long time.

src/ToolHandler.ts
```diff
@@ -143,6 +148,22 @@ function buildUnknownArgumentsMessage(
   return `Unknown ${unknownLabel} for tool "${toolName}": ${formatArgumentNames(unknownArgumentNames)}. ${expectedArguments} ${correction} and retry.`;
 }
 
+async function getDevToolsStatus(
+  page?: McpPage,
+  context?: McpContext,
```
OrKoN: what are the situations that there is no McpContext? let's make this a getter on  McpContext: `context.getDevToolsData(page)` instead?

src/ToolHandler.ts
OrKoN: Scratch the part about it throwing; I saw we catch the error within the function. Nonetheless let's move the data fetching next to the tool handling logic.

src/ToolHandler.ts
yulunz: done.

src/ToolHandler.ts
yulunz: done.

OrKoN (approved):

PR #2447, opened 2026-07-30
fix: improve file writing

OrKoN:


Lightning00Blade (approved):

OrKoN:
Added tests

tests/McpContext.test.ts
```diff
@@ -485,6 +485,105 @@ describe('McpContext', () => {
     });
   });
 
+  describe('symlink security checks', () => {
+    it('saveFile refuses to write through a symlink to an existing file', async () => {
+      if (os.platform() === 'win32') {
```
Lightning00Blade: Can we add a comment why we skip it on Windows?

Lightning00Blade: If it's for every test, we can add it to the describe.

Lightning00Blade (approved):

PR #2449, opened 2026-07-30
fix(memory): dispose the heap-snapshot worker when loading fails

### LINKED ISSUE #2448: Heap-snapshot worker is leaked when loading a snapshot fails (missing/invalid file)
### Description of the bug

`HeapSnapshotManager.#loadSnapshot` creates a `HeapSnapshotWorkerProxy` (a live worker) *before* it reads the snapshot file, but the manager only ever disposes workers that are tracked in the `#snapshots` map:

```ts
// src/HeapSnapshotManager.ts
async #loadSnapshot(absolutePath, uid) {
  const workerProxy = new ...HeapSnapshotWorkerProxy(...);   // spawns a worker
  const loaderProxy = workerProxy.createLoader(uid, ...);
  const fileStream = fsSync.createReadStream(absolutePath, {...});
  for await (const chunk of fileStream) {   // throws ENOENT for a missing/unreadable file
    await loaderProxy.write(chunk);
  }
  await loaderProxy.close();
  const snapshot = await snapshotPromise;
  return {snapshot, worker: workerProxy};
}
```

`getSnapshot` has no try/catch and only calls `#snapshots.set(...)` on success. So when a load fails, the worker is never added to `#snapshots`, and therefore neither `dispose(filePath)` nor `disposeAll()` (added in #2428 for context teardown) can ever reach it. The worker leaks for the life of the process — one worker per failed load.

`verifyFilesSchema: ['filePath']` on the read tools only checks the path is inside the workspace roots; it does not verify the file exists or is a valid snapshot.

### Reproduction

1. Call any of `get_heapsnapshot_summary` / `get_heapsnapshot_details` / `get_heapsnapshot_class_nodes` / `get_heapsnapshot_retainers` / `get_heapsnapshot_retaining_paths` / `get_heapsnapshot_edges` with a `filePath` that is inside a workspace root but does not exist (or is a truncated / non-heapsnapshot file).
2. The call fails, and the worker created for it is never disposed.
3. Repeat → workers accumulate.

A direct unit test that fails on `main`:

```ts
const disposeSpy = sinon.spy(
  DevTools.HeapSnapshotModel.HeapSnapshotProxy.HeapSnapshotWorkerProxy.prototype,
  'dispose',
);
const manager = new HeapSnapshotManager();
await assert.rejects(manager.getSnapshot('/nonexistent/does-not-exist.heapsnapshot'));
sinon.assert.calledOnce(disposeSpy); // 0 calls on main
```

### Expectation

A failed snapshot load disposes its worker rather than leaking it.

Fix: wrap the load body in try/catch and `workerProxy.dispose()` before rethrowing. I have this plus the regression test passing locally and can open a PR.

main @ 574c320 (1.6.0)

### Chrome version

ch-bas:
Fixes #2448

## Problem

`HeapSnapshotManager.#loadSnapshot` creates a `HeapSnapshotWorkerProxy` (a live worker) before reading the file, but the manager only disposes workers tracked in the `#snapshots` map. `getSnapshot` adds the entry only on success, so a failed load — e.g. a missing or invalid `.heapsnapshot` path passed to any `get_heapsnapshot_*` tool — throws before the entry exists, and neither `dispose()` nor `disposeAll()` (added in #2428) can reach the worker. It leaks for the life of the process, one per failed load.

## Fix

Wrap the load body in `try/catch` and `workerProxy.dispose()` before rethrowing, so a failed load cleans up its own worker.

## Testing

- Added `tests/HeapSnapshotManager.test.ts`: spies on `HeapSnapshotWorkerProxy.prototype.dispose`, forces a load failure via a nonexistent path, and asserts the worker was disposed. Fails on `main` (`dispose` called 0 times) and passes with this change.
- `npm run test tests/HeapSnapshotManager.test.ts tests/tools/memory.test.ts` — all pass.
- ESLint + Prettier clean.

### DIFF
--- a/src/HeapSnapshotManager.ts
+++ b/src/HeapSnapshotManager.ts
@@ -362,26 +362,35 @@ export class HeapSnapshotManager {
         import.meta.resolve('./third_party/devtools-heap-snapshot-worker.js'),
       );
 
-    const {promise: snapshotPromise, resolve: resolveSnapshot} =
-      Promise.withResolvers<DevTools.HeapSnapshotModel.HeapSnapshotProxy.HeapSnapshotProxy>();
+    try {
+      const {promise: snapshotPromise, resolve: resolveSnapshot} =
+        Promise.withResolvers<DevTools.HeapSnapshotModel.HeapSnapshotProxy.HeapSnapshotProxy>();
 
-    const loaderProxy = workerProxy.createLoader(uid, snapshotProxy => {
-      resolveSnapshot(snapshotProxy);
-    });
+      const loaderProxy = workerProxy.createLoader(uid, snapshotProxy => {
+        resolveSnapshot(snapshotProxy);
+      });
 
-    const fileStream = fsSync.createReadStream(absolutePath, {
-      encoding: 'utf-8',
-      highWaterMark: 1024 * 1024,
-    });
+      const fileStream = fsSync.createReadStream(absolutePath, {
+        encoding: 'utf-8',
+        highWaterMark: 1024 * 1024,
+      });
 
-    for await (const chunk of fileStream) {
-      await loaderProxy.write(chunk);
+      for await (const chunk of fileStream) {
+        await loaderProxy.write(chunk);
+      }
+
+      await loaderProxy.close();
+
+      const snapshot = await snapshotPromise;
+      return {snapshot, worker: workerProxy};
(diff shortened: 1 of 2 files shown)


Lightning00Blade (approved):
Thanks for the PR.

PR #2452, opened 2026-07-31
fix: manage third-party element handle lifetimes

haopengmai:
## Summary

- scope third-party tool input handles to a single execution
- transfer returned element handles to `McpPage` only after a snapshot succeeds
- release retained handles on replacement, main-frame navigation, and page teardown
- return a fresh caller-owned handle for every UID lookup
- keep the page-side element stash invocation-local and clear it on success and failure

## Root cause

`TextSnapshot` stored third-party element handles in snapshot nodes while also mutating `McpPage.extraHandles` implicitly. The explicit resource-management change in #2443 then disposed those handles as soon as snapshot creation completed, so UIDs for elements outside the accessibility tree immediately referenced disposed handles.

The page-side stash also accumulated elements across tool calls, and retained handles were not invalidated on navigation or protected against teardown races.

This change makes ownership explicit:

- tool input handles are temporary
- `McpPage` owns retained output anchors
- `TextSnapshot` only consumes anchors and mints a fresh handle per lookup
- browser-side stashes are temporary handoff storage

## User impact

UIDs returned by third-party developer tools remain reusable across subsequent tool calls, including detached or otherwise non-accessible DOM elements. Old resources are released deterministically instead of leaking or surviving navigation and teardown.

## Tests

- `npm run check-format`
- `PUPPETEER_EXECUTABLE_PATH='/Applications/Google Chrome.app/Contents/MacOS/Google Chrome' npm run test tests/tools/thirdPartyDeveloper.test.ts`
- `PUPPETEER_EXECUTABLE_PATH='/Applications/Google Chrome.app/Contents/MacOS/Google Chrome' npm run test tests/McpPage.test.ts tests/McpContext.test.ts tests/PageCollector.test.ts tests/TextSnapshot.test.ts tests/tools/input.test.ts tests/tools/script.test.ts tests/tools/snapshot.test.ts tests/tools/thirdPartyDeveloper.test.ts`

### DIFF
--- a/src/McpPage.ts
+++ b/src/McpPage.ts
@@ -77,6 +77,7 @@ import {
   type WebMCPTool,
   type Protocol,
   type Page,
+  type Frame,
   type ConsoleMessage,
   type HTTPRequest,
   type DevTools,
@@ -103,6 +104,7 @@ import {
   type WaitForEventsResult,
   type DialogAction,
 } from './WaitForHelper.js';
+type DisposableStackInstance = InstanceType<typeof DisposableStack>;
 
 /**
  * Per-page state wrapper. Consolidates dialog, snapshot, emulation,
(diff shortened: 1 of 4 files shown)


OrKoN:
@Lightning00Blade could be a regression from the `using` PR.

PR #2455, opened 2026-07-31
feat(telemetry): report whether tool call is made on localhost.

yulunz:


src/ToolHandler.ts
```diff
@@ -249,6 +250,10 @@ export class ToolHandler {
           if (this.tool.blockedByDialog) {
             page.throwIfDialogOpen();
           }
+          const targetPage = page ?? context.getSelectedMcpPage();
```
OrKoN: can page be optional here?

yulunz: move it out of the if-else clause so page is undefined when the tool is not page scoped.

OrKoN (approved):

PR #2457, opened 2026-08-01
fix: resolve dangling symlinks to their target in resolveCanonicalPath

evilgensec:
### Problem

`fs.realpath()` reports `ENOENT` in two different situations: the path does not exist, and the path is a symlink whose target does not exist. `resolveCanonicalPath` treats both the same way, so for a dangling symlink it falls through to the nearest existing ancestor walk and returns

```
path.join(canonicalParent, path.basename(current), ...missingSegments)
```

which is the link's own location, not where the link points.

`McpContext.validatePath` uses that value to decide whether a client supplied path lies inside a configured workspace root:

```ts
canonicalPath === canonicalRoot ||
canonicalPath.startsWith(canonicalRoot + path.sep)
```

So a dangling symlink sitting inside a root resolves to a path inside the root and passes containment, even though a write to it lands at the link target. A live symlink pointing outside the root is handled correctly today, because `realpath()` succeeds and containment then fails. Only the dangling case slips through. The same applies when the dangling symlink is an intermediate directory component.

Measured against the current resolver, using the containment expression above:

| Input | resolved to | passes containment |
| :-- | :-- | :-- |
| new file in root | inside root | yes, correct |
| symlink to an existing file outside root | the outside target | no, correct |
| symlink to a non-existent file outside root | the link itself, inside root | **yes, incorrect** |
| dangling symlink as a parent directory | the link path, inside root | **yes, incorrect** |

Most write paths are not affected in practice, because `McpContext.#writeFile` passes `O_NOFOLLOW` and fails with `ELOOP` on a symlink. The gap matters for writers that do not, for example `page.screencast()`, which hands the path to ffmpeg.

### Change

When `realpath()` fails with `ENOENT`, `lstat` the path first. If it is a symlink, read the link and resolve the target instead of treating the path as missing. The ancestor walk now resolves the parent through `resolveCanonicalPath` as well, so a dangling symlink used as an intermediate directory is followed too.

Recursion is bounded by `MAX_SYMLINK_DEPTH`, set to the conventional `SYMLOOP_MAX` of 40, and a self referential link still rejects rather than looping.

### Tests

Three cases added to `tests/utils/files.test.ts`:

* a dangling symlink resolves to its target rather than to the link
* a dangling symlink used as a parent directory resolves through to the target
* a pair of self referential dangling links rejects instead of looping

The first two fail on the current resolver and pass with this change. The four existing `resolveCanonicalPath` tests continue to pass, so behaviour for existing files, missing files with existing parents, deeply nested missing paths, and live symlinks is unchanged.

### DIFF
--- a/src/utils/files.ts
+++ b/src/utils/files.ts
@@ -15,18 +15,52 @@ export async function getTempFilePath(filename: string) {
   return filepath;
 }
 
-export async function resolveCanonicalPath(filePath: string): Promise<string> {
+function isENOENT(err: unknown): boolean {
+  return (
+    !!err && typeof err === 'object' && 'code' in err && err.code === 'ENOENT'
+  );
+}
+
+// Matches the conventional SYMLOOP_MAX so a cycle of dangling links cannot
+// recurse without bound.
+const MAX_SYMLINK_DEPTH = 40;
+
+export async function resolveCanonicalPath(
+  filePath: string,
+  depth = 0,
+): Promise<string> {
   const absolutePath = path.resolve(filePath);
   try {
     // Get the true canonical path, resolving all symlinks.
     return await fs.realpath(absolutePath);
   } catch (err) {
-    if (
-      err &&
-      typeof err === 'object' &&
-      'code' in err &&
-      err.code === 'ENOENT'
-    ) {
+    if (isENOENT(err)) {
+      // realpath() also reports ENOENT for a symlink whose target does not
+      // exist. Treating that as a missing path would fall through to the
+      // ancestor walk below and return the link's own location, which hides
+      // the fact that writing to it lands wherever the link points. Follow the
+      // link explicitly instead, so the caller sees the real destination.
+      let linkStat;
+      try {
+        linkStat = await fs.lstat(absolutePath);
+      } catch {
+        linkStat = undefined;
+      }
+      if (linkStat?.isSymbolicLink()) {
+        if (depth >= MAX_SYMLINK_DEPTH) {
+          throw err;
+        }
+        const target = await fs.readlink(absolutePath);
+        const resolvedTarget = path.resolve(
+          path.dirname(absolutePath),
+          target,
+        );
+        if (resolvedTarget === absolutePath) {
+          throw err;
+        }
+        return await resolveCanonicalPath(resolvedTarget, depth + 1);
+      }
+
       // Find the nearest existing ancestor directory on the filesystem.
       let current = absolutePath;
       const missingSegments: string[] = [];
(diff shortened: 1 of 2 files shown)


evilgensec:
Rebased onto main and reformatted — the two `[Required]` checks that were red were both a single unformatted line, now fixed.

@OrKoN this has been open since 1 Aug without a review. It's one file, `src/utils/files.ts`.

`fs.realpath()` returns `ENOENT` for two different situations: the path genuinely does not exist, and the path is a symlink whose target does not exist. `resolveCanonicalPath` treats them the same, so for a dangling symlink it falls back to resolving the parent directory and appending the basename — returning a path that still contains the unresolved link. A caller that then uses that path for a containment check gets the wrong answer, because the link can point anywhere.

The change reads the link with `fs.readlink()` on `ENOENT` and resolves the target relative to the link's directory, with a guard for the self-referential case. Tests cover a dangling symlink, a dangling symlink as a parent directory, and a self-referential link; the first two fail before the change.

One question rather than a bump: is resolving a dangling symlink to its intended target the behaviour you want here, or would you rather `resolveCanonicalPath` reject dangling links outright? Both close the containment gap and I'm happy to switch it.

OrKoN:
Thanks for the ping and the PR! I think the right way to fix these issues is to have no-follow-symlinks policy when writing files in Puppeteer which I already started working on. I think any fixes here are not solving the problem fully, because by the time the write happens the symlinks and files on the system can be replaced to point elsewhere. Related https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2572

evilgensec:
Makes sense. No-follow at `open()` is the right layer for this, and it handles the live-swap race a resolver fix can't. I'll close this in favor of #2572 plus the puppeteer setting.

One loose end before I do: I don't see `setFollowSymlinks(false)` called anywhere in this repo yet, so the resolved path from #2572 still goes to puppeteer writers like screencast and screenshot, which follow symlinks by default. If it's useful I can repurpose this PR to just flip that setting on at startup once puppeteer ships it. Otherwise I'll close.

OrKoN:
> One loose end before I do: I don't see setFollowSymlinks(false) called anywhere in this repo yet, so the resolved path from https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2572 still goes to puppeteer writers like screencast and screenshot, which follow symlinks by default. If it's useful I can repurpose this PR to just flip that setting on at startup once puppeteer ships it. Otherwise I'll close.

we are still working on rolling puppeteer. Once it's done we will turn on the no-follow mode. I think there is no need for you to update the PR and it can be closed.

OrKoN:
here is the PR https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2592 turning on the no follow mode.

PR #2461, opened 2026-08-03
fix(cli): warn about version mismatch between cli and daemon

zyzyzyryxy:


src/bin/chrome-devtools.ts
```diff
@@ -264,6 +270,8 @@ for (const [commandName, commandDef] of Object.entries(commands)) {
       try {
         if (!isDaemonRunning(sessionId)) {
           await start(serializeArgs(cliOptions, argv), sessionId);
+        } else {
+          await verifyDaemonVersion(sessionId, VERSION);
```
OrKoN: can we cache the result or move it some other stage? I forgot if daemon supports concurrent commands but we could run it in parallel to the command and print a warning if any in the end of the command?

src/daemon/client.ts
```diff
@@ -194,6 +194,34 @@ export async function stopDaemon(sessionId: string) {
...
+  sessionId: string,
+  cliVersion: string,
+): Promise<void> {
+  if (!isDaemonRunning(sessionId)) {
+    return;
+  }
+  try {
+    const response = await sendCommand({method: 'status'}, sessionId);
+    if (response.success) {
+      const data: unknown = JSON.parse(response.result);
```
OrKoN: could we use types defined in the daemon.ts?

tests/daemon/client.test.ts
```diff
@@ -73,6 +75,74 @@ describe('daemon client', () => {
     });
   });
 
+  describe('verifyDaemonVersion', () => {
+    let sessionId: string;
+
+    beforeEach(async () => {
```
OrKoN: let's re-use hooks in the start/stop tests by lifting them up to the 'daemon client' describe block

src/daemon/client.ts
zyzyzyryxy: Sure, extracted `DaemonStatusResult`

tests/daemon/client.test.ts
zyzyzyryxy: Done

src/bin/chrome-devtools.ts
zyzyzyryxy: Looks like we can safely call status concurrently. Updated.

OrKoN (approved):

PR #2462, opened 2026-08-03
fix: regression after the #2443

Lightning00Blade:
This also includes a fix memory leak for the stashedElements.

### DIFF
--- a/src/McpPage.ts
+++ b/src/McpPage.ts
@@ -472,9 +472,6 @@ export class McpPage implements ContextPage {
           throw new Error('No tools found on the page');
         }
 
-        // Clear previously stashed elements so we only stash new ones from this execution
-        window.__dtmcp.stashedElements = [];
-
         const toolResult = await window.__dtmcp.executeTool(name, args);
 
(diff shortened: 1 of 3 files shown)


src/McpPage.ts
```diff
@@ -471,6 +471,10 @@ export class McpPage implements ContextPage {
         if (!window.__dtmcp?.executeTool) {
           throw new Error('No tools found on the page');
         }
+
+        // Clear previously stashed elements so we only stash new ones from this execution
+        window.__dtmcp.stashedElements = [];
```
wolfib: How about 
```
if window.__dtmcp?.stashedElements {
  window.__dtmcp.stachedElements = [];
}
```
?

wolfib: Oh, this can be reset even earlier. There is no need to wait for the next call of `executeThirdPartyDeveloperTool` before clearing this stash. This can be done immediately after they've been used in lines 556-566.

wolfib (approved):
LGTM

PR #2463, opened 2026-08-03
fix: load source maps on demand instead of for every parsed script

Dmitriusan:
Replaces the original contents of this PR. That version made the DevTools universe lazy, which @OrKoN pointed out collides with #1777 and, as it turns out, treated the symptom anyway: the cost is per page whether or not the universe is created eagerly. This version fixes the cost itself and has no bearing on how many pages get a universe.

Root cause and measurements are in https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2431#issuecomment-5170268714.

### Problem

`DebuggerModel.parsedScriptSource()` calls `attachSourceMap()` for every script that carries a `sourceMapURL`, because `SourceMapManager` defaults to enabled. Each map is downloaded, VLQ decoded, sorted, reverse mapped, and the decoded mappings are retained for as long as the model lives. On one Jira issue page that is 147 maps and 2427 MB retained, measured with a full GC forced every second. `createPagesSnapshot()` builds an `McpPage` for every page in the browser, so this is paid per open tab.

### Change

The universe only needs source maps to symbolicate the frames of a stack trace it is about to format, and `createStackTrace()` already walks exactly those scripts. So:

- Developer resource loads are refused by default. Both channels have to be gated: `PageResourceLoader.dispatchLoad()` falls back from the CDP loader to the host bindings, so gating one alone reroutes the download instead of preventing it. That is `PuppeteerDevToolsConnection.send` and `McpHostBindingAdapter.loadNetworkResource`.
- `createStackTrace()` allows the URL for a script it needs, then does `detachSourceMap()` + `attachSourceMap()` on that one script, which makes the manager run the load it skipped.
- It then awaits `sourceMap.waitForScopeInfo()`. Without that the frames resolve to positions but the function names come back empty, which is easy to miss.
- A map that already failed to load is not retried.

`SourceMapManager.setEnabled(false)` is not usable for this: `setEnabled(true)` re-attaches every registered client at once, so toggling it around a symbolication brings the whole avalanche back.

### Results

Fixture is a rollup bundle with a real source map back to `src/app.js`; the large one has a 2.8 MB map. Memory is measured against a real browser with 18 ordinary tabs open (Teams, Jira, Gmail, a few SPAs), one `list_pages` call and then no tool calls at all.

| | `main` | this branch |
| --- | --- | --- |
| RSS | OOM at 4.9 GB in 65 s | 210.1 -> 210.3 MB over 120 s |
| stack trace | `deliberatelyFail (app.js:2:9)` … | byte-identical |
| `get_console_message`, 2.8 MB map, 3 runs | 1893 / 1882 / 2340 ms | 133 / 139 / 142 ms |
| `npm run test` | baseline | no new failures |

Two things worth calling out. The memory result holds with every one of the 18 pages still getting a universe, so this does not depend on the lazy-universe idea at all. And symbolication came out roughly 14x faster rather than slower, which I did not expect; I have not dug into why the eager path is slow, so treat that number as an observation rather than a claim about the mechanism.

### Notes for review

- The allowlist in `sourceMapGate.ts` is module state. Per-context would be tidier if you would rather have it on `McpContext`, I kept it standalone so the two gate points do not need a context reference.
- I first put the fallback gate in `McpContext.loadResource` and it broke eight tests, correctly: that method is a public API with its own allowlist, blocklist, `file:` and path-validation semantics. It lives in the host binding adapter instead.
- Inline `data:` source maps never reach the loader and are unaffected.
- Separately, and not fixed here: detaching a CDP session while a source map is loading throws an uncaught `Error: Unknown session <id>` out of `PuppeteerDevToolsConnection.send` and kills the process. `send()` is declared to return a promise but throws synchronously, so the generated agent code never catches it. Happy to send that as its own PR.

### DIFF
--- a/src/devtools/DevToolsConnectionAdapter.ts
+++ b/src/devtools/DevToolsConnectionAdapter.ts
@@ -8,6 +8,11 @@ import type * as puppeteer from '../third_party/index.js';
 import type {DevTools} from '../third_party/index.js';
 import {CDPSessionEvent} from '../third_party/index.js';
 
+import {
+  isResourceUrlAllowed,
+  recordRefusedResourceUrl,
+} from './sourceMapGate.js';
+
 /**
  * This class makes a puppeteer connection look like DevTools CDPConnection.
  *
@@ -57,6 +62,22 @@ export class PuppeteerDevToolsConnection
     const session = this.#connection.session(sessionId);
     if (!session) {
       throw new Error('Unknown session ' + sessionId);
+    }
+    // The DevTools frontend loads developer resources (in practice: source
+    // maps) for every script it sees. We only want the ones we are about to
+    // use, so refuse the rest before they reach the browser. Answering with a
+    // protocol error is the path `PageResourceLoader` already handles.
+    if (method === 'Network.loadNetworkResource') {
+      const url = (params as {url?: string})?.url;
+      if (url !== undefined && !isResourceUrlAllowed(url)) {
+        recordRefusedResourceUrl(url);
+        return Promise.resolve({
+          error: {
+            code: -32000,
+            message: 'Developer resources are loaded on demand',
+          } as DevTools.CDPConnection.CDPError,
+        });
+      }
     }
     // Rolled protocol version between puppeteer and DevTools doesn't necessarily match
     /* eslint-disable @typescript-eslint/no-explicit-any */
--- a/src/devtools/DevtoolsUtils.ts
+++ b/src/devtools/DevtoolsUtils.ts
@@ -13,6 +13,8 @@ import type {
 
 import {PuppeteerDevToolsConnection} from './DevToolsConnectionAdapter.js';
 import {McpHostBindingAdapter} from './McpHostBindingAdapter.js';
+import {allowResourceUrl} from './sourceMapGate.js';
+import {logger} from '../utils/logger.js';
 
 /**
  * A mock implementation of an issues manager that only implements the methods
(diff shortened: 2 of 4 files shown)


src/McpContext.ts
```diff
@@ -421,6 +421,10 @@ export class McpContext implements Context {
   selectPage(newPage: McpPage): void {
     this.#selectedPage = newPage;
     newPage.updateTimeouts();
+    // Start the DevTools universe for the page we are about to work with, so
+    // that stack traces can be symbolicated by the time a response is
+    // formatted. Pages that are never selected never pay for it.
+    void newPage.ensureDevToolsUniverse();
```
OrKoN: with https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1777 we plan to allow agents to concurrently work on different pages, so I think we would need to find a different solution cc @yulunz

Dmitriusan: @OrKoN fyi I've changed the approach, so I hope this issue now has a chance to be resolved

szuend:
I'd like to push back on the approach taken in this PR. Sniffing on the CDP traffic and then massaging certain results opens pandoras' box. Most of the test coverage of DevTools/Universe related code lives in the devtools-frontend repository and this PR changes fundamental behaviors that have zero test coverage.

I agree that the MCP server needs lazy loading of source maps (in contrast to DevTools, which needs them eagerly), but the right solution is to fix this upstream by adding a config option to the `Universe` constructor and wire that to the appropriate places.

Lightning00Blade:
@yulunz with https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2486 can we close this one?

yulunz:
lazy loading is implemented on the upstream and the mcp is lazy loading by default now. Thanks for the suggestion! @Dmitriusan

Issue #2456, opened 2026-08-01
RSS grows unbounded in --browserUrl mode while live heap stays flat (V8 never returns freed pages)


theteam123:
### Summary

In `--browserUrl` attach mode the server's RSS ratchets to 2GB+ within a single session and never comes back down, **while the live heap stays flat**. Nothing is being retained; GC reclaims it fine. The memory is pages V8 committed to hold the peak working set of a heavy page load and never returned to the OS, and with a heap limit sized from total system RAM there is nothing to make it do so.

I'm filing separately from #1192 / #1214 because those diagnose `--autoConnect` reconnection orphans in `PageCollector`, and this setup never reconnects. #1192 is also closed. The symptom matches; the cause appears not to.

### Environment

- chrome-devtools-mcp **1.6.0**
- Node **v22.18.0** (also reproduced on v22.22.3)
- Chrome for Testing **151.0.7922.47**, `--headless=new`
- Linux x86_64, 16GB RAM, **no swap**, 4 vCPU
- `--browserUrl http://127.0.0.1:<port>` against one shared long-lived Chrome, with several server instances attached to it (one per agent session)

### Symptom

A server starts at ~130MB. Ordinary UI-verification work against a heavy single-page app takes it to **1959MB after ten rounds and still climbing**. It never releases. With 6-16 concurrent sessions this exhausts the host: on one occasion five instances sat at 1.1-2.7GB simultaneously and took the machine to 15/15GB used, zero swap, load average 39.

One round = `new_page` / `navigate_page` / `take_snapshot` / `take_screenshot` / `list_network_requests`.

### The measurement that identifies the cause

Rather than infer from RSS, I attached to the server's own Node inspector, forced a full mark-compact with `HeapProfiler.collectGarbage`, and read what survived. Representative run (navigation only, heavy SPA):

| round | RSS | live heap after forced GC | external |
|---:|---:|---:|---:|
| 0 | 371 MB | 179 MB | 12 MB |
| 7 | 1308 MB | 183 MB | 11 MB |

The live heap does not move. External/ArrayBuffer memory does not move. RSS grows ~1GB. The objects are collected; the pages holding them are not returned.

Contributing factor: `v8.getHeapStatistics().heap_size_limit` is **4144MB** here, sized from system RAM. Each server independently believes it may use ~4.1GB, so N concurrent servers hold an N x 4.1GB collective budget against a 16GB host. No individual process ever feels pressure, so the host fails first.

[...]

PR #2464, opened 2026-08-03
chore: add a script to capture memory snapshots

OrKoN:
I did not catch a leak using the test sites.
 
 Refs: https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2456 https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2431

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -149,3 +149,5 @@ build/
 log.txt
 
 .DS_Store
+
+mcp-heap-profiles/
--- a/scripts/profile_mcp.ts
+++ b/scripts/profile_mcp.ts
@@ -5,6 +5,7 @@
  */
 
 import fs from 'node:fs';
+import http from 'node:http';
 import path from 'node:path';
 import {parseArgs} from 'node:util';
 
@@ -23,7 +24,13 @@ interface Measurement extends HeapUsage {
 
 interface ProfileOptions {
   iterations: number;
+  sourceMapScriptCount: number;
   warmupIterations: number;
+}
+
+interface SourceMapTestServer {
+  close(): Promise<void>;
+  url: string;
 }
 
 interface PendingRequest {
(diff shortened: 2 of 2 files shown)


OrKoN:
Using our own memory tools:

```
This run still does not show an accumulating memory leak.

  DevTools comparison found:

  - Total snapshot growth: 267.7 KiB.
  - Compiled code: +236 KiB, about 88% of the increase.
  - Strings and object shapes: +14 KiB.
  - Workload objects—requests, responses, AX nodes, scripts, snapshots, buffers, maps, listeners, and execution contexts—had zero net growth.

  The only suspicious cluster was:

  - 5 Timeout objects
  - 15 promises
  - 15 closure contexts

  Their retaining paths lead to active one-second timers. These match fetchPageTitle() in src/McpResponse.ts:1419, whose Promise.race() does not cancel its timeout when page.title() finishes first:

  return Promise.race([
    page.title().catch(() => ''),
    new Promise<string>(resolve => setTimeout(() => resolve(''), 1000)),
  ]);

  These are transient, not leaked. The baseline was captured after the slow primer snapshot, giving its timers time to expire. The final snapshot was captured immediately after the last iteration, so five recent timers were still alive.
  Each retained only about 0.7 KiB and will disappear after one second.

  Conclusion: the reported 0.24 MiB is continued V8 compilation plus a baseline/final timing mismatch. No accumulating MCP request, page, network, screenshot, or accessibility data was found.

  The profiler should add a settling delay of slightly over one second before both baseline and final measurements. Independently, fetchPageTitle() could clear its timeout after page.title() resolves to avoid retaining short-lived
  promises and contexts. No files were changed during this diagnostic.
  ```

OrKoN:
Adding a page with source maps also did not reproduce.

OrKoN:
I think the page churn scenario (3rd commit) might be reproducing a leak.

yulunz (approved):

PR #2465, opened 2026-08-03
chore: re-use connection adapter and dispose devtools unvierse

OrKoN:
Pending https://chromium-review.git.corp.google.com/c/devtools/devtools-frontend/+/8210805 to fix the leaks

PTAL @yulunz @szuend 

Refs https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2456 https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2431


Dmitriusan:
btw, I've tried this patch, yet I still see the same OOM on my workloads as mentioned in #2431

Numbers from that run, same browser, 18 tabs, one `list_pages` call and then idle, `--max-old-space-size=3072`:

```
main            OOM
this branch     191 MB -> 4304 MB at 18 s -> gone by 48 s
#2463           flat, 647 MB over 75 s
both merged     flat, 655 MB over 90 s
```

So the two look complementary rather than competing.

szuend:
The `DevToolsConnectionAdapter` lives also upstream [1]. Maybe we can make the changes there as well and on the next roll remove the copy on the MCP side?

[1]: https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/core/protocol_client/PuppeteerDevToolsConnection.ts;drc=77c240f90e64b5f6915761276a5288fdb584cbfb

In general, I think we should also add a `dispose` onto the main `Universe` type and add memory fixes like this there. E.g. now the `Universe` dispose should call the `IsolatedFileSystemManager.dispose` which in turn unregisters host binding events. Cleaning this up externally feels rather brittle.

OrKoN:
@yulunz @szuend this is ready for review.

yulunz (approved):

PR #2466, opened 2026-08-03
fix: do not throw synchronously when a CDP session is gone

Dmitriusan:
`PuppeteerDevToolsConnection.send()` is declared to return a promise of a result or an error, but it throws synchronously when the session id no longer resolves. The generated agent code does not catch that, so it escapes as an uncaught exception and ends the process.

A session going away while commands for it are in flight is normal rather than exceptional. The way I kept hitting it is closing a page while its source maps are still loading:

```
Error: Unknown session 5C39E3164EE6C2C4AB1D3788F5AE03DA
    at PuppeteerDevToolsConnection.send (build/src/devtools/DevToolsConnectionAdapter.js:31:19)
    at AgentPrototype.invoke [as invoke_close]
    at IOModel.close
    at PageResourceLoader.loadFromTarget
    at async PageResourceLoader.dispatchLoad
    at async PageResourceLoader.loadResource
    at async loadSourceMap
```

Source maps are just the most likely way to be holding a load open long enough to notice; any in-flight command to a session that goes away does the same thing. In the server this means closing a tab at the wrong moment can take everything down.

This returns an error response instead, which is how a failed command is already reported a few lines below, so callers see it through the path they already handle.

The other synchronous throw in that method, for `sessionId === undefined`, is left alone: it guards an invariant the callers are supposed to uphold, so failing loudly there seems right. Happy to change that too if you would rather have both consistent.

The test fails on `main` with `Error: Unknown session session-that-went-away` and passes with the change.

Found while working on #2431, but unrelated to it and independent of #2463.

### DIFF
--- a/src/devtools/DevToolsConnectionAdapter.ts
+++ b/src/devtools/DevToolsConnectionAdapter.ts
@@ -56,7 +56,17 @@ export class PuppeteerDevToolsConnection
     }
     const session = this.#connection.session(sessionId);
     if (!session) {
-      throw new Error('Unknown session ' + sessionId);
+      // A session can go away while commands for it are still in flight, for
+      // example when a page is closed while its source maps are loading. That
+      // is normal, so report it the same way a failed command is reported.
+      // Throwing here would escape synchronously out of a method declared to
+      // return a promise, past the generated agent code, and end the process.
+      return Promise.resolve({
+        error: {
+          code: -32000,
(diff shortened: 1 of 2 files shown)


szuend:
lgtm, but ideally we upstream this and get rid of the copy on the MCP side. We can land for now to improve the stability.

szuend (approved):

PR #2471, opened 2026-08-05
ci: add min-release-age to npmrc

OrKoN:


.npmrc
```diff
@@ -1 +1,2 @@
 package-lock=true
+min-release-age=7 # requires npm 11.10.0+
```
nattallius, line 2: What about exceptions from https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/.github/dependabot.yml#L12

OrKoN: dependabot does not use .npmrc so it won't be affected. This would only be used by developers locally (where it is easy to turn off if needed).

OrKoN: we can add `min-release-age-exclude` though.

OrKoN: Added our packages to min-release-age-exclude

nattallius (approved):

PR #2473, opened 2026-08-05
docs: correct the user data directory documentation

kataokatsuki:
The documented user data directory does not match what the server creates.

`launch()` appends the channel name only for non-stable channels, so the default directory
is `chrome-profile`. This also updates the Windows path and clarifies that the same user
data directory cannot be used by multiple browser instances at once. The existing browser
test covers that behavior.


README.md
```diff
@@ -848,16 +848,19 @@ server instances.
 
 ### User data directory
 
-`chrome-devtools-mcp` starts a Chrome's stable channel instance using the following user
+By default, `chrome-devtools-mcp` starts a stable Chrome instance using the following user
```
nattallius:
```suggestion
By default, `chrome-devtools-mcp` starts a Chrome's stable channel instance using the following user
```

README.md
```diff
@@ -848,16 +848,19 @@ server instances.
...
+- Windows: `%USERPROFILE%\.cache\chrome-devtools-mcp\chrome-profile`
 
-The user data directory is not cleared between runs and shared across
-all instances of `chrome-devtools-mcp`. Set the `isolated` option to `true`
-to use a temporary user data dir instead which will be cleared automatically after
-the browser is closed.
+For non-stable channels, the channel name is appended to the directory name, for example
+`chrome-profile-canary`.
+
+The default user data directory is not cleared between runs and is reused for subsequent
```
nattallius:
```suggestion
The user data directory is not cleared between runs and is reused for subsequent
```

nattallius (changes requested):
Hi, LGTM % comments

kataokatsuki:
Thanks, applied both suggestions.

nattallius (approved):

PR #2474, opened 2026-08-05
build: split mcp publishing into a separate workflow

### LINKED ISSUE #2470: Release workflow downloads `mcp-publisher` from another org's `releases/latest` unpinned, then hands that binary the OIDC publishing credential
`.github/workflows/publish-to-npm-on-tag.yml`, as on `main`.

## The shape

`id-token: write` is granted at **workflow** level, so every job inherits it:

```yaml
permissions:
  id-token: write # Required for OIDC
  contents: read
```

That permission is the publishing credential — the npm step uses trusted publishing and there is no
token secret anywhere in the file:

```yaml
npm publish --provenance --access public
```

The second job inherits it and declares no override:

```yaml
  publish-to-mcp-registry:
    runs-on: ubuntu-latest
    needs: publish-to-npm
```

and inside that job:

```yaml
      - name: Install MCP Publisher
        run: |
          export OS=$(uname -s | tr '[:upper:]' '[:lower:]')_$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
          curl -L "https://github.com/modelcontextprotocol/registry/releases/latest/download/mcp-publisher_${OS}.tar.gz" | tar xz mcp-publisher

      - name: Login to MCP Registry
        run: ./mcp-publisher login github-oidc
```

A binary is fetched from a **different organisation's** `releases/latest` with no digest, no checksum
and no signature check — and two steps later it is invoked in a job that holds `id-token: write`.

## Three things I measured rather than assumed

**1. `releases/latest` is mutable.** Resolving the exact URL the workflow uses:

```
HTTP/1.1 302 Found
Location: https://github.com/modelcontextprotocol/registry/releases/download/v1.8.0/mcp-publisher_linux_amd64.tar.gz
```

Two runs of this unchanged workflow, either side of an upstream release, execute different bytes.

**2. Every Action in the same job is pinned by digest — this one download is not.** Counted in the
file, four pinned and one not:

```
  28:        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
  33:        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
  69:        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
  74:        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
```

So the standard is already the project's own; the gap is a third-party *binary* rather than a
third-party *action*.

**3. The verification material exists upstream and is not fetched.**
`modelcontextprotocol/registry` publishes, for the exact tarball this workflow downloads:

```
mcp-publisher_linux_amd64.tar.gz
mcp-publisher_linux_amd64.tar.gz.sbom.json
mcp-publisher_linux_amd64.tar.gz.sigstore.json
```

The workflow fetches the first and ignores the other two.

## What a binary in that position is handed

Reproducing the **shape** in a private repository I own — workflow-level `id-token: write`, a second
job with no override, and a stand-in invoked exactly where `./mcp-publisher login github-oidc` runs.
Nothing was published to npm or any registry, and the stand-in redacts what it receives.

Control first, an ordinary step in that job:

```
control: does an ordinary step see the OIDC variables?
  yes
```

Then the stand-in, in the position the downloaded binary occupies:

```
[canary] argv: login github-oidc
[canary]   ACTIONS_ID_TOKEN_REQUEST_URL   : present
[canary]   ACTIONS_ID_TOKEN_REQUEST_TOKEN : present (4086 chars)
[canary]   *** MINTED AN OIDC TOKEN: 1998 chars ***
[canary]   audience : npm:registry.npmjs.org
[canary]   the token itself is NOT printed
```

The audience is the point: `npm:registry.npmjs.org` is what npm trusted publishing consumes, so what
this job exposes to a downloaded binary is not a generic identity token.

## What I did NOT do, and cannot

I did not exploit this. Doing so would mean publishing a release in
`modelcontextprotocol/registry`, a third party — out of bounds, and it stays out of bounds. So this
is a report about **configuration** and about what a binary in that position is handed, not a
demonstrated compromise. The preconditions are correspondingly narrow: it needs someone able to
publish a release asset in that other project.

I would rather state that than imply an exploit chain I did not run.

## Suggested fix

Two independent changes, either of which helps and both of which are small:

1. **Verify the download.** `mcp-publisher` ships a Sigstore bundle for the exact asset, so
   `gh attestation verify` or `cosign verify-blob` closes it in one step. A pinned release tag plus a
   checksum comparison is the weaker but still adequate version.
2. **Scope the permission.** Move `id-token: write` out of the workflow-level `permissions` block
   into only the job that publishes, so a job that runs a downloaded binary does not inherit the
   ability to publish.

## Disclosure

Reported to Google's OSS VRP, which assessed it as below the threshold they use for escalating to
product teams and invited public disclosure — which is why it is filed here. Filing it as an ordinary
hardening issue rather than an advisory, since what is demonstrated is the configuration and not an
exploited chain.

OrKoN:
- separate workflow distinct from the one configured in npm
- binary validation using cosign
- pinned to 1.8.0 (it should be more stable now).

With this change MCP registry might update even if the npm publishing fails.

Fixes https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2470

### DIFF
--- a/.github/workflows/publish-to-mcp-registry-on-tag.yml
+++ b/.github/workflows/publish-to-mcp-registry-on-tag.yml
@@ -1,0 +1,71 @@
+name: publish-to-mcp-registry-on-tag
+
+on:
+  push:
+    tags:
+      - 'chrome-devtools-mcp-v*'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  publish-to-mcp-registry:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write # Required for OIDC
+      contents: read
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          cache: npm
+          node-version-file: '.nvmrc'
+          registry-url: 'https://registry.npmjs.org'
+
+      # Ensure npm 11.5.1 or later is installed
+      - name: Update npm
+        run: npm install -g npm@latest
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build and bundle
+        run: npm run bundle
+        env:
+          NODE_ENV: 'production'
+
+      - name: Install Cosign
+        uses: sigstore/cosign-installer@ba7bc0a3fef59531c69a25acd34668d6d3fe6f22 # v4.1.0
+
+      - name: Install MCP Publisher
+        run: |
+          export OS=$(uname -s | tr '[:upper:]' '[:lower:]')_$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
+          export VERSION=v1.8.0
+          # 1. Download the release archive and its sigstore bundle
+          curl -L -o mcp-publisher.tar.gz \
+            "https://github.com/modelcontextprotocol/registry/releases/download/${VERSION}/mcp-publisher_${OS}.tar.gz"
+
+          curl -L -o mcp-publisher.tar.gz.sigstore.json \
+            "https://github.com/modelcontextprotocol/registry/releases/download/${VERSION}/mcp-publisher_${OS}.tar.gz.sigstore.json"
+
+          # 2. Verify using cosign verify-blob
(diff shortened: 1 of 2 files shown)


OrKoN:
@Lightning00Blade do you have any other ideas?

Lightning00Blade (approved):

PR #2477, opened 2026-08-05
fix: do not block tool calls on a roots/list the client is slow to answer

### LINKED ISSUE #2476: A client that never answers roots/list blocks every tool for 60s
### Description of the bug

`getContext()` awaits `updateRoots()` whenever it builds a new `McpContext`, and `ToolHandler.handle()` has already taken the tool mutex by then. So the server blocks on a `roots/list` request to the client while holding the lock that serializes every tool call.

If the client negotiated the `roots` capability but does not answer, that request only settles when the SDK's default request timeout of 60s expires. The tool call then completes normally – it never needed the roots – but it took a minute, and every other tool call queued behind the mutex for the same minute.

`context` starts out undefined, so this is not only the path after a browser reconnect. The first tool call of every session takes it.

The wait does not look load-bearing: the `catch` around the request already logs and continues, and the other two `updateRoots()` call sites, in `oninitialized` and in the `roots/list_changed` handler, are both fire-and-forget. Nothing surfaces while it happens either, since that logging is off by default, so it reads as an unexplained hang.

Relevant lines: `src/index.ts:82` and `src/index.ts:86` (fire-and-forget), `src/index.ts:161` (awaited), `src/ToolHandler.ts:213` (mutex acquired) and `src/ToolHandler.ts:222` (`getContext()` awaited).

Related: #1860 added the roots support this sits on, and #2426 was the same shape, something held under the tool mutex freezing the whole session rather than the one call that triggered it.

### Reproduction

A client that advertises `roots` and never responds to `roots/list`, run from a built checkout. No browser restart or reconnect is needed – this is the first tool call of a fresh session.

```js
import {spawn} from 'node:child_process';

const CLI = 'build/src/bin/chrome-devtools-mcp.js';
const ANSWER_ROOTS = process.argv[2] === 'answer';

const child = spawn(process.execPath, [CLI, '--headless', '--isolated'], {
  stdio: ['pipe', 'pipe', 'ignore'],
});

let buffer = '';
const pending = new Map();
child.stdout.on('data', chunk => {
  buffer += chunk.toString();
  let index;
  while ((index = buffer.indexOf('\n')) >= 0) {
    const line = buffer.slice(0, index);
    buffer = buffer.slice(index + 1);
    if (!line.trim()) continue;
    let message;
    try {
      message = JSON.parse(line);
    } catch {
      continue;
    }
    if (message.method === 'roots/list' && message.id != null) {
      console.log(`  server -> client: roots/list (client ${ANSWER_ROOTS ? 'answers' : 'stays silent'})`);
      if (ANSWER_ROOTS) {
        child.stdin.write(JSON.stringify({jsonrpc: '2.0', id: message.id, result: {roots: []}}) + '\n');
      }
      continue;
    }
    if (message.id != null && pending.has(message.id)) {
      pending.get(message.id)(message);
      pending.delete(message.id);
    }
  }
});

let id = 0;
const call = (method, params) =>
  new Promise(resolve => {
    const myId = ++id;
    pending.set(myId, resolve);
    child.stdin.write(JSON.stringify({jsonrpc: '2.0', id: myId, method, params}) + '\n');
  });

await call('initialize', {
  protocolVersion: '2024-11-05',
  capabilities: {roots: {listChanged: true}},
  clientInfo: {name: 'repro', version: '1'},
});
child.stdin.write(JSON.stringify({jsonrpc: '2.0', method: 'notifications/initialized'}) + '\n');

const start = Date.now();
const result = await call('tools/call', {name: 'list_pages', arguments: {}});
console.log(`  first tools/call took ${((Date.now() - start) / 1000).toFixed(1)}s`, result.error ? 'ERROR' : 'ok');
child.kill();
```

```
$ node repro.mjs answer
  server -> client: roots/list (client answers)
  server -> client: roots/list (client answers)
  first tools/call took 0.8s ok

$ node repro.mjs
  server -> client: roots/list (client stays silent)
  server -> client: roots/list (client stays silent)
  first tools/call took 60.4s ok
```

Adding a second, unrelated `list_pages` three seconds into the stall shows it is not just the triggering call:

```
  call A (issued at 0s)   finished at 61.0s
  call B (issued at 3s)   finished at 61.0s
```

### Expectation

A client that answers `roots/list` slowly, or not at all, should not be able to freeze every tool for a minute. Delaying the call that triggered it is one thing, taking the whole server with it is another.

I hit this from a wrapper that proxies `chrome-devtools-mcp` over stdio and briefly held the client's response back, so the trigger was on my side and I have fixed it there. Filing it because the server has no bound of its own, the stall covers every tool rather than the one that triggered it, and a client answering a capability it advertised slowly is not that exotic.

### MCP configuration

The repro spawns the CLI directly rather than going through a client config, so there is no `mcpServers` entry behind it.

1.6.0, and the same shape is on `main` at `d302a47`

### Chrome version

Google Chrome for Testing 151.0.7922.47, launched by the server via `--headless --isolated`

### Coding agent version

Claude Code 2.1.92

### Model version

Claude Opus 5

### Chat log

n/a – the repro above is a scripted client, not an agent session

### Node version

v24.19.0

### Operating system

macOS

thomasbachem:
`getContext()` awaits `updateRoots()` while `ToolHandler.handle()` holds the tool mutex, so a client that negotiated `roots` and never answers blocks every tool for the SDK's 60s default.

This bounds the one awaited listing rather than `updateRoots()` as a whole, since the other two call sites are already fire-and-forget and bounding them would only throw away roots a slow client was about to send. The last listing is then kept and handed to each new `McpContext`. Roots are client state rather than browser state, so a reconnect does not invalidate them, and real changes still arrive through `roots/list_changed`.

Both halves turn out to be needed, which took me a couple of tries to see. Bound every listing and a client answering in 8s never gets roots at all, because each of its listings times out. Leave the background ones unbounded but keep no cache and they are lost anyway, since that listing can land while `context` is still undefined and `context?.setRoots()` drops it. Either way, file writes stay confined to the temp directory for the rest of the session, where today that client works and merely waits.

It does mean a client declaring `roots` without `listChanged` can see the previous list on its first call against a new context, until the background refresh lands.

Tests for each half:

- `does not block tools if the client never answers roots/list` fails on `main` and passes here. It raises the client-side timeout above the SDK default so the stall surfaces as the assertion rather than as a test timeout

```
✖ does not block tools if the client never answers roots/list (60902.121334ms)
  AssertionError [ERR_ASSERTION]: list_pages took 60438ms, expected the bounded roots request to settle well before the 60s SDK default
```

- `still applies roots from a client slower than the bound` passes on `main` and here, and fails if the bound covers every listing rather than the awaited one

`npm run test` shows four failures here, but the same four fail on unmodified `main`.

#2285 changes the `catch` inside `updateRoots()` that this rewrites, so whichever lands second needs a rebase. Happy to do it if this goes last.

Fixes #2476


FirdavsIdriszoda (approved):

nattallius:
Looks good to me

thomasbachem:
Rebased onto main – no change to this PR's code.

That picks up #2482, which should cover the two waits failing on the Windows cell at 12.06s. The Lighthouse timeouts in the same run look like separate runner slowness rather than anything this touches.

nattallius (approved):

PR #2481, opened 2026-08-06
build: consume devtools-frontend as a submodule

OrKoN:
Adding devtools-frontend from github as a submodule with depth of 1 to speed up checkouts. prepare.ts script should automatically pull the right version.


tsconfig.json
```diff
@@ -6,8 +6,10 @@
       "DOM",
       "ES2024.Promise",
       "ESNext.Iterator",
-      "ESNext.Collection"
+      "ESNext.Collection",
+      "ESNext.Disposable"
```
Lightning00Blade: If you want to do this you need to polyfill them. I commended this on one of Simon's tasks as well.

Lightning00Blade (approved):
LGTM

yulunz (approved):

tsconfig.json
OrKoN: I think this is actually got added by accident.

PR #2482, opened 2026-08-06
test: Support configurable navigation timeout and increase it for Windows tests

nattallius:
This change adds a configurable navigation timeout to McpContext options, allowing tests to override the default 10-second limit. We use this in `tests/utils.ts` to set a 20-second timeout on Windows. This **potentially** fixes flaky test timeouts on slow Windows environments by ensuring that both direct navigations and click-triggered navigations (which use WaitForHelper) are synchronized to the longer 20-second limit and are not reset back to 10 seconds during emulation.

### DIFF
--- a/src/McpPage.ts
+++ b/src/McpPage.ts
@@ -137,6 +137,7 @@ export class McpPage implements ContextPage {
 
   #hasNetworkBlockOrAllowlist: boolean;
   #locatorClass: typeof Locator;
+  #navigationTimeout: number;
 
   constructor(
     page: Page,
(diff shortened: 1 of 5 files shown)


src/McpPage.ts
```diff
@@ -409,7 +412,14 @@ export class McpPage implements ContextPage {
...
     networkMultiplier: number,
   ): WaitForHelper {
-    return new WaitForHelper(this.pptrPage, cpuMultiplier, networkMultiplier);
+    const navigationTimeout =
+      this.#navigationTimeout * networkMultiplier * cpuMultiplier;
+    return new WaitForHelper(
+      this.pptrPage,
+      cpuMultiplier,
+      networkMultiplier,
+      navigationTimeout,
```
OrKoN: navigation timeout in the WaitForHelper is a different timeout IIUC. I think we should not pass this.#navigationTimeout * networkMultiplier * cpuMultiplier as navigation timeout for WaitForHelper here. Additionally, I do not think the WaitForHelper timeout was the cause of flakiness.

OrKoN (approved):
Thanks!

PR #2488, opened 2026-08-06
chore: refactor the profiler into scenarios.

yulunz:
This is mostly a refactor except:
- the url param is removed. The source map server is always turned on. We can always manually connect the mcp server to an external site and check for memory usage / leaks. Testing should be deterministic and not rely on outside resources.
- the reports are now stored by scenario. Console still prints out report for each scenario, alongside a final report giving a one line summary per scenario.

Still a bit rough, but useful.


scripts/profile/profile_mcp.ts
```diff
@@ -0,0 +1,316 @@
...
+
+    const results: ScenarioResult[] = [];
+    for (const scenarioEntry of scenarios) {
+      const scenarioOutputDir =
+        scenarios.length === 1 && values['output-dir'] !== undefined
+          ? baseOutputDir
+          : path.join(baseOutputDir, scenarioEntry.name);
+
+      const scenarioArgs: ScenarioArgs = {
+        targetUrl: sourceMapTestServer.url,
```
OrKoN: let's keep the URL option to be able to run against real web sites. I think not everything can be caught with the test server.

OrKoN (approved):

scripts/profile/profile_mcp.ts
yulunz: done now. URL can now be overriden on the command line for all scenarios.

PR #2495, opened 2026-08-07
chore: change dependabot cooldown default days to 0

OrKoN:
it looks like exclude does not work and we only roll devtools-frontend as submodule

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -58,5 +58,3 @@ updates:
       timezone: 'Europe/Berlin'
     cooldown:
       default-days: 0
-      exclude:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -57,6 +57,6 @@ updates:
       time: '06:00'
       timezone: 'Europe/Berlin'
     cooldown:
-      default-days: 7
+      default-days: 0
       exclude:
         - 'devtools-frontend'
```
OrKoN:
```suggestion
```

Lightning00Blade (approved):

PR #2501, opened 2026-08-08
fix(input): select empty-value options by text

Iams4kura:
## Summary

- Fixes: When fill targets a native select option whose visible text matches the requested value but whose DOM value is the empty string, the tool reports success while the previously selected option remains selected.
- Root cause: selectOption tests the option's DOM value for truthiness before calling Locator.fill. An empty string is a valid option value, but the truthiness check skips the fill while optionFound remains true, so the handler returns a false success.

## Regression evidence

- Before: `npx --yes -p node@24 -c 'node --version && npm run test tests/tools/input.test.ts -- --test-name-pattern=empty.value.by.text'` exited `1`

- After: `npx --yes -p node@24 -c 'node --version && npm run test tests/tools/input.test.ts -- --test-name-pattern=empty.value.by.text'` exited `0`

## Verification

- `npx --yes -p node@24 -c 'node --version && npm run test tests/tools/input.test.ts'`
- `npx --yes -p node@24 -c 'node --version && npm run build'`
- `npx --yes -p node@24 -c 'node --version && npm run check-format'`
- `npx --yes -p node@24 -c 'node --version && npm run gen'`

The full `npm run test` was also attempted. On this machine it reaches an
unrelated Chrome 151 `Page.captureScreenshot: Page is too large` failure; the
same screenshot test fails identically on the exact unmodified base SHA under
Node 24. The complete affected input test suite passes.

## Scope

- 2 files changed, +40 / -2 lines

### DIFF
--- a/src/tools/input.ts
+++ b/src/tools/input.ts
@@ -221,8 +221,8 @@ async function selectOption(
         using childValueHandle = await childHandle.getProperty('value');
 
         const childValue = await childValueHandle.jsonValue();
-        if (childValue) {
-          await handle.asLocator().fill(childValue.toString());
+        if (typeof childValue === 'string') {
+          await handle.asLocator().fill(childValue);
         }
 
         break;
--- a/tests/tools/input.test.ts
+++ b/tests/tools/input.test.ts
@@ -590,6 +590,44 @@ describe('input', () => {
       });
     });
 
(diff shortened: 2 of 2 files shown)


Lightning00Blade (approved):
LGTM, thanks!

PR #2504, opened 2026-08-08
feat: report newly opened pages in the action response

### LINKED ISSUE #367: return all pages when click web element

ZayanKhan-12:
## What

When a click opens a new page (link with `target=_blank`, `window.open()`), the tool response now says so and includes the list of open pages. Previously the response only said "Successfully clicked on the element", so clients could not perceive that a new tab/window had opened and assumed the click had no effect.

## How

- `WaitForHelper` now listens for the page's `popup` event while the action runs and reports `newPagesOpened` in `WaitForEventsResult` (alongside the existing `navigatedToUrl`). This automatically covers every input tool that uses `waitForEventsAfterAction` (`click`, `click_at`, `fill`, `press_key`, etc.).
- `McpResponse.attachWaitForResult()` calls the existing `setIncludePages(true)` when a new page was opened, reusing the exact same page-list formatting as `list_pages` — no new format is introduced.
- A note line and `structuredContent.newPagesOpened` are emitted next to the existing `Page navigated to …` handling.

Token-conscious per the discussion in #367: the page list is only attached when a new page actually opened during the action; responses for ordinary clicks are unchanged, and no snapshot/tree is included.

## Before

```
Successfully clicked on the element
```

## After

```
Successfully clicked on the element
The action opened a new page. See the list of pages below and call select_page to switch to a page.
## Pages
1: My test page (about:blank) [selected]
2: about:blank
```

(sample captured from the new test run)

## Testing

- `npm run build` — pass
- `npm run test:no-build` — full suite pass (exit 0)
- new test: `input > click > reports newly opened pages` proves a `window.open` click surfaces the note and the `## Pages` list
- `npm run check-format` — pass
- `node scripts/generate-docs.ts` — no doc changes (no tool descriptions/schemas touched)

Fixes #367

### DIFF
--- a/src/WaitForHelper.ts
+++ b/src/WaitForHelper.ts
@@ -21,6 +21,8 @@ export class WaitForHelper {
   #dialogHandled = false;
   /** Track all dialogs as they pause the renderer. */
   #dialogDetected = false;
+  /** Whether the action opened one or more new pages. */
+  #newPagesOpened = false;
   #initialUrl: string;
 
   constructor(
@@ -185,6 +187,16 @@ export class WaitForHelper {
       this.#page.off('dialog', dialogHandler);
     });
 
+    // Detect pages opened by the action (e.g., a click on a link with
+    // target=_blank or a window.open() call) so the response can report them.
+    const popupHandler = () => {
+      this.#newPagesOpened = true;
+    };
+    this.#page.on('popup', popupHandler);
+    this.#abortController.signal.addEventListener('abort', () => {
+      this.#page.off('popup', popupHandler);
+    });
+
     const navigationFinished = this.waitForNavigationStarted()
       .then(navigationStated => {
         if (navigationStated) {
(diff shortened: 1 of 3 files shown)


src/WaitForHelper.ts
```diff
@@ -21,6 +21,8 @@ export class WaitForHelper {
   #dialogHandled = false;
   /** Track all dialogs as they pause the renderer. */
   #dialogDetected = false;
+  /** Whether the action opened one or more new pages. */
+  #newPagesOpened = false;
```
Lightning00Blade, line 25: Looked though the code, we can be explicit and return the page that was open.
Puppeteer creates a single object so from then we can update the snapshot and return only what page data was opened rather then returning the full list and leaving AI to try to figure it out.

Lightning00Blade (commented):
The idea makes sense, but the approach would be better if we only say what was opened.

OrKoN:
Closing as stale

PR #2505, opened 2026-08-08
feat: add optional stack traces to list_console_messages

### LINKED ISSUE #383: Expose error stack traces in list_console_messages (feasibility question)

Today list_console_messages returns plain console text without the error stack trace. When combining logs with LLM-based debugging, the model often can’t determine which file and line actually failed. This forces extra manual steps, guesswork, and back-and-forth context gathering—overall, not an elegant debugging experience.

At a high level (no design details): the ability to optionally obtain error stack traces alongside messages returned by list_console_messages. The goal is to let tools/LLMs immediately map an error to the concrete file/function/line so they can provide accurate, actionable fixes. 

Manually fetching stacks via separate DevTools/diagnostic calls and correlating them with console messages. This adds complexity, latency, and token cost.

### Additional context

This capability would significantly improve automated triage and remediation in MCP-based workflows by enabling precise source localization rather than plain-text analysis.

Open questions for maintainers:
- Does the underlying layer already capture or have access to stacks for error messages?
- If yes, is exposing them in `list_console_messages` feasible?
- If not, are there known technical/security/size constraints or an alternative path we should consider?

Thanks for considering!

ZayanKhan-12:
## What

Adds an opt-in `includeStackTraces` parameter to `list_console_messages`. When set, each listed message includes its resolved (source-mapped, ignore-list-filtered) stack trace — in the text output as indented frames under the message line, and in `structuredContent` as a pre-formatted `stackTrace` string, matching the format `get_console_message` already uses.

## Why

Fixes the workflow described in #383: today the list output only contains plain message text, so an LLM debugging a page has to call `get_console_message` once per message to learn which file/function/line failed. With `includeStackTraces: true` the model can map every error to its source in a single call.

Design notes:

- **Opt-in, default off** — the default output is byte-for-byte unchanged, keeping the token cost of the common path identical.
- Reuses the existing DevTools stack resolution (`createStackTraceForConsoleMessage` / `SymbolizedError`) and the existing `formatStackTrace` formatting, so frames are source-mapped and ignore-listed exactly like in `get_console_message`.
- The per-message "1-based indexing" note is emitted once for the whole list instead of once per message to keep the output compact.
- Works for console API messages, uncaught errors (including `Caused by:` chains), and grouped repeated messages (`[N times]`).

## Before / after

Default (unchanged):

```
## Console messages
Showing 1-1 of 1 (Page 1 of 1).
msgid=1 [log] Hello stack trace! (0 args)
```

With `includeStackTraces: true`:

```
## Console messages
Showing 1-1 of 1 (Page 1 of 1).
msgid=1 [log] Hello stack trace! (0 args)
  at foo (foo.ts:10:2)
  at bar (foo.ts:20:2)
  --- setTimeout -------------------------
  at schedule (util.ts:5:2)
Note: stack trace line and column numbers use 1-based indexing
```

`structuredContent.consoleMessages[i]` gains an optional `"stackTrace": "at foo (foo.ts:10:2)\nat bar (foo.ts:20:2)\n..."` field.

## Relationship to #2060

Complements #2060 (source location for `list_console_messages`, issue #903): that PR surfaces the top-frame *location* for every message by default, while this one exposes the *full stack trace* behind an explicit opt-in. The features are independent and the changes are designed to coexist with minimal conflict.

## Testing

- `npm run build` — passes
- `npm run test:no-build` — full suite passes (exit 0)
- `npm run check-format` — passes
- `npm run gen` — regenerated `docs/tool-reference.md`, CLI options, and tool-call metrics (included)

New coverage:

- `tests/formatters/ConsoleFormatter.test.ts` — snapshot tests for concise `toString`/`toJSON` with `fetchStackTrace` set (console message + uncaught error) and for the unchanged default.
- `tests/tools/console.test.ts` — integration tests asserting frames (`at failingFn`) appear in text and `structuredContent` when `includeStackTraces: true`, and are absent by default.

Fixes #383

### DIFF
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -391,6 +391,7 @@
 **Parameters:**
 
 - **includePreservedMessages** (boolean) _(optional)_: Set to true to return the preserved messages over the last 3 navigations.
+- **includeStackTraces** (boolean) _(optional)_: Set to true to include the stack trace for each message when available. Increases the response size.
 - **pageIdx** (integer) _(optional)_: Page number to return (0-based). When omitted, returns the first page.
 - **pageSize** (integer) _(optional)_: Maximum number of messages to return. When omitted, returns all messages.
 - **serviceWorkerId** (string) _(optional)_: Filter messages to only return messages of the specified service worker.
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -106,6 +106,7 @@ export class McpResponse implements Response {
     pagination?: PaginationOptions;
     types?: string[];
     includePreservedMessages?: boolean;
+    includeStackTraces?: boolean;
     serviceWorkerId?: string;
   };
   #listExtensions?: boolean;
@@ -212,6 +213,7 @@ export class McpResponse implements Response {
     options?: PaginationOptions & {
       types?: string[];
       includePreservedMessages?: boolean;
+      includeStackTraces?: boolean;
       serviceWorkerId?: string;
     },
   ): void {
@@ -231,6 +233,7 @@ export class McpResponse implements Response {
           : undefined,
       types: options?.types,
       includePreservedMessages: options?.includePreservedMessages,
+      includeStackTraces: options?.includeStackTraces,
       serviceWorkerId: options?.serviceWorkerId,
     };
   }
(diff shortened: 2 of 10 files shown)


OrKoN (approved):
LGTM (this matches what we should in DevTools UI so generally could be useful to agents if there are not so many console messages).

PR #2506, opened 2026-08-08
fix: capture navigation requests for pages opened via window.open

### LINKED ISSUE #2395: list_network_requests misses navigation requests in tabs opened via window.open
## Description of the feature

When a page opens a new tab itself (`window.open` / `target=_blank` from a click handler), `list_network_requests` on that tab returns nothing — the entire navigation, redirect chain included, is missed. `includePreservedRequests: true` doesn't help; nothing was ever recorded. Quiet failure: an empty listing reads as "this tab made no requests".

Cause: new targets are wired reactively (`McpContext.ts` — `targetcreated` → `await target.page()` → `void #createMcpPage(page)`), so the popup navigates before the collector's `Network.enable` lands. And unlike console messages (#2382), Chromium doesn't replay network events to late CDP sessions, so there's nothing to backfill.

## Reproduction

1. `new_page` →

   ```
   data:text/html,<a href="#" onclick="window.open('https://httpbingo.org/redirect-to?url=https%253A%252F%252Fexample.com%252F','_blank');return false;">OPEN</a>
   ```
2. `take_snapshot`, `click` the link
3. `list_pages` → the popup is there, already landed on `https://example.com/` (redirect chain completed)
4. `select_page` the popup, `list_network_requests` with `includePreservedRequests: true` → empty

Sanity check: navigate that same popup anywhere afterwards and the request shows up fine — capture works, it just started too late.

## Expectation

Requests in opened tabs are captured from the first request. Since CDP offers no replay for Network events, the reliable shape seems to be pausing new targets until the collector is wired: `Target.setAutoAttach` with `waitForDebuggerOnStart: true`, enable Network on attach, then `Runtime.runIfWaitingForDebugger`.

## Proposal

The proposed solution is to hold old new targets until the collector is wired: browser-level `Target.setAutoAttach` with `waitForDebuggerOnStart: true`, enable Network on attach, then `Runtime.runIfWaitingForDebugger`. The pause is a few milliseconds per new tab.

Enabling it could work like this:

* a server flag, e.g. --captureNewTabs, enabling pause-and-wire globally; or
* a runtime tool to arm it on demand

* `includePreservedRequests` — doesn't apply: it preserves already-captured requests across navigations, and here nothing was ever captured.
* `--chromeArg=--auto-open-devtools-for-tabs` — verified, no effect: the held first navigation feeds the DevTools frontend's own CDP session, not the server's.
* External sidecar: a second CDP client that arms `setAutoAttach` itself and logs Network events (our current workaround), but lives outside the server and its capture is invisible to list_network_requests, so results come from two sources.

### Additional context

Context: we run browser-ground-truth tests, where verifying click → redirect chain → landing is the whole point. The chain start is exactly the part that's lost.

## Chrome DevTools MCP version

1.6.0

ZayanKhan-12:
## What

`list_network_requests` on a tab that the page opened itself (`window.open`, `target=_blank`) returned an empty listing: the entire initial navigation, redirect chain included, was missing, and `includePreservedRequests: true` could not help because nothing was ever recorded.

Repro from the issue: `new_page` with a link that calls `window.open(<redirecting url>)`, `click` it, `select_page` the popup, `list_network_requests` → empty, even though the popup has long landed on the final URL.

## Root cause

New targets are wired up reactively: `targetcreated` → `await target.page()` → `#createMcpPage(page)`. But Puppeteer's `TargetManager` sends `Runtime.runIfWaitingForDebugger` immediately when the target attaches, so the popup starts navigating right away — long before `target.page()` creates the `CdpPage` whose `NetworkManager` sends `Network.enable`. Chromium does not replay Network events to sessions that enable the domain late (unlike e.g. console messages), so the navigation request is unobservable by the time the collector is listening.

Note: a second browser-level `Target.setAutoAttach` with `waitForDebuggerOnStart: true` (as proposed in the issue) does not work from within this server: I verified empirically that the target resumes as soon as Puppeteer sends its own `Runtime.runIfWaitingForDebugger`, regardless of another session still holding.

## Fix

New `NetworkEventBuffer` (owned by `McpContext`), based on two ordering guarantees:

1. Puppeteer's `Connection` emits `SessionAttached` for a new session *before* it routes the `Target.attachedToTarget` event to the parent session where `TargetManager`'s listener runs. A `Target.attachedToTarget` listener registered on each new session at that moment therefore runs *before* `TargetManager`'s for all child attaches (popup pages attach under their fresh tab session).
2. CDP processes the commands of a session in order. Sending `Network.enable` from that listener means it is dispatched — and processed — before the `Runtime.runIfWaitingForDebugger` with which `TargetManager` resumes the target moments later on the same session.

With the Network domain enabled from the first moment, the buffer records the Network events (capped, with timeout/detach cleanup) while nothing else is listening, and stops recording as soon as Puppeteer's `NetworkManager` attaches (detected via `listenerCount`). Once the `McpPage`'s collectors are wired up, `McpContext` flushes the buffer by re-emitting the events on the session: `NetworkManager` processes them as if live and produces regular `HTTPRequest`s — so the navigation request and its redirect chain show up in `list_network_requests` like any other request, with no second bookkeeping path.

## Latency

None. Unlike the pause-based proposal in the issue, this approach never delays the new tab: it only enqueues one extra `Network.enable` command per newly attached page target and buffers events that would otherwise be dropped. There is a small remaining theoretical gap (events arriving in the few milliseconds between `NetworkManager` attaching and the collectors being wired are processed live before the collector listens), but the navigation request and its redirect chain fire well before that window and are reliably captured.

## Testing

- New regression test: opener page calls `window.open()` on a URL that 302-redirects (with a small delay so the chain is in flight while the popup is being wired up); asserts both the initial navigation request and the redirect target appear in `list_network_requests` for the popup. The test fails without the fix with exactly the reported symptom (empty listing) and passes with it.
- Verified the repro standalone: without the fix the popup's request list is empty; with it, `GET /redirect` (`isNavigationRequest() === true`) and the redirect target are captured.
- `npm run build && npm run test:no-build`: 689/689 pass.
- `npm run check-format` clean.

Fixes #2395

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -18,6 +18,7 @@ import type {
   DuplicateStringGroup,
 } from './HeapSnapshotManager.js';
 import {McpPage} from './McpPage.js';
+import {NetworkEventBuffer} from './NetworkEventBuffer.js';
 import {type UncaughtError} from './PageCollector.js';
 import {ServiceWorkerConsoleCollector} from './ServiceWorkerCollector.js';
 import {
@@ -85,6 +86,7 @@ export class McpContext implements Context {
   #selectedPageFallback?: {wasClosed: boolean};
 
   #serviceWorkerConsoleCollector: ServiceWorkerConsoleCollector;
+  #networkEventBuffer: NetworkEventBuffer;
 
   #isRunningTrace = false;
   #screenRecorderData: {recorder: ScreenRecorder; filePath: string} | null =
@@ -126,6 +128,9 @@ export class McpContext implements Context {
     this.#serviceWorkerConsoleCollector = new ServiceWorkerConsoleCollector(
       this.browser,
     );
+    // Starts capturing network events for new page targets (e.g. tabs opened
+    // via window.open) before their McpPage exists.
+    this.#networkEventBuffer = new NetworkEventBuffer(this.browser);
   }
 
   async #init() {
@@ -142,6 +147,7 @@ export class McpContext implements Context {
     this.browser.off('targetdestroyed', this.#onTargetDestroyed);
 
     this.#serviceWorkerConsoleCollector.dispose();
+    this.#networkEventBuffer.dispose();
     this.#heapSnapshotManager.dispose();
     for (const mcpPage of this.#mcpPages.values()) {
       mcpPage.dispose();
@@ -505,6 +511,9 @@ export class McpContext implements Context {
         ),
       });
       this.#mcpPages.set(page, mcpPage);
+      // Now that the collectors are listening, replay any network events that
+      // were captured before the page was wired up.
+      this.#networkEventBuffer.flush(page);
       await mcpPage.init();
     }
     return mcpPage;
--- a/src/NetworkEventBuffer.ts
+++ b/src/NetworkEventBuffer.ts
@@ -1,0 +1,212 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type {
+  Browser,
+  CDPSession,
+  Connection,
+  Handler,
+  Page,
+  Protocol,
+} from './third_party/index.js';
+import {CDPSessionEvent} from './third_party/index.js';
+import {logger} from './utils/logger.js';
+
+/**
+ * The Network domain events consumed by Puppeteer's NetworkManager.
+ */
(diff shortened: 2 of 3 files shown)


OrKoN (changes requested):
Thanks for the PR but we do not want to add parallel CDP based implementation to the MCP server. I think this would be fixed by https://github.com/puppeteer/puppeteer/issues/15270

PR #2507, opened 2026-08-08
feat: add tool to configure download behavior

### LINKED ISSUE #2397: Add MCP tool to configure download behavior (prompt toggle, downloads location)

There's currently no MCP tool to control Chrome's download behavior at runtime. This makes it hard for an agent to reliably download files to a specific folder during a session and then revert to the user's default behavior afterward.

Concretely, two Chrome preferences aren't exposed:

1. **"Ask where to save each file before downloading"** — a boolean toggle (`download.prompt_for_download` in Chrome's profile prefs).
2. **Downloads Location** — a folder path (`download.default_directory`).

`emulate` currently covers viewport, network conditions, geolocation, color scheme, user agent, and CPU throttling, but nothing about downloads.

Add a tool (or extend `emulate`) that lets an agent set download behavior for the current page/browser context, e.g.:

- `downloadPath` (string) — target directory for downloads, mapping to CDP `Page.setDownloadBehavior` / `Browser.setDownloadBehavior` with `behavior: "allow"` and `downloadPath: <path>`.
- `promptForDownload` (boolean) — whether to prompt for a save location per file, if CDP/Chrome exposes a way to toggle this (otherwise `behavior: "allow"` vs `"default"`/`"deny"` may be the closest equivalent).

Ideally this is settable per-session and easy to reset back to the default (e.g. omitting the option, or an explicit `"default"` value, restores normal browser behavior — no permanent profile mutation required).

It'd also help to be able to **read back the current state** of these settings (current download path and whether prompting is enabled), so an agent can save the existing values before changing them and confirm what's actually in effect after — rather than only being able to set them blindly.

- Asking the user to manually change `chrome://settings/downloads` before/after each session — defeats the purpose of driving the browser via MCP.
- Launching Chrome with a different `--user-data-dir` per desired download folder — too heavyweight for a per-task override and doesn't compose with an already-running session.

### Additional context

Use case: instruct an agent to download some files into a specific folder during a particular task, then have subsequent downloads revert to the default downloads path (e.g. `~/Downloads`) without lingering side effects on the user's normal Chrome profile.

ZayanKhan-12:
## What

Adds a `set_download_behavior` tool (Emulation category) that configures how the browser handles file downloads at runtime, as requested in #2397:

- `behavior: "allow"` + `downloadPath` — save downloads into the given directory without prompting (the directory is created if missing, and the path is validated against the configured workspace roots like other file-path params).
- `behavior: "deny"` — cancel all downloads.
- `behavior: "default"` — restore the browser's default download behavior, so a session can revert cleanly with no lingering side effects on the user's profile.

The tool responds with the effective configuration (e.g. the resolved absolute download directory) so the agent can confirm what is in effect.

## Why

There was previously no MCP tool to control Chrome's download behavior, so an agent could not reliably download files into a specific folder during a task and then revert to the default behavior afterwards (see #2397 for the full use case).

## Scope choice

The issue left per-page vs. global scope open. This PR applies the behavior to the **browser context of the selected page** via Puppeteer's `Browser.setDownloadBehavior`-backed `BrowserContext.setDownloadBehavior()`:

- CDP's per-page `Page.setDownloadBehavior` is deprecated; `Browser.setDownloadBehavior` (scoped by `browserContextId`) is the supported API.
- Browser-context scope matches the reporter's need — set a folder for a task, then reset with `"default"` — and composes with the server's isolated browser contexts (pages created in an isolated context get that context's behavior).

A "prompt for each file" toggle is not exposed because CDP has no such switch; `"default"` restores whatever the profile's normal behavior is (including prompting if the user has it enabled). Reading back the pre-existing profile state is likewise not exposed by CDP, so the tool reports the configuration it applied instead.

## Testing

- New tests in `tests/tools/emulation.test.ts`:
  - configures `allow` with a not-yet-existing directory, triggers a real download from the local test server headlessly, and asserts the file lands in the configured directory with the expected content;
  - configures `deny` and asserts a subsequent download is cancelled (no new file appears);
  - rejects `allow` without `downloadPath`;
  - rejects `downloadPath` combined with `deny`/`default`.
- `npm run build`, full `npm run test:no-build`, and `npm run check-format` all pass locally.
- Docs, CLI options, and tool metrics regenerated via `npm run gen`.

Fixes #2397


OrKoN:
Thanks for the PR but are still collecting feedback on the feature request so are not ready to land the implementation.

Issue #2179, opened 2026-06-05
Feature request: safer defaults for local file, private network, and script evaluation access


junbyjun1238:
Hi Chrome DevTools MCP team,

Could the project consider optional guardrails for `chrome-devtools-mcp` when it is used by local MCP clients that process untrusted content?

I am not reporting a Chrome sandbox bypass or a same-origin policy bypass. The concern is about the local trust boundary created when a client is granted browser automation capabilities through this server.

## Summary

`chrome-devtools-mcp` exposes powerful browser automation primitives that can interact with local and internal boundaries, including:

- navigating to local `file://` URLs
- accessing `localhost`, loopback, private-network, or link-local/metadata-style endpoints from the browser context
- executing JavaScript in the page context
- allowing browser-mediated outbound requests from evaluated scripts

These are expected browser and DevTools capabilities. However, in MCP client deployments, tool calls may be influenced by untrusted web content, documents, repository text, or other external content. In that setting, these capabilities can become surprising data-movement paths for local files, internal services, or browser-visible content.

## Why this matters

In enterprise or developer environments, users may run MCP clients with broad local trust while browsing or analyzing untrusted content. If a client is induced to combine navigation, snapshotting, and script evaluation, sensitive data obtained from local or internal browser-accessible resources may leave the chat-visible path through normal browser network behavior.

Again, this is not a claim that Chrome itself is violating its security model. This is a request to consider safer MCP defaults and clearer controls around high-impact local browser capabilities.

## Suggested guardrails

Would it make sense to consider one or more of the following?

- An opt-in flag for `file://` navigation, such as `--allow-file-scheme`.
- An opt-in flag for `localhost`, loopback, private IP, and link-local/metadata IP access.
- An option to disable the `evaluate_script` tool entirely.
- Where feasible, stricter egress controls for local/internal navigations, such as CSP injection, request interception through CDP, or an equivalent policy.
- An allowlist/denylist policy configuration for URL schemes, hostnames, and IP ranges.
- Clear documentation warnings about local file access, internal network access, and blind egress risks when exposing this server to MCP clients that may process untrusted content.

## Expected outcome

The goal is not to remove DevTools functionality. The goal is to make powerful local browser capabilities explicit and configurable for MCP deployments where users may not realize that browser automation can bridge local files, internal services, and outbound browser traffic.

Thanks for considering this.

clcgbr9852:
[...]

PR #2508, opened 2026-08-08
feat: add --disabled-tool flag to disable individual tools

ZayanKhan-12:
## What

Adds a `--disabledTool` / `--disabled-tool` CLI option (array, repeatable) that prevents the named tools from being exposed to the MCP client:

```
npx chrome-devtools-mcp@latest --disabled-tool=evaluate_script
```

- Disabled tools are never registered with the MCP client (not even in `--viaCli` mode), and `ToolHandler.handle()` additionally refuses to execute them as defense in depth.
- Unknown tool names fail fast at server startup, so a typo cannot silently leave a tool enabled.
- Telemetry only records the presence of the flag (`disabled_tool_present`), not which tools were disabled.

## Why

#2179 asks for security guardrails when `chrome-devtools-mcp` is used by MCP clients that process untrusted content. Several of the requested guardrails already exist (`--allowedUrlPattern`/`--blockedUrlPattern` for URL policy, `--allowUnrestrictedPaths` for file paths), but one concrete request is still open: *"An option to disable the `evaluate_script` tool entirely."*

Category flags (`--no-category-*`) are too coarse for this: script tools cannot be turned off individually. A per-tool disable flag lets operators remove high-impact capabilities (script evaluation, file upload, etc.) from the toolset when tool calls may be influenced by untrusted web content, while keeping the rest of the server fully functional. It is generic, so it also covers future tools without adding one flag per capability.

**Default behavior is unchanged** — the flag is opt-in; without it, the exposed toolset is identical to before.

## Testing

- `npm run build`, `npm run test:no-build` (full suite green) and `npm run check-format`.
- New unit tests in `tests/ToolHandler.test.ts`: disabled tool is not registered and refuses invocation with a clear message; multiple names supported; not registered even with `--viaCli`; other tools unaffected.
- Manual stdio smoke test against the built server: default run lists 29 tools including `evaluate_script`; with `--disabled-tool=evaluate_script` the list drops to 28 without it; `--disabled-tool=bogus_tool` exits at startup with `Unknown tool name(s) passed to --disabled-tool: bogus_tool.`
- `npm run gen` regenerated the README flag reference and `flag_usage_metrics.json`.

Refs #2179

### DIFF
--- a/README.md
+++ b/README.md
@@ -697,6 +697,11 @@ The Chrome DevTools MCP server supports the following configuration option:
   - **Type:** array
   - **Default:** `false`
 
+- **`--disabledTool`/ `--disabled-tool`**
+  Disables the tool with the given name so that it is not exposed to the MCP client. Accepts an array of tool names. Useful as a security guardrail when MCP clients process untrusted content, e.g. --disabled-tool=evaluate_script prevents arbitrary JavaScript execution in the browser.
+  - **Type:** array
+  - **Default:** `false`
+
 - **`--ignoreDefaultChromeArg`/ `--ignore-default-chrome-arg`**
   Explicitly disable default arguments for Chrome. Only applies when Chrome is launched by chrome-devtools-mcp.
   - **Type:** array
--- a/src/ToolHandler.ts
+++ b/src/ToolHandler.ts
@@ -76,10 +76,24 @@ function getConditionStatus(
   return {disabled: false};
 }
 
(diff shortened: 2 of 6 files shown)


OrKoN:
Thanks for the PR but we are still collecting feedback on the feature request so are not ready to land an implementation. Many MCP clients allow configuring allowed/disallowed tools per MCP server.

PR #2510, opened 2026-08-08
test: add unit tests for pagination util

ZayanKhan-12:
## Why

`src/utils/pagination.ts` drives pagination for tool output (`list_network_requests`, `list_console_messages`, memory tools, etc.), so off-by-one mistakes here directly corrupt what agents see. It currently has no unit tests — `tests/utils/` only covers `files.ts` and `url.ts`. This PR adds coverage for the pagination boundaries without touching any source code.

## Boundary cases covered

- No-pagination path: options omitted / empty options object return the original array untouched (`totalPages: 1`, indexes spanning the whole list).
- Empty list, with and without pagination options.
- Single item, and `pageSize` larger than the total.
- Exact-multiple boundary: 10 items with `pageSize: 5` yields exactly 2 pages and page 1 is the last page (no empty trailing page).
- Middle page (`hasNextPage`/`hasPreviousPage` both true) and partially filled last page (`startIndex`/`endIndex` correct).
- Default `pageSize` of 20 when only `pageIdx` is provided.
- Invalid `pageIdx` (beyond range, negative, and beyond range on an empty list): falls back to the first page with `invalidPage: true`; `pageIdx: 0` on an empty list is valid.
- Non-positive `pageSize` values (0 and negative), documenting current behavior. The tool schemas in `console.ts`/`network.ts` reject these via `zod.number().int().positive()`, but the plain `zod.number()` schemas in `memory.ts` let them through.

## Observation (no code change)

With `pageSize: 0` and a non-empty list, `paginate` currently returns `totalPages: Infinity` with `hasNextPage: true` and no items; with a negative `pageSize`, `Array.prototype.slice` receives a negative end index and returns items from the start of the list minus `|pageSize|` from the end. Both are only reachable through the `memory.ts` tools whose zod schemas don't constrain `pageSize` to positive integers. The tests assert the current behavior; tightening those schemas could be a follow-up if desired.

## Testing

- `npm run build` — passes
- `node scripts/test.js tests/utils/pagination.test.ts` — 16/16 pass
- `npm run test:no-build` — full suite passes
- `npm run check-format` — clean

### DIFF
--- a/tests/utils/pagination.test.ts
+++ b/tests/utils/pagination.test.ts
@@ -1,0 +1,236 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'node:assert';
+import {describe, it} from 'node:test';
+
+import {paginate} from '../../src/utils/pagination.js';
+
+function getItems(count: number): string[] {
+  return Array.from({length: count}, (_, idx) => `item-${idx}`);
+}
+
+describe('paginate', () => {
+  describe('without pagination options', () => {
+    it('should return all items when options are omitted', () => {
+      const items = getItems(3);
+      const result = paginate(items);
+      assert.strictEqual(result.items, items);
+      assert.strictEqual(result.currentPage, 0);
+      assert.strictEqual(result.totalPages, 1);
+      assert.strictEqual(result.hasNextPage, false);
+      assert.strictEqual(result.hasPreviousPage, false);
+      assert.strictEqual(result.startIndex, 0);
+      assert.strictEqual(result.endIndex, 3);
+      assert.strictEqual(result.invalidPage, false);
+    });
+
+    it('should return all items when options are empty', () => {
+      const items = getItems(50);
+      const result = paginate(items, {});
+      assert.strictEqual(result.items, items);
+      assert.strictEqual(result.totalPages, 1);
+      assert.strictEqual(result.hasNextPage, false);
+      assert.strictEqual(result.endIndex, 50);
+    });
+
+    it('should handle an empty list', () => {
+      const result = paginate([]);
+      assert.deepStrictEqual(result.items, []);
+      assert.strictEqual(result.currentPage, 0);
+      assert.strictEqual(result.totalPages, 1);
+      assert.strictEqual(result.hasNextPage, false);
+      assert.strictEqual(result.hasPreviousPage, false);
+      assert.strictEqual(result.startIndex, 0);
+      assert.strictEqual(result.endIndex, 0);
+      assert.strictEqual(result.invalidPage, false);
+    });
+  });
+
+  describe('with pagination options', () => {
+    it('should handle an empty list', () => {
+      const result = paginate([], {pageSize: 5, pageIdx: 0});
+      assert.deepStrictEqual(result.items, []);
+      assert.strictEqual(result.currentPage, 0);
+      assert.strictEqual(result.totalPages, 1);
+      assert.strictEqual(result.hasNextPage, false);
+      assert.strictEqual(result.hasPreviousPage, false);
+      assert.strictEqual(result.startIndex, 0);
+      assert.strictEqual(result.endIndex, 0);
+      assert.strictEqual(result.invalidPage, false);
+    });
+
+    it('should handle a single item', () => {
+      const result = paginate(['only'], {pageSize: 5});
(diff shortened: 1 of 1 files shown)


tests/utils/pagination.test.ts
```diff
@@ -0,0 +1,236 @@
...
+    it('should handle a single item', () => {
+      const result = paginate(['only'], {pageSize: 5});
+      assert.deepStrictEqual(result.items, ['only']);
+      assert.strictEqual(result.currentPage, 0);
+      assert.strictEqual(result.totalPages, 1);
+      assert.strictEqual(result.hasNextPage, false);
+      assert.strictEqual(result.hasPreviousPage, false);
+      assert.strictEqual(result.startIndex, 0);
+      assert.strictEqual(result.endIndex, 1);
+    });
```
Lightning00Blade, line 75:
```suggestion
```

tests/utils/pagination.test.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    it('should return all items when options are empty', () => {
+      const items = getItems(50);
+      const result = paginate(items, {});
+      assert.strictEqual(result.items, items);
+      assert.strictEqual(result.totalPages, 1);
+      assert.strictEqual(result.hasNextPage, false);
+      assert.strictEqual(result.endIndex, 50);
+    });
```
Lightning00Blade, line 38: This equivalent with the above, TS makes sure of that.
```suggestion
```

tests/utils/pagination.test.ts
```diff
@@ -0,0 +1,236 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'node:assert';
+import {describe, it} from 'node:test';
+
+import {paginate} from '../../src/utils/pagination.js';
+
+function getItems(count: number): string[] {
+  return Array.from({length: count}, (_, idx) => `item-${idx}`);
```
Lightning00Blade, line 13: Maybe using non 0 for the name would be easier to check.
```suggestion
  return Array.from({length: count}, (_, idx) => `item-${idx + 1}`);
```

tests/utils/pagination.test.ts
```diff
@@ -0,0 +1,236 @@
...
+    it('should return a single page when pageSize exceeds the total', () => {
+      const items = getItems(3);
+      const result = paginate(items, {pageSize: 10, pageIdx: 0});
+      assert.deepStrictEqual(result.items, items);
+      assert.strictEqual(result.totalPages, 1);
+      assert.strictEqual(result.hasNextPage, false);
+      assert.strictEqual(result.hasPreviousPage, false);
+      assert.strictEqual(result.startIndex, 0);
+      assert.strictEqual(result.endIndex, 3);
+    });
```
Lightning00Blade, line 86:
```suggestion
```

tests/utils/pagination.test.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    it('should return the first page by default when pageIdx is omitted', () => {
+      const result = paginate(getItems(10), {pageSize: 4});
```
Lightning00Blade, line 89: Let's keep pages sized multiple of 5, so it's a little more consistent.

tests/utils/pagination.test.ts
```diff
@@ -0,0 +1,236 @@
...
+        'item-4',
+      ]);
+      assert.strictEqual(result.currentPage, 0);
+      assert.strictEqual(result.invalidPage, true);
+    });
+
+    it('should treat pageIdx 0 as valid for an empty list', () => {
+      const result = paginate([], {pageSize: 5, pageIdx: 0});
+      assert.strictEqual(result.invalidPage, false);
+    });
```
Lightning00Blade, line 197:
```suggestion
```

tests/utils/pagination.test.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    it('should flag pageIdx beyond range for an empty list', () => {
+      const result = paginate([], {pageSize: 5, pageIdx: 1});
+      assert.deepStrictEqual(result.items, []);
+      assert.strictEqual(result.currentPage, 0);
+      assert.strictEqual(result.invalidPage, true);
+    });
+  });
+
+  describe('unusual pageSize values', () => {
```
Lightning00Blade, line 207: Remove ... https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2517/changes

OrKoN:

PR #2513, opened 2026-08-08
fix: truncate long urls in concise network request output

ZayanKhan-12:
## What

Implements the existing `// TODO truncate the URL` in `src/formatters/NetworkFormatter.ts`: URLs in the concise one-line network request format are now capped at 150 characters, with the file's existing `getSizeLimitedString` helper appending the same `... <truncated>` marker already used for request/response bodies.

## Why

`list_network_requests` emits one concise line per request, so a single page with a few `data:` URLs (which can be tens of kilobytes each) can blow up the tool output by orders of magnitude. Truncating in the concise format keeps list output token-friendly while losing nothing that matters for identifying a request.

Scope is deliberately narrow — only `convertNetworkRequestConciseToString` changes:

- the detailed view (`## Request <url>` heading) still shows the full URL, and
- `toJSON()` / structured content still carries the full URL,

so nothing that needs the complete URL is affected. (Concise redirect-chain lines inside the detailed view use the same function and are truncated too, which keeps that list compact as well.)

## Limit rationale

150 characters keeps origin + path + the start of the query string intact for virtually all real-world URLs (typical URLs are well under 100 characters), while capping pathological `data:`/blob payloads. It reuses the existing `getSizeLimitedString` pattern with a new `URL_CONTEXT_SIZE_LIMIT` constant next to `BODY_CONTEXT_SIZE_LIMIT`.

## Before / after

Before:

```
reqid=12 GET data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAA... (10,000+ more chars) ...5CYII= [200]
```

After:

```
reqid=12 GET data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAA...(150 chars total)... <truncated> [200]
```

Short URLs are unchanged.

## Testing

- `npm run build` — pass
- `npm run test:no-build` — pass (all tests, exit 0)
- `npm run check-format` — pass (eslint + prettier)

Added three cases to `tests/formatters/NetworkFormatter.test.ts`:

- long URL is truncated at 150 chars with the `... <truncated>` marker (and `toJSON()` keeps the full URL),
- `data:` URL is truncated,
- a URL exactly at the limit is left untouched.

No existing snapshots changed.

### DIFF
--- a/src/formatters/NetworkFormatter.ts
+++ b/src/formatters/NetworkFormatter.ts
@@ -13,7 +13,7 @@ import {
 } from '../third_party/index.js';
 
 const BODY_CONTEXT_SIZE_LIMIT = 10000;
-const URL_CONTEXT_SIZE_LIMIT = 150;
+const URL_CONTEXT_SIZE_LIMIT = 255;
 
 export interface NetworkFormatterOptions {
   requestId?: number | string;
--- a/src/formatters/NetworkFormatter.ts
+++ b/src/formatters/NetworkFormatter.ts
@@ -13,6 +13,7 @@ import {
 } from '../third_party/index.js';
 
 const BODY_CONTEXT_SIZE_LIMIT = 10000;
+const URL_CONTEXT_SIZE_LIMIT = 150;
 
 export interface NetworkFormatterOptions {
   requestId?: number | string;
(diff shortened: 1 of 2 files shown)


src/formatters/NetworkFormatter.ts
```diff
@@ -13,6 +13,7 @@ import {
 } from '../third_party/index.js';
 
 const BODY_CONTEXT_SIZE_LIMIT = 10000;
+const URL_CONTEXT_SIZE_LIMIT = 150;
```
OrKoN:
```suggestion
const URL_CONTEXT_SIZE_LIMIT = 255;
```

Let's use a longer limit to make sure more URLs are fully visible.

OrKoN (approved):

PR #2514, opened 2026-08-08
test: add unit tests for trace processing helpers

ZayanKhan-12:
## Why

`src/trace-processing/parse.ts` exports several pure, browser-free helpers, but `tests/trace-processing/parse.test.ts` only exercised the happy parse path, the trace summary snapshot, and the `undefined`-buffer error. This PR adds direct unit tests for the remaining exported helpers and branches, so regressions in the discriminant logic or the error contracts are caught without needing a browser.

## Coverage added

- `traceResultIsSuccess`
  - returns `true` for the result of a real successful parse (fixture trace)
  - returns `true` for a `TraceResult` whose `insights` is `null` (the guard must only depend on `parsedTrace`)
  - returns `false` for the result of a failed parse
  - returns `false` for a `TraceParseError` object
  - returns `false` for a near-miss where the error *message text* is `'parsedTrace'` — the guard checks the key, not values
- `parseRawTraceBuffer` (previously untested error branches)
  - empty buffer → `'Decoding the trace buffer returned an empty string.'`
  - invalid JSON → returns a `TraceParseError` instead of throwing
  - valid JSON that is not a trace → returns a `TraceParseError` instead of throwing
- `getInsightOutput` (previously untested)
  - formatted output for a known insight (`NAVIGATION_0` / `LCPBreakdown` from the `web-dev-with-commit` fixture)
  - error when the trace has no insights
  - error for an unknown insight set id
  - error for an unknown insight name

Tests only — no changes under `src/`, and no new snapshots (the insight output is asserted with a targeted match so devtools-frontend formatter updates don't churn a snapshot).

## Testing

- `npm run build` — clean
- `node scripts/test.js tests/trace-processing/parse.test.ts` — 15/15 pass (3 suites)
- `npm run test:no-build` — full suite green (exit 0)
- `npm run check-format` — eslint + prettier clean

### DIFF
--- a/tests/trace-processing/parse.test.ts
+++ b/tests/trace-processing/parse.test.ts
@@ -7,12 +7,28 @@
 import assert from 'node:assert';
 import {describe, it} from 'node:test';
 
+import type {
+  InsightName,
+  TraceParseError,
+  TraceResult,
+} from '../../src/trace-processing/parse.js';
 import {
+  getInsightOutput,
   getTraceSummary,
   parseRawTraceBuffer,
+  traceResultIsSuccess,
 } from '../../src/trace-processing/parse.js';
 
 import {loadTraceAsBuffer} from './fixtures/load.js';
+
+async function parseTrace(fileName: string): Promise<TraceResult> {
+  const rawData = loadTraceAsBuffer(fileName);
+  const result = await parseRawTraceBuffer(rawData);
+  if (!traceResultIsSuccess(result)) {
+    assert.fail(`Unexpected trace parse error: ${result.error}`);
+  }
+  return result;
+}
 
 describe('Trace parsing', async () => {
   it('can parse a Uint8Array from Tracing.stop())', async () => {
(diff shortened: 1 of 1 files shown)


tests/trace-processing/parse.test.ts
```diff
@@ -44,4 +60,112 @@ describe('Trace parsing', async () => {
...
+  });
+});
+
+describe('getInsightOutput', () => {
+  it('returns the formatted output for a known insight', async () => {
+    const result = await parseTrace('web-dev-with-commit.json.gz');
+    const insight = getInsightOutput(result, 'NAVIGATION_0', 'LCPBreakdown');
+    if ('error' in insight) {
+      assert.fail(`Unexpected insight error: ${insight.error}`);
+    }
```
Lightning00Blade, line 130:
```suggestion
    assert(!('error' in insight`),Unexpected insight error: ${insight.error}`);
```

tests/trace-processing/parse.test.ts
```diff
@@ -44,4 +60,112 @@ describe('Trace parsing', async () => {
...
+
+  it('returns false for a parse error object', () => {
+    const error: TraceParseError = {error: 'Something went wrong.'};
+    assert.strictEqual(traceResultIsSuccess(error), false);
+  });
+
+  it('checks for the parsedTrace key, not the error message contents', () => {
+    const error: TraceParseError = {error: 'parsedTrace'};
+    assert.strictEqual(traceResultIsSuccess(error), false);
+  });
```
Lightning00Blade, line 121:
```suggestion
```

tests/trace-processing/parse.test.ts
```diff
@@ -44,4 +60,112 @@ describe('Trace parsing', async () => {
...
+
+  it('will return a message if the JSON is not a valid trace', async () => {
+    const result = await parseRawTraceBuffer(
+      new TextEncoder().encode('{"notATrace": true}'),
+    );
+    if (traceResultIsSuccess(result)) {
+      assert.fail('Expected a parse error for a non-trace JSON input.');
+    }
+    assert.ok(result.error.length > 0);
```
Lightning00Blade, line 88:
```suggestion
    assert('error' in result);
    assert.ok(result.error.length > 0);
```

tests/trace-processing/parse.test.ts
```diff
@@ -44,4 +60,112 @@ describe('Trace parsing', async () => {
...
+  });
+
+  it('will return a message if the buffer is not valid JSON', async () => {
+    const result = await parseRawTraceBuffer(
+      new TextEncoder().encode('this is not valid JSON'),
+    );
+    if (traceResultIsSuccess(result)) {
+      assert.fail('Expected a parse error for invalid JSON input.');
+    }
+    assert.match(result.error, /JSON/);
```
Lightning00Blade, line 78:
```suggestion
    assert('error' in result);
    assert.match(result.error, /JSON/);
```

tests/trace-processing/parse.test.ts
```diff
@@ -7,13 +7,29 @@
...
 } from '../../src/trace-processing/parse.js';
 
 import {loadTraceAsBuffer} from './fixtures/load.js';
 
+async function parseTrace(fileName: string): Promise<TraceResult> {
+  const rawData = loadTraceAsBuffer(fileName);
+  const result = await parseRawTraceBuffer(rawData);
+  if (!traceResultIsSuccess(result)) {
+    assert.fail(`Unexpected trace parse error: ${result.error}`);
+  }
```
Lightning00Blade, line 29:
```suggestion
  assert(traceResultIsSuccess(result), `Unexpected trace parse error: ${result.error}`);
```

tests/trace-processing/parse.test.ts
```diff
@@ -44,4 +60,112 @@ describe('Trace parsing', async () => {
...
+      new TextEncoder().encode('{"notATrace": true}'),
+    );
+    if (traceResultIsSuccess(result)) {
+      assert.fail('Expected a parse error for a non-trace JSON input.');
+    }
+    assert.ok(result.error.length > 0);
+  });
+});
+
+describe('traceResultIsSuccess', () => {
```
Lightning00Blade, line 92: For this function we should just use a object that mimics the trace, so we can control it.
Also we can't use `parseTrace` as that itself include it.

tests/trace-processing/parse.test.ts
Lightning00Blade, line 49:
```suggestion
assert(!('error' in result), `Unexpected parse failure: ${result.error}`);
```

OrKoN:

PR #2515, opened 2026-08-08
test: cover waitforhelper network multiplier and navigation paths

ZayanKhan-12:
## Why

`tests/WaitForHelper.test.ts` currently only covers dialog handling. The exported pure function `getNetworkMultiplierFromString` (`src/WaitForHelper.ts`) had no coverage at all, and the navigation/timeout paths of `waitForEventsAfterAction` were untested. This PR expands the test file to lock in the current behavior of those paths. No `src/` changes.

## Coverage added

- `getNetworkMultiplierFromString`: exact multipliers for each predefined condition (`Fast 4G` → 1, `Slow 4G` → 2.5, `Fast 3G` → 5, `Slow 3G` → 10), fallback to 1 for unknown strings and for `null`.
- Cross-document navigation: an action that navigates resolves with `navigatedToUrl` set to the destination.
- No navigation: resolves quickly (bounded well below `#stableDomTimeout`) with no `navigatedToUrl`.
- Navigation timeout: a navigation that starts but never completes has its timeout error swallowed (logged, not thrown) and still resolves with a result; the URL stays unchanged so no `navigatedToUrl` is reported.
- Same-document hash navigation: skips the full navigation wait but is still surfaced via the before/after URL comparison.
- Action errors are rethrown to the caller.

## Observation (current behavior, asserted as-is)

When a started navigation times out and remains pending, the subsequent stable-DOM wait cannot settle (in-page evaluations do not complete while the navigation is pending), so it consumes the full `#stableDomTimeout`. With `{timeout: 500}` the call still takes ~3.5s total (500ms navigation timeout + 3s stable-DOM timeout). The test documents this rather than changing it.

Note: this is independent of the open PR #2504 (popup tracking in `WaitForHelper`); it only touches network-multiplier and navigation/timeout behavior, so the two do not conflict.

## Testing

- `node scripts/test.js tests/WaitForHelper.test.ts` — 11/11 pass (run 3x to check for flakes)
- `npm run test:no-build` — full suite green
- `npm run check-format` — clean

### DIFF
--- a/tests/WaitForHelper.test.ts
+++ b/tests/WaitForHelper.test.ts
@@ -150,8 +150,6 @@ describe('WaitForHelper', () => {
         `expected the wait to be bounded by the internal timeouts, took ${elapsed}ms`,
       );
 
-      // Note: no in-page cleanup here. Evaluations do not settle while the
-      // navigation is pending; the harness closes the page during teardown.
     });
   });
 
@@ -168,8 +166,6 @@ describe('WaitForHelper', () => {
         });
       });
 
-      // Same-document navigations skip the full navigation wait but are still
-      // surfaced through the before/after URL comparison.
       assert.strictEqual(result.navigatedToUrl, `${pageUrl}#section`);
       assert.strictEqual(result.dialogHandled, false);
     });
--- a/tests/WaitForHelper.test.ts
+++ b/tests/WaitForHelper.test.ts
@@ -7,9 +7,33 @@
 import assert from 'node:assert';
 import {describe, it} from 'node:test';
 
+import {getNetworkMultiplierFromString} from '../src/WaitForHelper.js';
+
+import {serverHooks} from './server.js';
 import {html, withMcpContext} from './utils.js';
 
+describe('getNetworkMultiplierFromString', () => {
+  it('maps each predefined network condition to its multiplier', () => {
+    assert.strictEqual(getNetworkMultiplierFromString('Fast 4G'), 1);
+    assert.strictEqual(getNetworkMultiplierFromString('Slow 4G'), 2.5);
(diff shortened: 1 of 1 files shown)


tests/WaitForHelper.test.ts
```diff
@@ -43,4 +67,125 @@ describe('WaitForHelper', () => {
...
+      // the full 3s #stableDomTimeout, because the stable-DOM evaluation does
+      // not settle while the navigation is still pending (~3.5s in total).
+      // Assert an upper bound well below protocolTimeout-scale hangs.
+      assert.ok(
+        elapsed < 8_000,
+        `expected the wait to be bounded by the internal timeouts, took ${elapsed}ms`,
+      );
+
+      // Note: no in-page cleanup here. Evaluations do not settle while the
+      // navigation is pending; the harness closes the page during teardown.
```
Lightning00Blade:
```suggestion
```

tests/WaitForHelper.test.ts
```diff
@@ -43,4 +67,125 @@ describe('WaitForHelper', () => {
...
+      await mcpPage.pptrPage.goto(pageUrl);
+
+      const result = await mcpPage.waitForEventsAfterAction(async () => {
+        await mcpPage.pptrPage.evaluate(() => {
+          window.location.hash = '#section';
+        });
+      });
+
+      // Same-document navigations skip the full navigation wait but are still
+      // surfaced through the before/after URL comparison.
```
Lightning00Blade:
```suggestion
```

Lightning00Blade (approved):
Thanks

OrKoN:
closing as stale

PR #2518, opened 2026-08-08
test: add unit tests for logger and update check utils

ZayanKhan-12:
## Why

`src/utils/logger.ts` and `src/utils/check-for-updates.ts` had no unit test coverage in `tests/utils/`. Both contain observable behavior (file log sink, namespace enabling, flush timeout, update cache freshness/once-per-process logic) that can regress silently, so this adds focused tests asserting the current behavior. No `src/` changes.

## What's covered

**`tests/utils/logger.test.ts` (6 tests)**
- `saveLogsToFile` writes `mcp:log` output to the given file, appends (flag `a+`) instead of truncating, writes one newline-terminated line per log call, and enables the `mcp:log` debug namespace even when previously disabled.
- `flushLogs` resolves once the stream has finished, and rejects when the stream does not finish within the timeout (via a stub stream whose `end` callback never fires).
- Global `debug` state (enabled namespaces, `debug.log` sink) is saved and restored around the suite.

**`tests/utils/check-for-updates.test.ts` (8 tests)**
- Early return when `CHROME_DEVTOOLS_MCP_NO_UPDATE_CHECKS` is set (the test runner sets this globally, so the tests save/clear/restore it).
- First run with no cache: writes `{version: VERSION}` to `~/.cache/chrome-devtools-mcp/latest.json` and spawns `check-latest-version.js` detached with `stdio: 'ignore'`.
- Warns with `Update available: <current> -> <cached>` plus the caller-provided message when the cached version is newer; no warning when equal.
- Fresh cache (<24h): no re-check spawn, cache untouched. Stale cache (>24h): re-check spawn, mtime refreshed, content preserved.
- Invalid-JSON cache is ignored without throwing.
- Once-per-process guard, including `resetUpdateCheckFlagForTesting()`.
- Tests redirect `os.homedir()` to a per-test temp dir and stub `child_process.spawn`/`console.warn` via `node:test`'s `mock.method`, so no real home-directory files are touched and no subprocess/network activity happens.

## What's not covered (and why)

- `saveLogsToFile`'s stream `error` handler calls `process.exit(1)`, which cannot be exercised in-process without killing the test runner.
- The spawned `bin/check-latest-version.js` subprocess itself (registry fetch) is out of scope for these unit tests; `spawn` is stubbed.
- Observation (not asserted, no behavior change made): while logs are redirected to a file, `debug.log` is replaced with a plain `chunks.join(' ')` sink, so printf-style placeholders (e.g. `logger('hello %s', 'world')`) are written literally as `hello %s world` instead of being substituted the way the default `util.formatWithOptions`-based sink does. The tests use plain messages to assert current behavior; happy to file a follow-up issue if that quirk is worth fixing.

## Testing

- `npm run build` — clean.
- `node scripts/test.js tests/utils/logger.test.ts tests/utils/check-for-updates.test.ts` — 14/14 pass.
- `npm run test:no-build` — 702 tests, 702 pass, 0 fail.
- `npm run check-format` — eslint + prettier clean.

### DIFF
--- a/tests/utils/check-for-updates.test.ts
+++ b/tests/utils/check-for-updates.test.ts
@@ -1,0 +1,170 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'node:assert';
+import childProcess from 'node:child_process';
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+import process from 'node:process';
+import {afterEach, beforeEach, describe, it, mock} from 'node:test';
+
+import {
+  checkForUpdates,
+  resetUpdateCheckFlagForTesting,
+} from '../../src/utils/check-for-updates.js';
+import {VERSION} from '../../src/version.js';
+
+const NO_UPDATE_CHECKS_ENV = 'CHROME_DEVTOOLS_MCP_NO_UPDATE_CHECKS';
+const NEWER_VERSION = '999.999.999';
+
+describe('checkForUpdates', () => {
+  let tmpHome: string;
+  let savedEnv: string | undefined;
+  let warnings: string[];
+  let spawnCalls: Array<{
+    command: string;
+    args: string[];
+    options: childProcess.SpawnOptions;
+  }>;
+
+  function cachePath(): string {
+    return path.join(tmpHome, '.cache', 'chrome-devtools-mcp', 'latest.json');
+  }
+
+  async function writeCache(content: string): Promise<void> {
+    await fs.mkdir(path.dirname(cachePath()), {recursive: true});
+    await fs.writeFile(cachePath(), content);
+  }
+
+  beforeEach(async () => {
+    tmpHome = await fs.mkdtemp(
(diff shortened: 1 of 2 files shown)


tests/utils/logger.test.ts
```diff
@@ -0,0 +1,113 @@
...
+      assert.ok(logger);
+      logger('hello world');
+      await flushLogs(stream);
+      const content = await fsPromises.readFile(file, 'utf8');
+      assert.ok(content.includes('mcp:log'));
+      assert.ok(content.includes('hello world'));
+      assert.ok(content.endsWith('\n'));
+    });
+
+    it('enables the mcp:log debug namespace', async () => {
```
Lightning00Blade, line 53: This is a side-effect and we should not test for it.

tests/utils/check-for-updates.test.ts
```diff
@@ -0,0 +1,170 @@
+/**
```
Lightning00Blade, line 1: There is test in `tests/check-for-updates.test.ts` Why do we need this?

Lightning00Blade (changes requested):

OrKoN:

PR #2520, opened 2026-08-08
test: expand mcp response composition coverage

ZayanKhan-12:
## Why

`src/McpResponse.ts` is the single point where every tool response is composed (text sections, images, structured content, pagination), but several composition behaviors had no test coverage anywhere in the repo. This PR adds 11 focused unit tests to `tests/McpResponse.test.ts` asserting the current behavior, so future refactors of `format()`/`handle()` can't silently change the response contract. No `src/` changes.

## Gap analysis

Already covered by the existing suite (left untouched): page listing, snapshots (plain/verbose/file/stable ids), emulation settings, dialogs, single image, network include/exclude/attach/resource-type filtering, network pagination, basic console include/empty/issue paths, trace summary/insights, extensions, lighthouse, third-party developer tools, WebMCP.

Not covered anywhere (verified by grepping the whole `tests/` tree), now added:

- **Error propagation**: `setError()` → `Error: <message>` text line and `structuredContent.errorMessage` (the `ToolHandler` error path relies on this; it was never asserted).
- **Tab id**: `setTabId()` → `structuredContent.tabId`.
- **Image attachment ordering**: multiple images are appended after the text block in attachment order with their mime types preserved.
- **Include-option resets**: `setIncludeNetworkRequests(true, …)`/`setIncludeConsoleData(true, …)` followed by `(false)` clears the options (no section, no structured data, getters reset).
- **Section ordering**: response lines → `## Network requests` → `## Console messages` → error message last.
- **Console message pagination**: `pageSize`/`pageIdx` header lines (`Showing 3-4 of 5 (Page 2 of 3).`, next/previous hints), paginated `structuredContent.consoleMessages`, structured pagination fields, and invalid-page fallback. (Network pagination was tested extensively; console pagination was not.)
- **Missing-page guards**: `handle()` rejects with `Response must have a page` / `Response must have an McpPage` when a snapshot or network request is requested on a response without a page.

Observation (documented by a test, not changed): `structuredContent.pagination` is a single shared field, so when both network requests and console messages are paginated in one response, it holds the values of the section formatted last (console). Each section's own text output still reports its correct pagination. If per-section structured pagination is preferred, that would be a behavior change for a separate PR.

The new tests use explicit assertions rather than snapshots, matching the style of the existing pagination/reconnect tests; no snapshot file changes are needed.

## Testing

- `node scripts/test.js tests/McpResponse.test.ts` → 65 tests, 65 pass, 0 fail (54 existing + 11 new)
- `npm run test:no-build` → full suite green (exit 0)

### DIFF
--- a/tests/McpResponse.test.ts
+++ b/tests/McpResponse.test.ts
@@ -14,7 +14,7 @@ import sinon from 'sinon';
 
 import type {ParsedArguments} from '../src/bin/chrome-devtools-mcp-cli-options.js';
 import type {McpContext} from '../src/McpContext.js';
-import type {McpResponse} from '../src/McpResponse.js';
+import {McpResponse} from '../src/McpResponse.js';
 import type {Extension} from '../src/third_party/index.js';
 import {
   closePage,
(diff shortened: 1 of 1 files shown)


OrKoN (changes requested):
Thanks for the PR. I think the e2e-like tests for the composition would be hard to maintain and longer to run. I think we should find a way to test composition with a unit-like test.

OrKoN:

PR #2522, opened 2026-08-08
test: expand heap snapshot formatter and manager coverage

ZayanKhan-12:
## Why

`tests/HeapSnapshotManager.test.ts` had a single test while `HeapSnapshotManager` exposes the aggregation/stats/static-data/native-context API the memory tools are built on, and the exported `isNodeLike`/`isEdgeLike` type guards in `src/formatters/HeapSnapshotFormatter.ts` had no coverage at all. This adds 346 lines of tests only — no `src/` changes.

## Coverage

**HeapSnapshotManager** (against the existing example fixture, reusing the file's harness):
- snapshot caching by resolved path; lifecycle tracking + dispose; class-key lookups rejected for unloaded snapshots
- `getAggregates`: class totals, stable class ids that resolve back to class keys, `objectsRetainedByContexts` filter, `attributedToSpecificNativeContext` without objectId rejected
- `getStats`, `getStaticData`, `getNativeContextSizes`

**HeapSnapshotFormatter guards** (pure):
- `isNodeLike`: valid minimal/full node shapes, null/undefined, primitives, empty object, missing id/name, edge-shaped near-miss
- `isEdgeLike`: valid edge with node-like node, null/undefined, empty object, each missing top-level property

## Testing

- `node scripts/test.js tests/formatters/HeapSnapshotFormatter.test.ts tests/HeapSnapshotManager.test.ts` → pass (exit 0)
- `npm run test:no-build` → full suite pass (exit 0)

### DIFF
--- a/tests/HeapSnapshotManager.test.ts
+++ b/tests/HeapSnapshotManager.test.ts
@@ -5,12 +5,14 @@
  */
 
 import assert from 'node:assert';
-import {describe, it, afterEach} from 'node:test';
+import {join} from 'node:path';
+import {describe, it, after, afterEach} from 'node:test';
 
 import sinon from 'sinon';
 
 import {HeapSnapshotManager} from '../src/HeapSnapshotManager.js';
 import {DevTools} from '../src/third_party/index.js';
+import {stableIdSymbol} from '../src/utils/id.js';
 
 describe('HeapSnapshotManager', () => {
   afterEach(() => {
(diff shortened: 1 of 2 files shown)


tests/formatters/HeapSnapshotFormatter.test.ts
```diff
@@ -130,6 +245,50 @@ describe('HeapSnapshotFormatter', () => {
...
+        `1,NodeA,object,2,${formatBytesToKb(1024)},${formatBytesToKb(2048)}`,
+        `2,NodeB,closure,3,${formatBytesToKb(256)},${formatBytesToKb(512)}`,
+      ].join('\n');
+
+      assert.strictEqual(result, expected);
+    });
+
+    it('returns an empty string for an empty array', () => {
+      assert.strictEqual(HeapSnapshotFormatter.formatNodes([]), '');
+    });
```
Lightning00Blade, line 291:
```suggestion
```

tests/HeapSnapshotManager.test.ts
```diff
@@ -36,4 +38,187 @@ describe('HeapSnapshotManager', () => {
...
+        assert.ok(typeof id === 'number' && id > 0);
+        ids.add(id);
+      }
+
+      assert.ok(data.objectCount > 0);
+      assert.ok(data.totalSelfSize > 0);
+      assert.strictEqual(data.objectCount, objectCount);
+      assert.strictEqual(data.totalSelfSize, totalSelfSize);
+      // Every class gets a distinct id.
+      assert.strictEqual(ids.size, aggregates.length);
```
Lightning00Blade, line 85: This is not need it handled by line 76

tests/HeapSnapshotManager.test.ts
```diff
@@ -36,4 +38,187 @@ describe('HeapSnapshotManager', () => {
...
+
+    it('aggregates classes with totals and stable ids', async () => {
+      const data = await manager.getAggregates(fixturePath);
+      const aggregates = Object.values(data.aggregates);
+      assert.ok(aggregates.length > 0);
+
+      let objectCount = 0;
+      let totalSelfSize = 0;
+      const ids = new Set<number>();
+      for (const aggregate of aggregates) {
```
Lightning00Blade, line 72: This is the same code that we have, which means we are not testing anything.
We should assert at hard coded values as the fixture is hard coded as well.

tests/HeapSnapshotManager.test.ts
```diff
@@ -36,4 +38,187 @@ describe('HeapSnapshotManager', () => {
...
+
+      const arrayAggregate = aggregates.find(
+        aggregate => aggregate.name === 'Array',
+      );
+      assert.ok(arrayAggregate);
+      assert.strictEqual(arrayAggregate.count, 8);
+    });
+
+    it('returns stable class ids that resolve back to class keys', async () => {
+      const first = await manager.getAggregates(fixturePath);
```
Lightning00Blade, line 95: We have a test that showcases that the two are the same so we can just use one.

tests/HeapSnapshotManager.test.ts
Lightning00Blade, line 72: This is the same code that we have, which means we are not testing anything.
We should assert at hard coded values as the fixture is hard coded as well.

tests/HeapSnapshotManager.test.ts
```diff
@@ -36,4 +38,187 @@ describe('HeapSnapshotManager', () => {
...
+      assert.ok(staticData);
+      assert.strictEqual(staticData.nodeCount, 27466);
+      assert.strictEqual(staticData.rootNodeIndex, 0);
+      assert.strictEqual(staticData.maxJSObjectId, 54005);
+
+      const stats = await manager.getStats(fixturePath);
+      assert.strictEqual(staticData.totalSize, stats.total);
+    });
+
+    it('computes native context sizes', async () => {
```
Lightning00Blade, line 168: We hand this off to DevTools so this should not be tested here.

tests/HeapSnapshotManager.test.ts
```diff
@@ -36,4 +38,187 @@ describe('HeapSnapshotManager', () => {
...
+        stats.v8heap.code,
+        stats.v8heap.jsArrays,
+        stats.v8heap.strings,
+        stats.v8heap.system,
+      ]) {
+        assert.ok(typeof value === 'number' && value >= 0);
+      }
+    });
+
+    it('exposes static data once the snapshot is loaded', async () => {
```
Lightning00Blade, line 157: We hand this off to DevTools so this should not be tested here.

tests/HeapSnapshotManager.test.ts
```diff
@@ -36,4 +38,187 @@ describe('HeapSnapshotManager', () => {
...
+        manager.getAggregates(fixturePath, 'attributedToSpecificNativeContext'),
+        {
+          message:
+            'objectId is required when filterName is attributedToSpecificNativeContext',
+        },
+      );
+    });
+
+    it('reports snapshot statistics', async () => {
+      const stats = await manager.getStats(fixturePath);
```
Lightning00Blade, line 142: We hand this off to DevTools so this should not be tested here.

OrKoN:

PR #2523, opened 2026-08-08
test: cover symbolized error message extraction

ZayanKhan-12:
## Why

`SymbolizedError.fromDetails`/`fromError` in `src/devtools/DevtoolsUtils.ts` implement the message-extraction rules used to surface page exceptions to agent users, but were only exercised indirectly through one browser-based test. This adds direct coverage of the extraction permutations (222 added lines, tests only — no `src/` changes).

## Coverage

- `fromDetails`: message taken from the exception description up to the first stack frame; full description kept when it has no frames; multi-line messages preserved; exception with only a value → empty message; no exception object → details text used verbatim (including non-"Uncaught" text); url/line/column/raw stack trace ignored for the message; resolved stack trace and cause retained
- `fromError`: message from the error description; full description without frames; empty-message fallback when no description exists

## Testing

- `node scripts/test.js tests/devtools/DevtoolsUtils.test.ts` → pass (exit 0)
- `npm run test:no-build` → full suite pass (exit 0)

### DIFF
--- a/tests/devtools/DevtoolsUtils.test.ts
+++ b/tests/devtools/DevtoolsUtils.test.ts
@@ -9,8 +9,12 @@ import {afterEach, describe, it} from 'node:test';
 
 import sinon from 'sinon';
 
-import {createTargetUniverse} from '../../src/devtools/DevtoolsUtils.js';
+import {
+  createTargetUniverse,
+  SymbolizedError,
+} from '../../src/devtools/DevtoolsUtils.js';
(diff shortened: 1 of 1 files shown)


szuend (approved):

szuend:
Drive-by comment @OrKoN: DevTools by now has it's own `SymbolizedError` that handles cause chaining, stack trace translation and Error.stack parsing that the MCP server could reuse. Same as stack traces, the entrypoint is `DebuggerWorkspaceBinding`.

PR #2527, opened 2026-08-10
ci: wait for package to be on npm before publishing to MCP

OrKoN:
We need a separate workflow and, thus, cannot wait for the npm workflow directly. This adds a waiter based on the registry status and will only attempt publishing to the MCP registry if npm package is available.

### DIFF
--- a/.github/workflows/publish-to-mcp-registry-on-tag.yml
+++ b/.github/workflows/publish-to-mcp-registry-on-tag.yml
@@ -71,14 +71,14 @@ jobs:
           PKG=$(node -p "require('./package.json').name + '@' + require('./package.json').version")
           echo "Waiting for $PKG to be available on npm..."
           for i in {1..5}; do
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-to-mcp-registry-on-tag.yml
```diff
@@ -66,6 +66,25 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Wait for npm package availability
+        run: |
+          PKG=$(node -p "require('./package.json').name + '@' + require('./package.json').version")
+          echo "Waiting for $PKG to be available on npm..."
+          for i in {1..5}; do
+            if npm view "$PKG" version > /dev/null 2>&1; then
```
zyzyzyryxy: Consider adding `--prefer-online` to mane sure offline status is not cached.

zyzyzyryxy (approved):

PR #2529, opened 2026-08-10
feat: Show self and retained sizes for edges

dinfuehr:
This required explicit tool calls so far but is pretty cheap to provide directly.


Lightning00Blade (approved):
LGTM

dinfuehr:
Thanks for the quick review!

PR #2536, opened 2026-08-11
fix: respect screenshot bounds on HiDPI displays

### LINKED ISSUE #2531: `--screenshotMaxWidth` returns images at twice the bound on a HiDPI display
### Description of the bug

`--screenshotMaxWidth` caps the image before the device scale factor is applied, so on a 2x display the returned screenshot is twice the requested width. Asking for 800 gives back 1600, four times the area.

That matters most where the flag came from: #1823 added it to keep screenshots under the 8000 pixel limit behind #879, and a cap picked to stay under that would still exceed it by the same factor.

The overshoot is also what gets billed. Anthropic prices an image at `ceil(width/28) * ceil(height/28)` visual tokens, so the 800x433 the flag should return costs 464 and the 1600x865 it does return costs 1798 – 3.9x on every capture, and still over 3x on a model whose long edge cap downscales the wider one.

### Reproduction

Start Chrome with a given scale factor and connect to it:

```
"Google Chrome for Testing" --remote-debugging-port=9333 --headless=new \
  --window-size=1400,900 --force-device-scale-factor=2 https://example.com/

npx chrome-devtools-mcp@1.7.0 --browserUrl http://localhost:9333 --screenshotMaxWidth 800
```

Then call `take_screenshot` and read the PNG header. The bound holds exactly at a scale factor of 1 and is out by the factor above it:

```
DPR 1    no bound 1400x757     --screenshotMaxWidth 800 ->   800x433
DPR 2    no bound 2800x1514    --screenshotMaxWidth 800 ->  1600x865
```

### Expectation

An image no wider than the requested maximum, since the flag is there to bound what reaches the model.

### MCP configuration

```json
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": ["-y", "chrome-devtools-mcp@1.7.0", "--browserUrl", "http://localhost:9333", "--screenshotMaxWidth", "800"]
    }
  }
}
```

1.7.0

### Chrome version

151.0.7922.71 (Chrome for Testing)

### Coding agent version

Claude Code 2.1.92

### Model version

Claude Opus 5

### Node version

v24.19.0

### Operating system

macOS 26.5.1

shixi-li:
Fixes #2531

## What changed

- factor the page's actual `window.devicePixelRatio` into screenshot downscaling
- keep CDP clip dimensions in CSS pixels while bounding the resulting bitmap dimensions
- add a regression test for the real `defaultViewport: null` path with a forced 2x device scale factor

## Why

Screenshot source boxes and CDP clip dimensions are expressed in CSS pixels, but the returned bitmap is scaled by the page's device pixel ratio. The previous calculation compared the CSS width directly with `screenshotMaxWidth`, so a 2x page could return an image twice the configured bound.

The new calculation derives the clip scale from `box × devicePixelRatio`, making the limit apply to the image pixels sent to the model.

## Validation

- `npm run check-format`
- `npm run build`
- `npm run test -- tests/tools/screenshot.test.ts --test-name-pattern='honors screenshotMaxWidth at device scale factors above 1'`
- `npm run test -- tests/tools/screenshot.test.ts --test-skip-pattern='with full page resulting in a large screenshot'`

The new regression fails on `main` with a 200px-wide image and passes with an exact 100px result after this change. I also attempted the full suite with retries; this local environment still times out in unrelated daemon/extension E2E tests and hits the existing `Page.captureScreenshot: Page is too large` case, which reproduces on pristine `main`.

### DIFF
--- a/src/tools/screenshot.ts
+++ b/src/tools/screenshot.ts
@@ -63,17 +63,25 @@ function computeDownscaleClip(
   box: BoundingBox,
   maxWidth: number | undefined,
   maxHeight: number | undefined,
+  devicePixelRatio: number,
 ): ScreenshotClip | undefined {
   const widthScale =
-    maxWidth !== undefined ? Math.min(1, maxWidth / box.width) : 1;
+    maxWidth !== undefined
+      ? Math.min(1, maxWidth / (box.width * devicePixelRatio))
+      : 1;
   const heightScale =
-    maxHeight !== undefined ? Math.min(1, maxHeight / box.height) : 1;
+    maxHeight !== undefined
+      ? Math.min(1, maxHeight / (box.height * devicePixelRatio))
+      : 1;
   const scale = Math.min(widthScale, heightScale);
   if (scale >= 1) {
     return undefined;
   }
   // Skip degenerate sub-pixel results.
-  if (Math.round(box.width * scale) < 1 || Math.round(box.height * scale) < 1) {
+  if (
+    Math.round(box.width * devicePixelRatio * scale) < 1 ||
+    Math.round(box.height * devicePixelRatio * scale) < 1
+  ) {
     return undefined;
   }
   return {
@@ -174,10 +182,14 @@ export const screenshot = definePageTool(args => {
       ) {
         const box = await getSourceBox(page, element, fullPage);
         if (box) {
+          const devicePixelRatio = await page.evaluate(
+            () => window.devicePixelRatio,
+          );
           clip = computeDownscaleClip(
             box,
             screenshotMaxWidth,
             screenshotMaxHeight,
+            devicePixelRatio,
           );
         }
       }
--- a/src/tools/screenshot.ts
+++ b/src/tools/screenshot.ts
@@ -17,14 +17,24 @@ import {definePageTool} from './ToolDefinition.js';
 
 type ScreenshotFormat = 'png' | 'jpeg' | 'webp';
 
+type SourceBox = BoundingBox & {
+  devicePixelRatio: number;
+};
+
 async function getSourceBox(
   page: Page,
   element: ElementHandle | undefined,
   fullPage: boolean,
-): Promise<BoundingBox | undefined> {
+): Promise<SourceBox | undefined> {
   if (element) {
-    const box = await element.boundingBox();
-    return box ?? undefined;
+    const viewport = page.viewport();
+    const [box, devicePixelRatio] = await Promise.all([
+      element.boundingBox(),
+      viewport
+        ? (viewport.deviceScaleFactor ?? 1)
+        : page.evaluate(() => window.devicePixelRatio),
+    ]);
+    return box ? {...box, devicePixelRatio} : undefined;
   }
   if (fullPage) {
     const dims = await page.evaluate(() => ({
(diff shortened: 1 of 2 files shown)


shixi-li:
Rebased again onto the latest `main` (`0b4390d`); the current head is `c801c83`. `git range-diff` reports the PR commit as unchanged. The two new upstream commits include the repository-wide `verifyFilesSchema` migration, which touches `src/tools/screenshot.ts` in a separate hunk; the rebase preserved both that new object-schema form and the HiDPI calculation without conflict. The introduced HiDPI regression test and the related no-viewport downscaling test pass on the current head. For transparency, the full screenshot file still hits `Page.captureScreenshot: Page is too large` in the pre-existing large full-page case; the identical command reproduced the identical failure on a detached clean `origin/main`, so no unrelated code change was made. The `Memory leak tests`, `Check code before submitting`, and `Compile and run tests` workflows are again `action_required` before creating jobs. Could a maintainer approve those runs and review the rebased PR when convenient?

thomasbachem:
Diff looks right to me. The emitted bitmap is `box * scale * dpr`, so folding the ratio into the bound is the right place to fix it, and at a ratio of 1 the calculation is unchanged.

I pulled both versions of `computeDownscaleClip()` out and ran them over a matrix of box sizes, ratios and bounds. Nothing came back above the bound, nothing `main` clips stopped being clipped, and the sub-pixel guard fires on identical inputs either way since the ratio cancels out of it.

The new test also can't pass on `main`, which returns 200 where it asserts 100.

Still behind `main`, and the test jobs need a maintainer to approve the run.

shixi-li:
Thanks for the independent validation. I rebased the unchanged patch onto the current `main@fadbf41d`; the new head is `5c93e0e`, and `git range-diff` reports the commit as equivalent. `npm run build` and `git diff --check` pass. After syncing the locked dependencies, the two focused browser tests still stop before their assertions because local Chrome/Puppeteer closes during `Target.setDiscoverTargets`; I did not change the patch to mask that environment-level launch failure. The title, change scan, and CLA checks are green; the full test jobs still need maintainer approval when convenient.

thomasbachem:
Re-checked the rebase. The patch at `5c93e0e` is identical to the one I reviewed down to the context lines, with one hunk header shifted by two, so my review above still stands.

Worth knowing before those runs get approved: `Compile and run tests` has been red on `main` for its last 11 runs, the last green one being `0b4390d` on the 13th. Every failure in that streak is a Windows or macOS cell, and the failing test moves around between them, mostly timing and navigation cases. No Ubuntu cell has failed once. So one of those cells going red here would not say much about the patch.

src/tools/screenshot.ts
```diff
@@ -174,10 +182,14 @@ export const screenshot = definePageTool(args => {
       ) {
         const box = await getSourceBox(page, element, fullPage);
         if (box) {
+          const devicePixelRatio = await page.evaluate(
+            () => window.devicePixelRatio,
```
OrKoN: I think we should fetch it in getSourceBox and potentially return the source box in device pixels as it is only used for this. This allows us to save a CDP roundtrip.

shixi-li: Implemented in `90e02e9`. `getSourceBox` now returns the source dimensions together with the effective device scale factor. Emulated viewports reuse `viewport.deviceScaleFactor`; full-page and `defaultViewport: null` paths collect dimensions and `window.devicePixelRatio` in the same page evaluation. The element path reuses the emulated scale factor when available and otherwise fetches it in parallel with the bounding box.

This removes the extra sequential page-evaluation roundtrip while keeping CDP clip coordinates in CSS pixels. `npm run build` and `git diff --check` pass. The screenshot test file is still blocked before assertions by the previously reported local Chrome `Target.setDiscoverTargets: Target closed` launch failure.

OrKoN (approved):
LGTM thanks!

PR #2541, opened 2026-08-11
fix: flakiness in wait for helper

OrKoN:
Start 100ms timer only after the action is done to avoid the race with the action code itself.

### DIFF
--- a/src/WaitForHelper.ts
+++ b/src/WaitForHelper.ts
@@ -165,14 +165,13 @@ export class WaitForHelper {
     const navigationListener = (
       event: Protocol.Page.FrameStartedNavigatingEvent,
     ) => {
+      if (event.frameId !== this.#page.mainFrame()._id) {
+        return;
+      }
       if (
-        [
-          'historySameDocument',
-          'historyDifferentDocument',
-          'sameDocument',
-        ].includes(event.navigationType)
+        event.navigationType === 'sameDocument' ||
+        event.navigationType === 'historySameDocument'
       ) {
-        navigationStartedResolvers.resolve(false);
         return;
       }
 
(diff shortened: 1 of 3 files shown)


src/WaitForHelper.ts
```diff
@@ -206,6 +201,14 @@ export class WaitForHelper {
       throw error;
     }
 
+    await Promise.race([
+      navigationStartedResolvers.promise,
+      this.timeout(this.#expectNavigationIn).then(() => {
+        navigationStartedResolvers.resolve(false);
+      }),
+    ]);
+    this.#page._client().off('Page.frameStartedNavigating', navigationListener);
```
Lightning00Blade: We should always call the abort controller when we exit so we can remove this.
```suggestion
```

Lightning00Blade (approved):

OrKoN:
@Lightning00Blade please review again as I made more changes due to more flakiness.

src/WaitForHelper.ts
```diff
@@ -186,15 +154,56 @@ export class WaitForHelper {
...
-      .then(navigationStated => {
-        if (navigationStated) {
-          return this.#page.waitForNavigation({
-            timeout: options?.timeout ?? this.#navigationTimeout,
-            signal: this.#abortController.signal,
-          });
-        }
+    // A scoped AbortController used to remove listeners associated with the
+    // navigation check once the race between navigation detection and
+    // expectNavigationIn completes, avoiding duplicate off() calls and listener leaks.
```
Lightning00Blade: The text here is a little bit confusing, it can be read as if calling `off` multiple times creates memory leaks.

Lightning00Blade (approved):

OrKoN:
@Lightning00Blade sorry I found more issues. Could you please review again to see if any potential regressions stand out?

tests/WaitForHelper.test.ts
```diff
@@ -19,28 +22,135 @@ describe('WaitForHelper', () => {
...
+          src="${iframeSrc}"
+        ></iframe>`,
+      );
+
+      const startTime = Date.now();
+      const result = await mcpPage.waitForEventsAfterAction(
+        async () => {
+          await mcpPage.pptrPage.evaluate(targetUrl => {
+            const frame = document.querySelector('iframe');
+            if (frame) {
```
Lightning00Blade: This should throw if not found

Lightning00Blade (approved):

PR #2547, opened 2026-08-11
fix: reject unknown MCP server options

### LINKED ISSUE #2530: Unknown CLI flags are silently ignored, so a typo'd `--browserUrl` launches a browser instead of connecting
### Description of the bug

The MCP server accepts unrecognized CLI flags without an error or a warning, so a single character typo in `--browserUrl` is silent and the server falls back to launching its own browser. The agent then works against an empty browser while its configuration says it is attached to an existing one.

The CLI's `start` command launches the same server under `.strict()`, with `--browserUrl` as its own documented example, so the two entry points disagree about the same flag.

### Reproduction

Start any Chrome with `--remote-debugging-port=9333` on a page you can recognize, then run the server twice and call `list_pages`:

```
npx chrome-devtools-mcp@1.7.0 --browserUrl http://localhost:9333
npx chrome-devtools-mcp@1.7.0 --browserURL http://localhost:9333
```

```
--browserUrl  ->  1: Example Domain (https://example.com/) [selected]
--browserURL  ->  1: about:blank [selected]
```

In the second case `evaluate_script` reports `about:blank` with `navigator.webdriver === true`, so it is a freshly launched browser, not the one on `:9333`, where the same probe reports `false`.

`--browser-url` works, so only genuine misspellings are affected. For comparison, `@playwright/mcp` answers `error: unknown option '--nonsenseFlagXyz'`.

### Expectation

An unrecognized flag fails fast, the way it already does through `start`.

One wrinkle: `.strict()` on its own is not a drop-in. Yargs camelizes mixed forms such as `--category-experimentalWebmcp` into the declared option, so they work today and do enable the category, but strict validates the literal token and rejects them. `tests/index.test.ts` builds its category flags from the camelCase enum values, so `.strict()` alone fails the suite there.

### MCP configuration

```json
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": ["-y", "chrome-devtools-mcp@1.7.0", "--browserURL", "http://localhost:9333"]
    }
  }
}
```

1.7.0

### Chrome version

151.0.7922.71 (Chrome for Testing)

### Coding agent version

Claude Code 2.1.92

### Model version

Claude Opus 5

### Node version

v24.19.0

### Operating system

macOS 26.5.1

Boulea7:
## Summary

- reject unknown MCP server options instead of silently ignoring them
- preserve accepted mixed camel/kebab option forms by registering aliases before strict validation

## Testing

- `npm run test -- tests/cli.test.ts`
- `npm run test`
- `npm run check-format`

Fixes #2530

AI assistance: This change was implemented and verified with OpenAI Codex.

BEGIN_COMMIT_OVERRIDE
chore: reject unknown MCP server options

### DIFF
--- a/src/bin/chrome-devtools-mcp-cli-options.ts
+++ b/src/bin/chrome-devtools-mcp-cli-options.ts
@@ -381,9 +381,23 @@ export function parseArguments(
   argv = process.argv,
   env = process.env,
 ) {
+  // Preserve yargs' mixed camel/kebab-case expansion under strict validation.
+  const kebabCaseAliases: Record<string, string> = {};
+  for (const option of Object.keys(cliOptions)) {
+    const alias = option.replace(
+      /[A-Z]/g,
+      letter => `-${letter.toLowerCase()}`,
+    );
+    if (alias !== option) {
+      kebabCaseAliases[option] = alias;
+    }
+  }
+
   const yargsInstance = yargs(hideBin(argv))
     .scriptName('npx chrome-devtools-mcp@latest')
     .options(cliOptions)
+    .alias(kebabCaseAliases)
+    .strictOptions()
     .middleware(args => {
       // We can't set default in the options else
       // Yargs will complain
--- a/tests/cli.test.ts
+++ b/tests/cli.test.ts
@@ -5,6 +5,7 @@
  */
 
 import assert from 'node:assert';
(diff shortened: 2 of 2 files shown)


Lightning00Blade (approved):
Thanks for the PR!

PR #2552, opened 2026-08-11
fix: fix formatter bugs

nattallius:
This PR is created on top of PR #2519 and addresses two formatter bugs discovered during review:

ConsoleFormatter bug: When a console log has an empty message text but contains arguments, the formatter correctly shifts the first argument out of the arguments list to serve as the message text. However, it never sets the text field to this shifted argument, causing the first argument to be completely lost in the output.

IssueFormatter bug: When a concise issue has no description, the formatting helper prints the literal string "undefined" instead of falling back to "Unknown Issue". 

Closes #2519

### DIFF
--- a/src/formatters/ConsoleFormatter.ts
+++ b/src/formatters/ConsoleFormatter.ts
@@ -186,16 +186,6 @@ export class ConsoleFormatter {
     });
   }
 
-  get text(): string {
-    if (this.#text) {
-      return this.#text;
-    }
-    if (this.#resolvedArgs.length > 0) {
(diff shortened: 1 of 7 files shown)


src/formatters/ConsoleFormatter.ts
```diff
@@ -186,6 +186,16 @@ export class ConsoleFormatter {
     });
   }
 
+  get text(): string {
```
OrKoN: does this getter need to be public? let's make it a private or `#` property?

OrKoN (approved):

PR #2553, opened 2026-08-11
feat: Add query_heapsnapshot MCP tool

dinfuehr:
`query_heapsnapshot_objects` allows the agent to filter objects by properties like class name, self size, retained size, property name and/or detachedness. The resulting list of objects can be sorted as well.


dinfuehr:
I've rebased this PR.

dinfuehr:
I've rebased this PR on tip-of-tree.

src/tools/memory.ts
```diff
@@ -465,3 +465,80 @@ export const getHeapSnapshotObjectDetails = defineTool({
...
+      .string()
+      .optional()
+      .describe('Optional property name filter for outgoing reference edges.'),
+    nodeType: zod
+      .string()
+      .optional()
+      .describe(
+        'Optional V8 node type filter (e.g. object, closure, string, array, code).',
+      ),
+    minRetainedSize: zod
```
Lightning00Blade, line 496: I am curious if we can combine this min/max into a range and use a `-` as a separator, WDYT?

Lightning00Blade (approved):
LGTM, with a small question.

src/tools/memory.ts
dinfuehr: Makes sense imho. Also I think it would be nice to allow e.g. 1MB as argument. Both should make this tool easier to use manually. I would do it as a follow-up to this CL. Wdyt?

Lightning00Blade: Follow-up sounds great!

dinfuehr:
Thanks for the review!

PR #2554, opened 2026-08-11
test: use 127.0.0.1 to avoid host resolution

OrKoN:
Attempt to fix flakiness in e2e tests that could be potentially caused by a host lookup that resolves to an IPV6 address.


Lightning00Blade (commented):
You need to update the code that stabilizes the snapshots.

nattallius (approved):

PR #2559, opened 2026-08-12
chore: add puppeteer to allowed list

Lightning00Blade:


package.json
```diff
@@ -108,5 +108,12 @@
       "eslint": "$eslint"
     },
     "puppeteer-core": "$puppeteer"
+  },
+  "allowScripts": {
+    "core-js": false,
+    "@google/genai": false,
+    "protobufjs": false,
+    "unrs-resolver": false,
+    "puppeteer@25.5.0": true
```
OrKoN: does it need to be a specific version?

Lightning00Blade: I think not, but didn't see it 
Ran `npm install-scripts deny unrs-resolver` for the disable and 
And `npm install-scripts approve puppeteer` for the enable.
So it seems NPM is trying to be extra protective + only allow the top level to pull it.

OrKoN: could we remove `@25.5.0` and see if it still works?

Lightning00Blade: Done

OrKoN (approved):

PR #2560, opened 2026-08-12
test: return JSON for stabilizeStructuredContent

Lightning00Blade:
We never all this method without then converting it to JSON, if we need it we can the expose a flag to return it as non json.

### DIFF
--- a/tests/utils.ts
+++ b/tests/utils.ts
@@ -299,25 +299,30 @@ export function html(
 </html>`;
 }
 
-export function stabilizeStructuredContent(content: unknown): unknown {
-  if (typeof content === 'string') {
-    return stabilizeResponseOutput(content);
-  }
-  if (Array.isArray(content)) {
-    return content.map(item => stabilizeStructuredContent(item));
-  }
-  if (typeof content === 'object' && content !== null) {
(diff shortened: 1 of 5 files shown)


tests/utils.ts
```diff
@@ -299,25 +299,30 @@ export function html(
...
-    return content.map(item => stabilizeStructuredContent(item));
-  }
-  if (typeof content === 'object' && content !== null) {
-    const result: Record<string, unknown> = {};
-    for (const [key, value] of Object.entries(content)) {
-      if (key === 'snapshotFilePath' && typeof value === 'string') {
-        result[key] = '<file>';
-      } else {
-        result[key] = stabilizeStructuredContent(value);
+export function stabilizeStructuredContent(content: unknown): unknown;
```
nroscino: do we need this? It seems to me we return always a string here 🤔

nroscino (approved):
LGTM! % small comment!

tests/utils.ts
Lightning00Blade:
```suggestion
```

Lightning00Blade: Done.

PR #2561, opened 2026-08-12
feat: Add additional params to get_heapsnapshot_edges

dinfuehr:
Add additional parameters to get_heapsnapshot_edges for filtering and sorting the emitted edges. By default this tool now does not print edges to primitive values like true/false/null/undefined anymore unless explicitly enabled. Edges are also sorted by retained size by default.


Lightning00Blade (approved):
LGTM

dinfuehr:
Thanks for the review!

PR #2562, opened 2026-08-12
refactor: replace debug with NodeJS native

Lightning00Blade:


OrKoN (approved):
LGTM I think we are still missing timestamps

src/utils/logger.ts
```diff
@@ -45,4 +39,29 @@ export function flushLogs(
...
 }
 
-export const logger: Logger = debug(mcpDebugNamespace) as Logger;
+export const logger: Logger = (...args: unknown[]) => {
+  if (logFileStream) {
+    logFileStream.write(
+      `${new Date().toISOString()} ${mcpDebugNamespace} ${util.format(...args)}\n`,
+    );
+  } else if (_debugLog.enabled) {
+    _debugLog('%s', util.format(...args));
```
OrKoN: does this branch exclude mcpDebugNamespace and timestamp?

OrKoN: `const _debugLog = util.debuglog(mcpDebugNamespace);` namespace is there, but what about the timestamp?

Lightning00Blade: Added the timestamp to the regular logs as well.

PR #2563, opened 2026-08-12
chore: use sparse checkout

Lightning00Blade:
With this change `devtools-frontend` takes around ~135MB of space on the repo.

### DIFF
--- a/scripts/prepare.ts
+++ b/scripts/prepare.ts
@@ -65,10 +65,13 @@ function ensureSubmodule(): void {
     );
 
     // 2. Set the sparse checkout configuration
-    execSync('git sparse-checkout set front_end mcp extension-api', {
-      cwd: devtoolsFrontendDir,
-      stdio: 'inherit',
-    });
+    execSync(
+      'git sparse-checkout set --no-cone "/*" "!/*/" "/front_end/" "!/front_end/panels/" "!/front_end/ui/" "/mcp/" "/extension-api/"',
+      {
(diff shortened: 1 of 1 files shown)


OrKoN (commented):
It seems that does not update existing checkout outs but breaks deinit:

```
❯ git submodule deinit --all
warning: Submodule work tree 'third_party/devtools-frontend' contains a .git directory. This will be replaced with a .git file by using absorbgitdirs.
Migrating git directory of 'third_party/devtools-frontend' from
'/Users/alexrudenko/src/chrome-devtools-mcp/third_party/devtools-frontend/.git' to
'/Users/alexrudenko/src/chrome-devtools-mcp/.git/modules/devtools-frontend'
fatal: could not migrate git directory from '/Users/alexrudenko/src/chrome-devtools-mcp/third_party/devtools-frontend/.git' to '/Users/alexrudenko/src/chrome-devtools-mcp/.git/modules/devtools-frontend': Permission denied
```

OrKoN:
@Lightning00Blade do you still want to land it? see may comment about the issue I encountered. Is there a workaround?

Lightning00Blade:
Yes, I want to land this, but I think i need to write a migration for pre-existing checkouts.

OrKoN:
Do you think it's worth it given http://b/557070829?

Lightning00Blade:
Was not aware of this. In light of that I don't think this is worth it atm.

PR #2565, opened 2026-08-12
Fix arbitrary file write via dangling symlink escaping workspace root

prasanna8585:
## Summary
`resolveCanonicalPath()` treated a dangling symlink (target does not exist) the same as a plain non-existent path - it fell back to the nearest existing ancestor + the symlink's own basename, instead of resolving to where the symlink actually points.

## Impact
A dangling symlink placed inside an allowed workspace root, pointing outside of it, was treated by `validatePath()`/`ensureExtension()` as safely inside the root. `takeHeapSnapshot` (Puppeteer's `captureHeapSnapshot`, which uses plain `fs.createWriteStream` with no `O_NOFOLLOW`) and `startScreencast` (spawns `ffmpeg`, which also follows symlinks by default) would then write their output through the symlink to the attacker-chosen location outside the sandboxed root - e.g. a malicious/untrusted repository checked out for automated browser testing could contain a symlink like `snapshot.heapsnapshot -> ~/.ssh/authorized_keys` and have a routine heap-snapshot tool call overwrite it.

This is the same bug class already fixed for the screenshot path in #2269 (which added `O_NOFOLLOW` to the internal `#writeFile` helper) - heap snapshots and screencasts don't go through that helper, since their actual file I/O is delegated to Puppeteer-core/ffmpeg.

## Fix
`resolveCanonicalPath` now detects when the unresolvable path (or an ancestor of it) is itself a symlink, and recursively resolves to its actual target (handling chained symlinks), with cycle detection to prevent infinite recursion on a symlink loop.

## Testing
- Added unit tests in `tests/utils/files.test.ts`: dangling symlink resolves to its real outside target, chained dangling symlinks resolve to the ultimate target, symlink cycles throw instead of hanging.
- Added an end-to-end test in `tests/roots.test.ts` confirming `validatePath`/`ensureExtension` reject a dangling symlink pointing outside the root, and that the outside target is never created.
- All tests, including the Chrome-dependent `roots.test.ts` suite, were run and passed locally with a real Chrome install.
- Confirmed via revert-and-retest that the new tests correctly fail against the unfixed code (not tautological).

### DIFF
--- a/src/utils/files.ts
+++ b/src/utils/files.ts
@@ -15,50 +15,105 @@ export async function getTempFilePath(filename: string) {
   return filepath;
 }
 
-export async function resolveCanonicalPath(filePath: string): Promise<string> {
+export async function resolveCanonicalPath(
+  filePath: string,
+  visited: Set<string> = new Set(),
+): Promise<string> {
   const absolutePath = path.resolve(filePath);
+
+  // Guard against symlink cycles (A -> B -> A) causing infinite recursion.
+  if (visited.has(absolutePath)) {
+    throw new Error(
+      `Symlink cycle detected while resolving canonical path for: ${filePath}`,
+    );
+  }
+
   try {
     // Get the true canonical path, resolving all symlinks.
     return await fs.realpath(absolutePath);
   } catch (err) {
     if (
-      err &&
-      typeof err === 'object' &&
-      'code' in err &&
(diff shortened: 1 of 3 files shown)


OrKoN:
This appears to be a duplicate of https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2457/changes

PR #2566, opened 2026-08-12
chore: remove imported panel code

Lightning00Blade:
This was added long time ago, and now the code is correctly moved into the models/


nroscino (approved):
LGTM!

PR #2571, opened 2026-08-13
refactor(performance): use default trace categories from DevTools

jackfranklin:
Instead of maintaining a separate, manual list of tracing categories that can drift out of sync with DevTools, consume TracingDefaultCategories exposed by devtools-frontend.

Additionally include optional categories (JsSampling and Screenshot) enabled by default in the DevTools UI to retain JavaScript profiling samples and timeline screenshots.

### DIFF
--- a/src/tools/performance.ts
+++ b/src/tools/performance.ts
@@ -70,25 +70,12 @@ export const startTrace = definePageTool({
         });
       }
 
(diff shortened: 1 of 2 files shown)


OrKoN (approved):
Thanks, LGTM!

PR #2572, opened 2026-08-13
refactor: make validatePath return resolved url

OrKoN:
This changes the internal responsibility: validatePath is the only place that is expected to follow symlinks if it can. It then returns a resolved path for all consumers in the MCP server who are expected not to follow symlinks.

This is to be combined with the followSymlinks=false setting in https://github.com/puppeteer/puppeteer/pull/15335

### DIFF
--- a/src/ToolHandler.ts
+++ b/src/ToolHandler.ts
@@ -25,7 +25,7 @@ import type {
 import {pageIdSchema} from './tools/ToolDefinition.js';
 import {logger} from './utils/logger.js';
 import type {Mutex} from './third_party/index.js';
-import {fileURLToPath} from 'node:url';
+import {fileURLToPath, pathToFileURL} from 'node:url';
 import {isLocalhost} from './utils/url.js';
 
 export function buildFlag(category: ToolCategory) {
@@ -151,14 +151,21 @@ function buildUnknownArgumentsMessage(
   return `Unknown ${unknownLabel} for tool "${toolName}": ${formatArgumentNames(unknownArgumentNames)}. ${expectedArguments} ${correction} and retry.`;
 }
 
-function extractPaths(value: unknown): string[] {
-  if (typeof value === 'string') {
-    return [value];
+async function validateAndResolvePathOrUrl(
+  filePathOrUrl: string,
+  context: McpContext,
+): Promise<string> {
+  try {
+    const url = new URL(filePathOrUrl);
+    if (url.protocol === 'file:') {
+      return pathToFileURL(await context.validatePath(fileURLToPath(url))).href;
+    } else if (['http:', 'https:', 'ws:', 'wss:'].includes(url.protocol)) {
+      return filePathOrUrl;
+    }
+  } catch {
+    // Suppress parsing errors for regular file paths.
   }
-  if (Array.isArray(value)) {
-    return value.filter(item => typeof item === 'string');
-  }
-  return [];
+  return await context.validatePath(filePathOrUrl);
 }
 
 function isLocalBrowser(context: McpContext): boolean {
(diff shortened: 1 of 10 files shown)


Lightning00Blade (approved):
LGTM, 
Should we swap the value in the parameters with the resolved one?
`src/ToolHandler.ts:191`?

OrKoN:
> LGTM, Should we swap the value in the parameters with the resolved one? `src/ToolHandler.ts:191`?

I think no because we might want to use the user provided value for responses to not break expectations if the canonical path is different from the input path. The resolved path is also always absolute.

tests/tools/memory.test.ts
```diff
@@ -39,9 +40,10 @@ describe('memory', () => {
             response,
             context,
           );
+          const canonicalFilePath = await resolveCanonicalPath(filePath);
           assert.equal(
             response.responseLines.at(0),
-            `Heap snapshot saved to ${filePath}`,
+            `Heap snapshot saved to ${canonicalFilePath}`,
```
OrKoN, line 46: actually we might want to keep filePath here

nroscino (approved):
LGTM % question. 
Maybe it could be worth, like we do for `saveFile`, to use the path returned from the `validatePath` for the for `installExtension` and `installPWA` tools, WDYT?

OrKoN:
> LGTM, Should we swap the value in the parameters with the resolved one? `src/ToolHandler.ts:191`?

I thought about this more and I think we should do that after all since we already return resolved paths in a bunch of tools.

OrKoN:
> LGTM % question. Maybe it could be worth, like we do for `saveFile`, to use the path returned from the `validatePath` for the for `installExtension` and `installPWA` tools, WDYT?

Updated the PR with replacing paths with resolved ones in ToolHandler so this should cover those two as well. Re-configured extensions to only validate paths for local Chrome's.

nroscino (approved):
LGTM!

src/ToolHandler.ts
```diff
@@ -151,14 +151,28 @@ function buildUnknownArgumentsMessage(
...
+      filePath = fileURLToPath(url);
+    } else if (['http:', 'https:', 'ws:', 'wss:'].includes(url.protocol)) {
+      return filePathOrUrl;
+    }
+  } catch {
+    // Suppress parsing errors for regular file paths.
   }
-  if (Array.isArray(value)) {
-    return value.filter(item => typeof item === 'string');
+  const validated = (await context.validatePath(filePath)) ?? filePath;
```
Lightning00Blade: Is this needed, I think `filePath` should always be string so the type override should return it as well.

src/ToolHandler.ts
```diff
@@ -151,14 +151,28 @@ function buildUnknownArgumentsMessage(
...
+  if (isFileUrl) {
```
Lightning00Blade: Can we just have this in the `file:` if, we only set this flag once, so we can return the value from there, similar to what we do for network url

src/ToolHandler.ts
```diff
@@ -194,25 +208,23 @@ async function validateToolFiles(
...
+      const val = params[key];
+      if (typeof val === 'string') {
+        params[key] = await validateAndResolvePathOrUrl(val, context);
+      } else if (Array.isArray(val)) {
+        const updated: unknown[] = [];
+        for (const item of val) {
+          if (typeof item === 'string') {
+            updated.push(await validateAndResolvePathOrUrl(item, context));
+          } else {
+            updated.push(item);
```
Lightning00Blade: Should we add this? This seems wrong, if it's not string we should error I think?

Lightning00Blade (approved):

PR #2573, opened 2026-08-13
fix: do not validate local files for remote browsers

OrKoN:
I noticed that we validate file paths locally that are handled by potentially remote browsers.  This PR extends `verifyFilesSchema` to validate file paths for local or remote browsers or both  + test coverage.

### DIFF
--- a/src/ToolHandler.ts
+++ b/src/ToolHandler.ts
@@ -156,19 +156,16 @@ function extractPaths(value: unknown): string[] {
     return [value];
   }
   if (Array.isArray(value)) {
-    return values.filter(item => typeof item === 'string');
+    return value.filter(item => typeof item === 'string');
   }
   return [];
 }
 
 function isLocalBrowser(context: McpContext): boolean {
-  if (!context.browser) {
-    return false;
-  }
-  if (context.browser.process?.()) {
+  if (context.browser.process()) {
     return true;
   }
-  const wsEndpoint = context.browser.wsEndpoint?.();
+  const wsEndpoint = context.browser.wsEndpoint();
   if (wsEndpoint && isLocalhost(wsEndpoint)) {
     return true;
   }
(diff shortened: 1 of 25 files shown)


src/tools/pwa.ts
```diff
@@ -52,13 +50,12 @@ export const installPwa = defineTool({
     displayMode: displayModeSchema,
   },
   blockedByDialog: false,
+  // We do not validate file paths here because they can be on the remote
+  // browser instance and are not accessed by the MCP server itself.
   verifyFilesSchema: [],
+  verifyFilesSchemaLocalOnly: ['installUrlOrBundleUrl'],
```
Lightning00Blade: From API perspective this is hard to understand.
What if we refactor it to 
```
verifyFilesSchema: Record<Keyof Schema, true | {local: boolean, remote: boolean}>
```
That way we have the validation on a single place.

src/ToolHandler.ts
```diff
@@ -148,6 +151,83 @@ function buildUnknownArgumentsMessage(
...
+    return [value];
+  }
+  if (Array.isArray(value)) {
+    const paths: string[] = [];
+    for (const item of value) {
+      if (typeof item === 'string') {
+        paths.push(item);
+      }
+    }
+    return paths;
```
Lightning00Blade:
```suggestion
    return values.filter((item) => typeof item === 'string');
```

src/ToolHandler.ts
```diff
@@ -148,6 +151,83 @@ function buildUnknownArgumentsMessage(
...
+  }
+  return false;
+}
+
+async function validateToolFiles(
+  tool: ToolDefinition | DefinedPageTool,
+  params: Record<string, unknown>,
+  context: McpContext,
+): Promise<void> {
+  if (!tool.verifyFilesSchema) {
```
Lightning00Blade: Is this possible to be no set under our current types?

Lightning00Blade (approved):

PR #2577, opened 2026-08-14
fix: warn on unknown CLI args

Lightning00Blade:


src/bin/chrome-devtools-mcp-cli-options.ts
```diff
@@ -418,6 +409,23 @@ export function parser(
         );
         args.usageStatistics = false;
       }
+
+      const cliOptionsAlloweArgs = [
```
OrKoN:
```suggestion
      const cliOptionsAllowedArgs = [
```

src/bin/chrome-devtools-mcp-cli-options.ts
```diff
@@ -418,6 +409,23 @@ export function parser(
...
+        ...Object.keys(cliOptions),
+        // Yargs populated with positional args
+        '_',
+        '$0',
+      ];
+
+      const unknownArgs = Object.keys(args).filter(
```
OrKoN, line 420: does it still work for `--no-<bool>` flag variants?

Lightning00Blade: Yes running with 
```sh
npx @modelcontextprotocol/inspector node ./build/src/bin/chrome-devtools-mcp.js  --headless-new=true --no-slim
```
Produces `Unknown arguments: --headlessNew`

OrKoN (approved):

PR #2579, opened 2026-08-14
fix: require .heapsnapshot (or .heaptimeline) as file extension

dinfuehr:
This makes it harder to pass wrong arguments to the CLI tools.


Lightning00Blade:
@OrKoN Do you think we should make this part of the `verifyFilesSchema` and have `{extensions: Array<`.${string}>}`?

OrKoN:
I think we could do that, but I think we currently only have this one instance of requiring extension validation? or do we have more?

Lightning00Blade:
> I think we could do that, but I think we currently only have this one instance of requiring extension validation? or do we have more?

We have this check in the Memory tools and Screencast tools. And some places where we use `context.saveFile` but those need to be tracked per place.

Lightning00Blade (approved):

dinfuehr:
Thanks for the review!

PR #2585, opened 2026-08-17
fix(cli): enable tools which require --memoryDebugging on the CLI

dinfuehr:
Currently tools that require --memoryDebugging cannot directly be invoked from the CLI. The server needs to be start explicitly with the --memoryDebugging flag first. This PR makes --memoryDebugging a default for the CLI.

### DIFF
--- a/src/bin/chrome-devtools.ts
+++ b/src/bin/chrome-devtools.ts
@@ -31,41 +31,39 @@ import {checkForUpdates} from '../utils/check-for-updates.js';
 import {VERSION} from '../version.js';
 
 import {commands} from '../config/cli-options.js';
-import {mcpOptions, parseArguments} from '../config/mcp-options.js';
+import {
+  mcpOptions,
+  parseArguments,
+  getMcpOptionsForViaCli,
+} from '../config/mcp-options.js';
 
 await checkForUpdates(
   'Run `npm install -g chrome-devtools-mcp@latest` and `chrome-devtools start` to update and restart the daemon.',
 );
 
+const DEFAULT_CLI_ARGS = ['--viaCli'];
+
 async function start(args: string[], sessionId: string) {
-  const combinedArgs = [...args, ...defaultArgs];
+  const combinedArgs = [...DEFAULT_CLI_ARGS, ...args];
   await startDaemon(combinedArgs, sessionId);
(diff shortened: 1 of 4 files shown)


scripts/generate-cli.ts
```diff
@@ -169,12 +171,16 @@ async function generateCli() {
     const isOffByDefault = OFF_BY_DEFAULT_CATEGORIES.includes(categoryEnum);
     if (isOffByDefault) {
       const categoryFlag = buildFlag(categoryEnum);
-      requiredFlags.push(`--${categoryFlag}=true`);
+      if (!cliDefaults[categoryFlag]) {
```
OrKoN: let's keep pushing the required flags unconditionally because the defaults can be still overridden via args to the start command and the CLI documentation/options are not dependent on that.

src/config/cli-defaults.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {mcpOptions} from './mcp-options.js';
+
+export const DEFAULT_CLI_ARGS = [
```
OrKoN: let's iterate this list to change the defaults in getStartCliOptions? or extract another list for overriden defaults?

src/bin/chrome-devtools.ts
```diff
@@ -31,42 +31,24 @@ import {checkForUpdates} from '../utils/check-for-updates.js';
...
-  ...mcpOptions,
-} as Partial<typeof mcpOptions>;
-
-// Missing CLI serialization.
-delete startCliOptions.viewport;
-
-// Change the defaults for the CLI.
-delete startCliOptions.experimentalStructuredContent;
-delete startCliOptions.experimentalInteropTools;
-delete startCliOptions.experimentalPageIdRouting;
```
OrKoN, line 58: do we still keep these?

OrKoN (approved):

src/bin/chrome-devtools.ts
dinfuehr: Thanks! Sorry, Gemini decided to drop them. We now keep them.

dinfuehr:
Thanks for the quick reviews!

PR #2587, opened 2026-08-18
feat: Allow human readable sizes

dinfuehr:
Allow human readable sizes like 1M or 1gb as input.


dinfuehr:
Note: Ranges will be a follow-up to this PR.

dinfuehr:
Hmm, seems like the range support largely makes this PR unncessary. I will open a PR for ranges right away instead.

PR #2589, opened 2026-08-18
feat: Use ranges with human readable sizes

dinfuehr:
Replace separate minimum and maximum heap-size filters with range strings. Values can use byte counts or units, such as `1024-2048`, `1MB-2MB`, `-1MiB`, and `1GiB-`. A bare value such as `10M` is treated as a minimum.

The range syntax is used by `query_heapsnapshot_objects` for retained and self sizes and by `get_heapsnapshot_edges` for retained sizes.

### DIFF
--- a/src/tools/memory.ts
+++ b/src/tools/memory.ts
@@ -5,6 +5,7 @@
  */
 
 import {zod} from '../third_party/index.js';
+import {byteSizeRangeSchema} from '../utils/bytes.js';
 
 import {ToolCategory} from './categories.js';
 import {definePageTool, defineTool} from './ToolDefinition.js';
@@ -305,10 +306,9 @@ export const getHeapSnapshotEdges = defineTool({
       .enum(['retainedSize', 'selfSize', 'name'])
       .optional()
       .describe('Sort order for edges. Default is retainedSize.'),
-    minRetainedSize: zod
-      .number()
-      .optional()
-      .describe('Minimum retained size in bytes for target nodes.'),
+    retainedSize: byteSizeRangeSchema(
+      'Inclusive retained size range (e.g. "1MB-2MB", "-1MB", or "1MB-") for target nodes. A single value is treated as a minimum. Currently, only the lower bound is applied.',
+    ).optional(),
     excludePrimitives: zod
       .boolean()
       .optional()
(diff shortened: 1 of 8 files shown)


src/utils/bytes.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  const bytes = Math.round(num * multiplier);
+  if (!Number.isFinite(bytes)) {
+    throw new Error(`Invalid byte size: "${value}"`);
+  }
+  return bytes;
+}
+
+export interface ByteSizeRange {
+  min?: number;
```
Lightning00Blade:
```suggestion
  min: number;
```

src/tools/memory.ts
```diff
@@ -19,6 +20,23 @@ const HEAP_SNAPSHOT_FILTERS: readonly [string, ...string[]] = [
   'attributedToSpecificNativeContext',
 ];
 
+function byteSizeRangeSchema(description: string) {
+  return zod
+    .string()
+    .superRefine((value, context) => {
```
Lightning00Blade: Let's use a [transform](https://zod.dev/api#transform) with `return z.NEVER;`
That way we only need run the `parseByteSizeRange` once.

dinfuehr: Thanks a lot! This is **much** better indeed.

tests/tools/memory.test.ts
```diff
@@ -26,11 +26,32 @@ import {
...
   queryHeapSnapshotObjects,
 } from '../../src/tools/memory.js';
+import {parseByteSizeRange} from '../../src/utils/bytes.js';
 import {stableIdSymbol} from '../../src/utils/id.js';
 import {resolveCanonicalPath} from '../../src/utils/files.js';
 import {withMcpContext} from '../utils.js';
 
 describe('memory', () => {
+  it('validates byte-size range parameters', () => {
+    const schemas = [
```
Lightning00Blade: Let's export the Schema helper (we can even move it to `utils/bytes.ts`). 
As the technically it's that what we are testing. 
And we should just have one passing and failing test, as we test all the cases for the helper separately.

Lightning00Blade (approved):
LGTM, with one comment

Lightning00Blade (approved):

Issue #408, opened 2025-10-17
Cookie management support


natorion:
### Is your feature request related to a problem? Please describe.

**_FEEDBACK WANTED_**

### 1. Problem Statement

AI agents are currently "state-blind." They cannot see, modify, or delete one of the most common forms of client-side state: **browser cookies**.

This is a major debugging roadblock. An agent cannot diagnose or solve entire classes of problems related to:
* Authentication (e.g., missing session cookies)
* A/B testing or feature flags (e.g., `feature_flag_beta=on`)
* User preferences (e.g., `theme=dark`)

Without access to cookies, the AI is forced to *guess* about the application's state or ask the human to manually open DevTools, which defeats the purpose of an autonomous AI assistant.

### 2. Problems We Want to Solve

We believe there are three primary problems that AI agents face when dealing with cookies. We'd like to know how often you encounter these.

**Problem 1: Inspecting State (The "Read" Problem)**
An agent needs to read existing cookies to understand the current application state.
* **Goal:** "I need to know if a `session_token` cookie exists and what its value is to understand why the user is seeing a '401 Unauthorized' error."
* **Goal:** "I need to check the value of the `feature_flag` cookie to confirm my hypothesis about why a component is not visible."

**Problem 2: Simulating State (The "Write/Edit" Problem)**
An agent needs to set or modify cookies to test different application states or scenarios.
* **Goal:** "I need to test the 'dark mode' feature. The easiest way is to set the `theme` cookie to `dark` and reload the page."
* **Goal:** "I need to see if the new beta feature works. I want to *add* a new cookie `beta_user=true`."

**Problem 3: Resetting State (The "Delete" Problem)**
An agent needs to delete cookies to return the application to a "clean" or default state for reproducible debugging.
* **Goal:** "I want to debug the 'first-time user' login flow. To do this, I must clear all session cookies for the current domain."
* **Goal:** "The user's preferences seem corrupt. I want to delete the `user_settings` cookie and have them start fresh."

---

### 3. Key Use Cases (As Problems)

* **As an...** AI agent debugging a login issue, **I want to inspect all cookies** for the current domain to see if the `session_token` is missing, expired, or malformed.
* **As an...** AI agent testing a new feature, **I want to set a cookie** `feature_flag_x=true` and reload the page to see if the new UI appears.
* **As an...** AI agent trying to reproduce a bug, **I want to delete all cookies** for this domain to ensure I am testing the application from a clean, logged-out state.
[...]

PR #2596, opened 2026-08-20
feat: Add cookie-debugging skill and improve network/pages tool descriptions

natorion:
#408


scripts/eval_gemini.ts
```diff
@@ -17,7 +17,12 @@ import {TestServer} from '../build/tests/server.js';
 
 const ROOT_DIR = path.resolve(import.meta.dirname, '..');
 const SCENARIOS_DIR = path.join(import.meta.dirname, 'eval_scenarios');
-const SKILL_PATH = path.join(ROOT_DIR, 'skills', 'chrome-devtools', 'SKILL.md');
```
natorion, line 20: don't rename constants

scripts/eval_result.ts
```diff
@@ -37,19 +37,33 @@ export class Result {
...
       this.nextCallIndex++;
     }
 
-    const navCall = this.calls[this.nextCallIndex];
     assert.ok(
-      navCall &&
-        (navCall.name === 'new_page' || navCall.name === 'navigate_page'),
-      `Expected navigation call (new_page or navigate_page), but got: ${navCall?.name || 'none'}`,
+      hasNavigated,
+      `Expected at least one navigation call (new_page or navigate_page), but got none.`,
```
OrKoN: This message is now outdated, do we really need to listen to all 4 tool calls?

skills/cookie-debugging/references/cookie-snippets.md
```diff
@@ -0,0 +1,147 @@
+# Cookie Debugging Snippets
+
+Use these JavaScript snippets with the `evaluate_script` tool.
+
+> [!NOTE]
+> All JavaScript snippets below execute in the web page's JavaScript context and therefore only have access to non-`HttpOnly` cookies. To inspect `HttpOnly` session tokens, use `get_network_request` on outgoing request (`Cookie`) or incoming response (`Set-Cookie`) headers.
+
+---
+
+## 1. Parse `document.cookie`
```
OrKoN: I am surprised the model needs such snippets, I would have expected things like this are baked in. Should we reduce the size of the skill by including general steps of how to do cookie debugging (e.g., httpOnly vs local) without including specific snippets? or did you observe that not work?

OrKoN (approved):

skills/cookie-debugging/references/cookie-snippets.md
```diff
@@ -0,0 +1,147 @@
+# Cookie Debugging Snippets
+
+Use these JavaScript snippets with the `evaluate_script` tool.
+
+> [!NOTE]
+> All JavaScript snippets below execute in the web page's JavaScript context and therefore only have access to non-`HttpOnly` cookies. To inspect `HttpOnly` session tokens, use `get_network_request` on outgoing request (`Cookie`) or incoming response (`Set-Cookie`) headers.
```
matthiasrohmer:
```suggestion
> All JavaScript snippets below execute in the web page's JavaScript context and therefore only have access to cookies set through JavaScript. To inspect cookies set through HTTP headers and thus `HttpOnly`, use `get_network_request` on outgoing request (`Cookie`) or incoming response (`Set-Cookie`) headers.
```

The "non" construct seems like something smaller models might stumble over, and HTTP cookies could be used for more than session tokens, so staying consistent with the term might be good.

skills/cookie-debugging/references/cookie-snippets.md
matthiasrohmer: Relatedly is there any benefit in keeping this instead of only the newer 2.?

As long as the skill is Chrome (DevTools) specific I think we can just drop the legacy way of doing it, and instead advertise the newer alternative - which might also not be as prominent in models yet.

skills/cookie-debugging/references/cookie-snippets.md
```diff
@@ -0,0 +1,147 @@
...
+      return {name, value: decodeURIComponent(rest.join('='))};
+    });
+};
+```
+
+---
+
+## 3. Cookie Diff / Comparison Helper
+
+Takes a baseline cookie map (obtained from snippet 1) and compares it against current cookies to report added, modified, and removed cookies. Useful for testing cookie banner actions (Decline vs Accept).
```
matthiasrohmer: Relatedly I think ideally we use the newer CookieStore API for all snippets.

skills/cookie-debugging/references/cookie-snippets.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+---
+
+## 5. Parse Lighthouse Third-Party Cookie Audit (CLI One-Liner)
+
+Use this command in your terminal/Node environment to extract the `third-party-cookies` audit from a Lighthouse JSON report without loading the entire JSON file into the LLM context:
+
+```bash
+node -e "const r=require('/tmp/lh-report/report.json'); const a=r.audits['third-party-cookies']; console.log(JSON.stringify({score: a?.score, displayValue: a?.displayValue, items: a?.details?.items}))"
```
matthiasrohmer: Is the report always in this location? I haven't checked but have a faded memory to have it seen under various paths?

Not super crucial, any model will be able to self correct if the require fails, but maybe we find a better instruction here.

skills/cookie-debugging/references/cookie-snippets.md
matthiasrohmer: Looking at the actual skill it seems like we can drop this all along as it's better worded in the original skill and this would just be duplicated context...?

skills/cookie-debugging/SKILL.md
```diff
@@ -0,0 +1,133 @@
...
+3. **Inspect Premature Network Requests & Issues**:
+   - Call `list_network_requests` to ensure no third-party tracking beacons fired before consent.
+   - Call `list_console_messages` with `types: ["issue"]` to check for tracking warnings.
+4. **Interact with Consent Banner**:
+   - Capture snapshot with `take_snapshot` to locate the "Decline" or "Reject All" button `uid`.
+   - Click the button with `click`.
+5. **Verify Cookie Difference**:
+   - Run `evaluate_script` with the **"Cookie Diff / Comparison Helper" snippet** from [references/cookie-snippets.md](references/cookie-snippets.md), passing the baseline cookie map.
+   - Assert that only strictly necessary or consent-state cookies exist.
+6. **Teardown Context**: Call `close_page` when the audit is complete to prevent leftover cookies from affecting subsequent tasks.
```
matthiasrohmer: Do we want to add a step here that should try to revoke consent, and then check cookies that have previously been accepted are wiped again?

I know this user journey is usually under-tested, but legally as important.

skills/cookie-debugging/SKILL.md
```diff
@@ -0,0 +1,133 @@
...
+       "includePreservedMessages": true
+     }
+     ```
+   - Check for `CookieIssue` entries, such as:
+     - `SameSiteNoneInsecure`: `SameSite=None` without `Secure`.
+     - `ThirdPartyCookiePhaseout`: Third-party cookie blocked or restricted.
+     - `SchemefulSameSite`: Cross-scheme cookie issues.
+     - `PartitionedCookies`: Invalid CHIPS partitioning attributes.
+2. **Deep Audit: Lighthouse Third-Party Cookies**:
+   - Run `lighthouse_audit` with `mode: "navigation"` and `outputDirPath: "/tmp/lh-report"`.
```
matthiasrohmer, line 104: Just checking: ".json" is auto-appended?

skills/cookie-debugging/references/cookie-snippets.md
```diff
@@ -0,0 +1,147 @@
...
+    detectedCMPs,
+    consentStorageKeys,
+    cookieCount: document.cookie ? document.cookie.split(';').length : 0,
+  };
+};
+```
+
+---
+
+## 5. Parse Lighthouse Third-Party Cookie Audit (CLI One-Liner)
```
matthiasrohmer: Duplicate with the original skill, lines 100ff?

skills/cookie-debugging/SKILL.md
```diff
@@ -0,0 +1,137 @@
+---
+name: cookie-debugging
+description: Uses Chrome DevTools MCP for inspecting, debugging, and testing cookies, session state, authentication issues, and cookie consent compliance. Use when diagnosing 401/403 errors, authentication redirects, session expiration, Cookie/Set-Cookie header issues, cookie banner consent conformance, or third-party cookie/SameSite/Partitioned cookie warnings.
+---
+
+## Core Concepts
+
+### HttpOnly vs JavaScript Access
+
+Cookies marked `HttpOnly` cannot be accessed or modified by client-side JavaScript (`document.cookie` or `cookieStore`). However, the browser **automatically attaches active HttpOnly cookies to outgoing HTTP request headers (`Cookie`)**.
```
OrKoN: should we even mention document.cookie? per https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie it is not recommended although in this case there will be no performance impact.

scripts/eval_result.ts
natorion: Refactored.

skills/cookie-debugging/references/cookie-snippets.md
natorion: Makes sense. I removed cookie-snippets.md.

skills/cookie-debugging/SKILL.md
natorion: I think for the debugging use case its fine to use it.

skills/cookie-debugging/SKILL.md
natorion: Good point, let me check how many characters this adds.

skills/cookie-debugging/SKILL.md
natorion: Yeah. report.json is the file written in the directory given.

skills/cookie-debugging/SKILL.md
natorion: Added, is that what you had in mind?

matthiasrohmer: Yep, excellent!

matthiasrohmer (approved):

PR #2600, opened 2026-08-20
docs: split the docs in separate files

Lightning00Blade:


matthiasrohmer (approved):
Thanks for tackling Nik, I think that's overdue.

With this change we should direct less advanced users to https://developer.chrome.com/docs/devtools/agents in the future, to have install instructions for popular agents at a glance.

And once the doc set on developer.chrome.com is complete a link from the README.md to there would be good. But that's a follow up.

LGTM!

OrKoN (approved):

PR #2601, opened 2026-08-20
fix: capture navigation requests in window.open popups

### LINKED ISSUE #2395: list_network_requests misses navigation requests in tabs opened via window.open

mturac:
## Problem

`list_network_requests` misses the initial navigation request for tabs opened via `window.open`. Chrome fires `Network.requestWillBeSent` for the navigation before the MCP server's `NetworkCollector` is wired to the new page's session, so the request is never recorded.

## Root cause

Puppeteer uses `Target.setAutoAttach` with `waitForDebuggerOnStart: true`, which pauses new tabs before any JS runs. However the initial HTTP navigation request is already dispatched at the network layer before the debugger pause takes effect. `Network.enable` from Puppeteer's `NetworkManager` only arrives after it creates the page object (`target.page()`), which is async and happens well after `Runtime.runIfWaitingForDebugger` resumes the tab.

## Fix

Two-step approach in `McpContext`:

**Step 1 — early `Network.enable`:** Hook `CDPSessionEvent.SessionAttached` on the browser's shared `Connection`. This fires synchronously inside Chrome's `Target.attachedToTarget` handler, before `TargetManager` sends `Runtime.runIfWaitingForDebugger`. Sending `Network.enable` at this point ensures Chrome starts tracking the navigation request before the tab resumes.

**Step 2 — event buffering and replay:** The early `Network.enable` causes Chrome to fire `Network.requestWillBeSent`, but Puppeteer's `NetworkManager` (and the `NetworkCollector`) are not listening yet. We buffer these raw CDP events on the session. In `#onTargetCreated`, after both `target.page()` (wires `NetworkManager`) and `#createMcpPage()` (wires `NetworkCollector`) complete, we replay the buffered events by emitting them on the session. The events flow through the normal pipeline: `CDPSession` → `NetworkManager` → `HTTPRequest` → page `request` event → `NetworkCollector`.

## Testing

Added `captures navigation request in window.open popup` to `tests/tools/network.test.ts`. The test opens a popup via `window.open`, then calls `list_network_requests` on the popup's `McpPage` and asserts the navigation URL is present.

```
node scripts/test.js tests/tools/network.test.ts
...........   ← 11/11 pass
```

Fixes #2395

### DIFF
--- a/src/McpContext.ts
+++ b/src/McpContext.ts
@@ -23,14 +23,17 @@ import {McpPage} from './McpPage.js';
 import {type UncaughtError} from './collectors/PageCollector.js';
 import {ServiceWorkerConsoleCollector} from './collectors/ServiceWorkerCollector.js';
 import {
+  CDPSessionEvent,
   Locator,
   type Browser,
   type BrowserContext,
+  type CDPSession,
   type ConsoleMessage,
   type GetPWAStateOptions,
   type InstallPWAOptions,
   type LaunchPWAOptions,
   type Page,
+  type Protocol,
   type PWAState,
   type ScreenRecorder,
   type Target,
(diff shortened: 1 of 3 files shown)


OrKoN (commented):
Thanks for the PR but will be solving it via https://github.com/puppeteer/puppeteer/issues/15270. There is not need to send a PR for this feature. Stay tuned for updates.

PR #2603, opened 2026-08-21
fix: default pageId routing on for page-scoped tools

### LINKED ISSUE #1245: Make pageId routing default to prevent multi-agent race conditions
## Problem

In multi-agent workflows (e.g., Claude Code with parallel agents), concurrent `select_page` calls create race conditions because `select_page` + action is not atomic. Agent A calls `select_page(page1)`, then Agent B calls `select_page(page2)` before
Agent A's next tool call — Agent A now operates on the wrong page.

This is already solved by `--experimental-page-id-routing`, which allows tools to accept `pageId` directly. But since it's behind an experimental flag, most users don't know about it.

## Proposal

1. Make `pageId` always available on page-scoped tools (remove the experimental gate)
2. Fix `waitForEventsAfterAction` to use the explicit target page's emulation settings instead of the global selected page — currently it reads CPU throttling / network conditions from `#selectedPage`, which may be a different page than the one being
operated on

### Reproduction

_No response_

### Expectation

_No response_

### MCP configuration

_No response_

0.20.3

### Chrome version

_No response_

### Coding agent version

_No response_

### Model version

_No response_

### Chat log

_No response_

### Node version

_No response_

### Operating system

macOS

### Extra checklist

SomSamantray:
Two AI agents sharing one `chrome-devtools-mcp` server could silently act on the wrong browser tab: every page-scoped tool call operated on whatever page was "selected" server-side, so a concurrent `select_page` from one agent could redirect the next tool call from a different agent onto the wrong tab. A `--experimentalPageIdRouting` flag already existed to fix this by letting a call target its page explicitly, but it was opt-in and essentially unused in practice.

This makes that routing the default: `pageId` is now accepted on every page-scoped tool without any flag, while a caller that never mentions `pageId` sees no change in behavior. `--no-experimentalPageIdRouting` and `--slim` still force selected-page-only routing for anyone who wants the old strict semantics.

Two related gaps surfaced while implementing the default flip and are fixed alongside it: the merged `pageId` schema field was required rather than optional, which would have broken every ordinary tool call that omits `pageId` once the flag defaulted on; and `evaluate_script`'s own independent page-resolution branch didn't guard against an omitted `pageId` or respect `--slim`, unlike the generic page-scoped path.

Fixes #1245

## Test plan

- Added/updated tests covering the new default, the `--slim` and `--no-experimentalPageIdRouting` opt-outs, and a regression test proving a page-scoped action's emulation settings (CPU throttling, network conditions) stay scoped to the page it actually targets rather than leaking from whichever page happens to be globally selected.
- `npm run typecheck` and the full touched test files (`tests/cli.test.ts`, `tests/ToolHandler.test.ts`, `tests/McpContext.test.ts`, `tests/tools/script.test.ts`) pass locally against a real headless Chromium.


OrKoN:
Thanks for the PR. This is a duplicate of https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/1777 and we are working on enabling this by default.

PR #2604, opened 2026-08-21
fix: enforce URL guardrail on Lighthouse's out-of-band fetches

### LINKED ISSUE #2567: Lighthouse secondary fetches can cross --allowed-url-pattern after redirects
### Description of the bug

In `chrome-devtools-mcp@1.7.0`, Lighthouse secondary fetches for
`/robots.txt` and `/llms.txt` can follow an HTTP redirect from an allowed
origin to an origin excluded by `--allowed-url-pattern`.

The direct control behaves as expected: a normal subresource request from the
audited page to the excluded origin is blocked and the excluded server receives
no request. During `lighthouse_audit`, however, the same excluded server
receives the redirected `/robots.txt` and `/llms.txt` requests. The requests
include the browser cookie, and the controlled `robots.txt` response body is
included in the generated Lighthouse report.

I understand that the current `SECURITY.md` describes the URL patterns as
optional network guardrails rather than a complete network sandbox. I am
reporting this as a guardrail correctness and hardening bug: the documented CLI
behavior says non-matching runtime requests, including navigations and
subresources, are blocked, while these Lighthouse-owned requests are treated
differently after a redirect.

This behavior was previously reported privately to Google VRP as issue
`545472983`. It was assessed as below their security-tracking threshold and
cleared for public disclosure as a GitHub issue.

### Reproduction

1. Install `chrome-devtools-mcp@1.7.0` and
   `@modelcontextprotocol/sdk@1.30.0` under Node.js 20 LTS.
2. Start two loopback-only HTTP servers on different ports:
   - the allowed origin serves the audited page;
   - the excluded origin records requests and returns controlled canary text.
3. On the allowed origin:
   - include a direct image request to the excluded origin as a negative
     control;
   - redirect `/robots.txt` and `/llms.txt` to the excluded origin.
4. Launch `chrome-devtools-mcp` with `--headless`, `--isolated`, and
   `--allowedUrlPattern=http://127.0.0.1:<allowed-port>/*`.
5. Open the allowed page and run `lighthouse_audit`.

Observed result on the current release:

```text
DIRECT_CONTROL_HITS 0
REDIRECT_HITS 2
CREDENTIALED_REDIRECT true
REPORT_CONTAINS_CANARY true
REPRODUCTION_PASSED true
```

The excluded origin received both redirected requests with the controlled
cookie. The `robots-txt` audit in the generated report contained:

```text
INTERNAL_CANARY_/robots-leak
```

The same self-contained, loopback-only harness passed without modification on
macOS and a separate Debian machine.

### Expectation

The final destination of redirects followed by Lighthouse secondary fetches
should be checked against the configured URL patterns before the request is
made and before credentialed response content is consumed. If this behavior is
intended, the CLI help and documentation should state explicitly that
Lighthouse secondary requests and their redirect destinations are outside the
guardrail.

### MCP configuration

The reproduction launches the published MCP entry point with:

```text
--headless
--isolated
--no-usage-statistics
--no-performance-crux
--executablePath <current Chrome executable>
--allowedUrlPattern http://127.0.0.1:<allowed-port>/*
```

All traffic is loopback-only. Cookies and response data are synthetic canaries.

1.7.0

### Chrome version

151.0.7922.109 (macOS); independently checked with Chrome for Testing 151.0.7922.138 on Debian

### Coding agent version

Not applicable — direct MCP SDK harness

### Model version

Not applicable

### Chat log

Not applicable

### Node version

20.20.2

### Operating system

macOS

### Extra checklist

SomSamantray:
`lighthouse_audit` could leak cookies and page content to an origin excluded by `--allowedUrlPattern`/`--blockedUrlPattern`: an audited page could redirect its own same-origin `/robots.txt` (or `/llms.txt`) to an excluded origin, and that origin's response — including cookies — would end up embedded in the generated Lighthouse report. A direct, page-initiated request to the same excluded origin was (and still is) correctly blocked; only Lighthouse's own internal fetch escaped.

Root cause: Lighthouse's `Fetcher` utility (used by the robots-txt, llms-txt, and source-maps gatherers) fetches resources via the CDP command `Network.loadNetworkResource`, which is documented to ignore normal browser network constraints such as CORS. Verified empirically against real Chromium during planning: this command is invisible to Puppeteer's `--allowedUrlPattern`/`--blockedUrlPattern` guardrail (implemented via `Network.emulateNetworkConditionsByRule`, which only governs page-initiated traffic) and to every CDP Network domain event — a spike showed **zero** `Network.requestWillBeSent` events for a `loadNetworkResource` call that did reach an excluded origin via redirect. The vulnerable code lives inside `lighthouse`'s own `Fetcher` class, which this repo vendors as a pre-built bundle generated from a separate upstream fork (`scripts/update-lighthouse.ts`) — not something a PR here can edit directly.

The fix intercepts the CDP command at the one point this repo *can* own: the shared Puppeteer `CdpCDPSession.prototype.send` method, which every CDP session (including the one Lighthouse creates for itself) funnels through. While `lighthouse_audit` runs and a guardrail is configured, `Network.loadNetworkResource` is fulfilled by a Node-side fetch that validates the initial URL and every redirect hop against the guardrail before following it. No behavior change when no guardrail is configured, and the existing page-initiated-request guardrail is untouched.

Three independent review passes (maintainability, reliability, security) converged on the same secondary issue during review: the original interception patched the shared prototype via a naive closure-captured save/restore, which isn't safe against overlapping installs. Not reachable today — a pre-existing global mutex serializes every MCP tool call in this server — but cheap to close properly, so it's now reference-counted instead.

Previously reported to Google VRP as issue 545472983; cleared for public disclosure as this GitHub issue.

Fixes #2567

## Test plan

- A new integration test (`tests/network_blocking.test.ts`) reproduces the disclosed scenario end-to-end against real Chromium: an allowed origin redirects `/robots.txt` to an excluded origin, and the test asserts the excluded origin is never reached and its content never appears in the saved Lighthouse report — manually verified to **fail** against the code without this fix, and to pass with it.
- Unit tests (`tests/utils/guardedNetworkFetch.test.ts`) cover: no-op behavior when unconfigured, install/restore reference equality, redirect-hop validation rejecting a disallowed destination without ever fetching it, cookie forwarding (both the positive path and the opt-out path), pass-through of unrelated CDP traffic, and reentrancy under overlapping installs with multiple simultaneously-active guardrails.
- `npm run typecheck` and the full touched test suite pass locally.

### DIFF
--- a/docs/residual-review-findings/fix-lighthouse-redirect-allowed-url-pattern.md
+++ b/docs/residual-review-findings/fix-lighthouse-redirect-allowed-url-pattern.md
@@ -1,0 +1,24 @@
+# Residual Review Findings
+
+**Branch:** `fix/lighthouse-redirect-allowed-url-pattern`
+**Source:** `ce-code-review` run `20260821-194855-a3bf3374` (issue #2567 fix) + one `ce-simplify-code` pass
+
+## Fixed during review (not residual — included for context)
+
+- **P0 (security, confidence 50) / P1 (reliability) / P2 (maintainability)** — three independent reviewers converged on the same gap: `installGuardedNetworkFetch` patched the shared `CdpCDPSession.prototype.send` via closure-captured save/restore, which is not safe against overlapping installs. Correctness's own investigation confirmed this is **not reachable today** (`ToolHandler`'s `toolMutex` serializes every tool call), but it was cheap to close properly with module-level reference counting, so it was fixed rather than left open.
+- **P1 (reliability)** — the replacement fetch had no timeout (previously bounded by Puppeteer's own protocol timeout). Fixed with a 10s `AbortSignal.timeout` per redirect hop.
+- **P2 (testing)** — the regression test's only guardrail-side assertion was "excluded origin never hit"; it could pass even if Lighthouse never requested the redirecting `robots.txt` at all. Fixed with an `allowedRobotsHit` assertion.
+- **P2 (testing)** — no positive-path test proved cookie forwarding actually builds a `Cookie` header from `Network.getCookies` output (the existing test only exercised the swallow-the-lookup-error fallback). Fixed with a dedicated test.
+- **P2 (project-standards)** — a `handle!` non-null assertion in `src/utils/guardedNetworkFetch.ts`, against `AGENTS.md`'s "do not use `!`" rule. Fixed by narrowing with an explicit `undefined` check instead.
+- Simplify-review pass (reuse/quality/efficiency): reused `createIdGenerator()` instead of a hand-rolled counter, honored `includeCredentials` before forwarding cookies, collapsed a redundant `Map` `has()`+`get()` lookup.
+
+## Residual (not fixed in this PR)
+
+- **P2 — advisory (`no_sink`)** Redirect-hop `Set-Cookie` responses are not applied to the in-flight redirect chain: if an allowed origin's redirect response itself sets a cookie required by the next hop, the guarded fetch never captures or forwards it, so the next hop only sees whatever cookies existed in Chrome's jar *before* the redirect started. A correct fix means maintaining a redirect-local cookie jar (reproducing Chrome's domain/path/secure matching rules) — real design work, out of scope for this security fix. (Found by the cross-model adversarial peer.)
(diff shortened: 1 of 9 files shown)


OrKoN (changes requested):
Adding this complexity of patching the send in cdp sessions is not it here. Lighthouse sends cookies to the server on the same origin, the server would have access to those cookies already.

OrKoN:
Thanks for the PR but we do not plan to address this issue for now and I marked the related issue as such.

PR #2605, opened 2026-08-21
feat: support configurable filesystem roots

### LINKED ISSUE #2598: Add a CLI flag to configure MCP roots

Currently, the file system access can only be configured via MCP. In the next version of the MCP spec, the roots are getting removed.

We should also support a CLI flag like `--filesystem-root` (array) and via `--workspace` (array) keeping the defaults to the os.tmpdir().

Removing roots altogether, discarded since it's useful to set fs guardrails for the mcp server and cli.

### Additional context

_No response_

joebasrawi:
Fixes #2598

This lets people set filesystem roots through the MCP server or the Chrome DevTools CLI. Multiple roots work together with any roots sent by the MCP client, and temp directory access stays the same.

I added tests for the flags and root handling.

### DIFF
--- a/src/bin/chrome-devtools.ts
+++ b/src/bin/chrome-devtools.ts
@@ -35,7 +35,6 @@ import {
   mcpOptions,
   parseArguments,
   getMcpOptionsForViaCli,
-  applyCliFilesystemAccess,
 } from '../config/mcp-options.js';
 
 await checkForUpdates(
@@ -156,7 +155,6 @@ y.command(
     ) {
       argv.headless = true;
     }
-    applyCliFilesystemAccess(argv, hideBin(process.argv));
     const args = serializeArgs(mcpOptions, argv);
     await start(args, argv.sessionId);
     process.exit(0);
--- a/src/bin/chrome-devtools.ts
+++ b/src/bin/chrome-devtools.ts
@@ -35,6 +35,7 @@ import {
   mcpOptions,
   parseArguments,
   getMcpOptionsForViaCli,
+  applyFilesystemRootOverrides,
   checkFilesystemRootConflict,
 } from '../config/mcp-options.js';
 
@@ -161,6 +162,13 @@ y.command(
       !argv.wsEndpoint
     ) {
       argv.headless = true;
+    }
+    applyFilesystemRootOverrides(argv, hideBin(process.argv));
+    // CLI defaults unrestricted paths on. Clear the yargs filesystemRoot default
+    // so it is not serialized onto the daemon as an explicit flag that would
+    // then conflict with --allow-unrestricted-paths.
+    if (argv.allowUnrestrictedPaths === true) {
+      argv.filesystemRoot = undefined;
     }
     const args = serializeArgs(mcpOptions, argv);
     await start(args, argv.sessionId);
--- a/src/bin/chrome-devtools.ts
+++ b/src/bin/chrome-devtools.ts
@@ -35,6 +35,7 @@ import {
   mcpOptions,
   parseArguments,
   getMcpOptionsForViaCli,
+  checkFilesystemRootConflict,
 } from '../config/mcp-options.js';
 
 await checkForUpdates(
@@ -128,6 +129,12 @@ y.command(
   y =>
     y
       .options(getCliOptions())
+      .check(argv => {
+        return checkFilesystemRootConflict(
+          argv.allowUnrestrictedPaths,
+          hideBin(process.argv),
+        );
+      })
       .example(
         '$0 start --browserUrl http://localhost:9222',
         'Start the server connecting to an existing browser',
--- a/src/bin/chrome-devtools.ts
+++ b/src/bin/chrome-devtools.ts
@@ -35,8 +35,7 @@ import {
   mcpOptions,
   parseArguments,
   getMcpOptionsForViaCli,
-  applyFilesystemRootOverrides,
-  checkFilesystemRootConflict,
+  applyCliFilesystemAccess,
 } from '../config/mcp-options.js';
 
 await checkForUpdates(
@@ -130,12 +129,6 @@ y.command(
   y =>
     y
       .options(getCliOptions())
-      .check(argv => {
-        return checkFilesystemRootConflict(
-          argv.allowUnrestrictedPaths,
-          hideBin(process.argv),
-        );
-      })
       .example(
         '$0 start --browserUrl http://localhost:9222',
         'Start the server connecting to an existing browser',
@@ -163,13 +156,7 @@ y.command(
     ) {
       argv.headless = true;
     }
-    applyFilesystemRootOverrides(argv, hideBin(process.argv));
-    // CLI defaults unrestricted paths on. Clear the yargs filesystemRoot default
-    // so it is not serialized onto the daemon as an explicit flag that would
-    // then conflict with --allow-unrestricted-paths.
-    if (argv.allowUnrestrictedPaths === true) {
-      argv.filesystemRoot = undefined;
-    }
+    applyCliFilesystemAccess(argv, hideBin(process.argv));
     const args = serializeArgs(mcpOptions, argv);
     await start(args, argv.sessionId);
     process.exit(0);
--- a/src/config/mcp-options.ts
+++ b/src/config/mcp-options.ts
@@ -406,8 +406,7 @@ export const mcpOptions = {
     default: [os.tmpdir()],
     defaultDescription: 'OS temp directory',
     describe:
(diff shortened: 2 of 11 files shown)


src/config/mcp-options.ts
```diff
@@ -374,6 +374,13 @@ export const mcpOptions = {
       'no roots are configured. Use this only when connecting a trusted local client that does not implement ' +
       'MCP roots and requires access to paths outside the temp directory.',
   },
+  filesystemRoot: {
+    type: 'array',
+    alias: 'workspace',
+    describe:
+      'A directory that filesystem tools are allowed to access. May be specified more than once. ' +
+      'The OS temp directory is always allowed.',
```
OrKoN: let's set the defaults to os temp dir here, by using yargs default value option

src/index.ts
```diff
@@ -71,10 +73,25 @@ export async function createMcpServer(
...
-  let lastRoots: Root[] | undefined;
+  const configuredRoots = (serverArgs.filesystemRoot ?? []).map(root => {
+    const rootPath = path.resolve(String(root));
+    return {
+      uri: pathToFileURL(rootPath).href,
+      name: path.basename(rootPath) || rootPath,
+    };
+  });
+
+  // Client roots stay valid across browser reconnects and only the client can
```
OrKoN: I wonder if it is better if MCP-client configured roots override the CLI-configured roots. Otherwise it might be confusing for the MCP clients.

src/config/mcp-options.ts
```diff
@@ -374,6 +374,13 @@ export const mcpOptions = {
       'no roots are configured. Use this only when connecting a trusted local client that does not implement ' +
       'MCP roots and requires access to paths outside the temp directory.',
   },
+  filesystemRoot: {
```
OrKoN, line 405: we should probably make it conflict with allowUnrestrictedPaths. I think allowUnrestrictedPaths could be deprecated in favor of specifying `--workspace=/`.

src/index.ts
OrKoN: actually, I think it might be better if the CLI args are always included

src/config/mcp-options.ts
joebasrawi: Updated. `filesystemRoot` now uses `os.tmpdir()` as its yargs default. I also added a stable default description so generated docs show `OS temp directory` instead of a machine-specific path.

src/index.ts
joebasrawi: Kept the configured CLI roots in the combined list, so they are always included alongside roots reported by the MCP client.

src/config/mcp-options.ts
joebasrawi: Added the conflict and deprecated `--allow-unrestricted-paths` in favor of `--workspace=/`. The old flag still works by itself for compatibility, but it is rejected when an explicit filesystem root is also supplied.

joebasrawi:
Rebased onto current `main` (conflicts resolved). The Aug 22 notes are still in: `os.tmpdir()` as the yargs default, CLI roots always included with client roots, and `--allow-unrestricted-paths` conflicts with an explicit filesystem root (deprecated in favor of `--workspace=/`).

@OrKoN ready for another look when you have a minute.

src/config/mcp-options.ts
```diff
@@ -374,16 +375,95 @@ export const mcpOptions = {
...
+    type: 'array',
+    alias: 'workspace',
+    default: [os.tmpdir()],
+    defaultDescription: 'OS temp directory',
+    describe:
+      'A directory that filesystem tools are allowed to access. May be specified more than once. ' +
+      'The OS temp directory is always allowed.',
```
OrKoN:
```suggestion
```

I think we should remove this line as it's not true anymore: the os.tmpdir() is the default but it is not always allowed.

OrKoN:
I think it looks good to me. @Lightning00Blade WDYT?

joebasrawi:
The `filesystemRoot` describe line no longer says the OS temp directory is always allowed — `os.tmpdir()` is the default, not an extra always-allowed root. Rebased onto current `main` for @Lightning00Blade.

tests/cli.test.ts
```diff
@@ -150,6 +152,44 @@ describe('cli args parsing', () => {
     });
   });
 
+  it('parses filesystem roots', async () => {
```
Lightning00Blade: Let's wrap in a describe so we can easily find and collapse.

src/index.ts
```diff
@@ -244,17 +269,20 @@ export class McpServer {
...
         // holding the tool mutex, and a later background refresh still lands
         await this.#updateRoots(ROOTS_REQUEST_TIMEOUT);
       } else {
         // Carry the known roots over and refresh out of band, so a reconnect
         // never pays for a client round-trip
-        this.#context.setRoots(this.#lastRoots);
         void this.#updateRoots();
       }
     }
+    if (this.#context === undefined) {
```
Lightning00Blade: Why is this check needed? Currently TypeScript correctly deduces that this is set.

src/index.ts
```diff
@@ -68,6 +71,15 @@ export class McpServer {
   ) {
     this.#serverArgs = serverArgs;
     this.#options = options;
+    this.#configuredRoots = (
```
Lightning00Blade: We already store the system args, can we just do this on the `#combinedRoots` that way we don't store this information 2 times.

src/config/mcp-options.ts
Lightning00Blade: Instead of having this custom logic can we do the following:
-> Unset the `--allow-unrestricted-paths` for CLI default to true 
-> In the middleware check if we have any `--workspace` set, and pass them along
-> If not we set manually the  `--allow-unrestricted-paths` to true

Lightning00Blade (commented):
Overall looks good, please address the comments and rebase.

joebasrawi:
Comments addressed and rebased onto current main. @Lightning00Blade @OrKoN

src/bin/chrome-devtools.ts
```diff
@@ -155,6 +156,7 @@ y.command(
     ) {
       argv.headless = true;
     }
+    applyCliFilesystemAccess(argv, hideBin(process.argv));
```
Lightning00Blade: This is redundant as we already have done this via `getMcpOptionsForViaCli`.

src/config/mcp-options.ts
```diff
@@ -391,25 +392,67 @@ export const mcpOptions = {
...
     describe:
       'If set, disables the default path restriction that applies when the MCP client does not negotiate ' +
       'the roots capability. By default, file-writing tools are restricted to the OS temp directory when ' +
       'no roots are configured. Use this only when connecting a trusted local client that does not implement ' +
       'MCP roots and requires access to paths outside the temp directory.',
   },
+  filesystemRoot: {
+    type: 'array',
+    alias: 'workspace',
+    default: [os.tmpdir()],
```
Lightning00Blade: Extract this into a `const DEFAULT_FILESYSTEM_ROOT = [os.tmpdir()]`;

src/config/mcp-options.ts
```diff
@@ -469,6 +508,9 @@ export function parser(
     .options(options)
     .showHelpOnFail(false, 'Specify --help for available options')
     .middleware(args => {
+      if (isViaCli) {
+        applyCliFilesystemAccess(args, hideBin(argv));
+      }
```
Lightning00Blade: Yargs reuses the object so we can simply do:
```suggestion
      if (isViaCli && args.filesystemRoot === DEFAULT_FILESYSTEM_ROOT) {
        args.allowUnrestrictedPaths = true;
        args.filesystemRoot  = undefined;
      }
```

joebasrawi:
Latest Lightning nits are in: dropped the redundant chrome-devtools filesystem middleware, extracted DEFAULT_FILESYSTEM_ROOT, and switched CLI middleware to the yargs default-object identity check.

src/bin/chrome-devtools.ts
joebasrawi: Removed. CLI defaults now come from `getMcpOptionsForViaCli` and the parser middleware.

src/config/mcp-options.ts
joebasrawi: Extracted `DEFAULT_FILESYSTEM_ROOT` and used it as the yargs default.

src/config/mcp-options.ts
joebasrawi: Applied the identity check on the reused yargs default object.

OrKoN (approved):

Lightning00Blade (approved):

PR #2612, opened 2026-08-24
feat: add get_css_style tool

samiyac:
Add `get_css_styles` tool which exposes matched styles and the CSS cascade to empower accurate CSS debugging workflows.

It integrates `CssFormatter` into `McpPage` and `McpResponse`, adding container query resolution and rule pagination support.

### DIFF
--- a/src/McpPage.ts
+++ b/src/McpPage.ts
@@ -78,7 +78,7 @@ import {
   type Page,
   type ConsoleMessage,
   type HTTPRequest,
-  type DevTools,
+  DevTools,
   type JSONSchema7Definition,
 } from './third_party/index.js';
 import {takeSnapshot} from './tools/snapshot.js';
@@ -88,6 +88,7 @@ const NAVIGATION_TIMEOUT = 10_000;
 import type {
   ContextPage,
   DevToolsData,
+  MatchedStyles,
   Response,
 } from './tools/ToolDefinition.js';
 import type {
@@ -102,6 +103,12 @@ import {
   type WaitForEventsResult,
   type DialogAction,
 } from './utils/WaitForHelper.js';
+
+function isBackendNodeId(
+  id: unknown,
+): id is DevTools.Protocol.DOM.BackendNodeId {
+  return typeof id === 'number';
+}
 
 /**
  * Per-page state wrapper. Consolidates dialog, snapshot, emulation,
@@ -746,6 +753,71 @@ export class McpPage implements ContextPage {
     } catch {
       return undefined;
     }
+  }
+
+  async getMatchedStylesForUid(uid: string): Promise<MatchedStyles> {
+    if (!this.textSnapshot) {
+      throw new Error(
+        `No snapshot found for page ${this.id ?? '?'}. Use ${takeSnapshot.name} to capture one.`,
+      );
+    }
+    const node = this.textSnapshot.idToNode.get(uid);
+    if (!node) {
+      throw new Error(`Element uid "${uid}" not found on page ${this.id}.`);
+    }
+
+    const backendNodeId = node.backendNodeId;
+    if (!isBackendNodeId(backendNodeId)) {
+      throw new Error(
+        `Failed to resolve backend node ID for element with uid "${uid}".`,
+      );
+    }
+
+    if (!this.#devtoolsUniverse) {
+      throw new Error(
+        `DevTools universe is not available for page ${this.id ?? '?'}.`,
+      );
+    }
+
+    const targetManager = this.#devtoolsUniverse.universe.context.get(
+      DevTools.TargetManager,
+    );
+    let domNode: DevTools.DOMModel.DOMNode | undefined;
+    let cssModel: DevTools.CSSModel.CSSModel | null = null;
+
+    for (const dom of targetManager.models(DevTools.DOMModel.DOMModel)) {
+      const nodeMap = await dom.pushNodesByBackendIdsToFrontend(
+        new Set([backendNodeId]),
+      );
+      const frontendNode = nodeMap?.get(backendNodeId);
+      if (frontendNode) {
+        domNode = frontendNode;
+        cssModel = dom.target().model(DevTools.CSSModel.CSSModel);
+        break;
+      }
+    }
+
+    if (!domNode || !cssModel) {
+      throw new Error(
+        `Element with uid "${uid}" was detached or no longer exists on the page. Please take a new snapshot with ${takeSnapshot.name}.`,
+      );
+    }
+
+    const targetElement = domNode.enclosingElementOrSelf();
+    if (!targetElement) {
+      throw new Error(
+        `Element with uid "${uid}" is not an element node and has no parent element.`,
+      );
+    }
+
+    const matchedStyles = await cssModel.getMatchedStyles(targetElement.id);
+    if (!matchedStyles) {
+      throw new Error(
+        `Could not retrieve matched styles for element with uid "${uid}".`,
+      );
+    }
+
+    return matchedStyles;
   }
 
   async getDevToolsData(): Promise<DevToolsData> {
--- a/src/formatters/CssFormatter.ts
+++ b/src/formatters/CssFormatter.ts
@@ -1177,3 +1177,70 @@ export class CssFormatter {
     };
   }
 }
+
+/**
+ * Resolves container element details for all `@container` queries impacting the node.
+ *
+ * Container queries evaluate against an ancestor container element in the DOM tree.
+ * The DevTools SDK performs an asynchronous CDP request (`CSS.getContainerForNode`)
+ * to discover the concrete container node for each query.
+ *
+ * Resolving these upfront in parallel allows `CssFormatter` methods to remain purely
+ * synchronous while providing container selectors and UIDs in headers and comments.
+ */
+export async function resolveContainerQueries(
+  matchedStyles: MatchedStyles,
+  resolveUid?: UidResolver,
+): Promise<Map<ContainerQuery, ResolvedContainerDetails>> {
+  const resolved = new Map<ContainerQuery, ResolvedContainerDetails>();
+  const targetNode = matchedStyles.node?.();
+  if (!targetNode) {
+    return resolved;
+  }
+  const queries = new Set<ContainerQuery>();
+
+  const allStyles = [...(matchedStyles.nodeStyles?.() ?? [])];
+  for (const name of matchedStyles.customHighlightPseudoNames?.() ?? []) {
+    allStyles.push(
+      ...(matchedStyles.customHighlightPseudoStyles?.(name) ?? []),
+    );
+  }
+  for (const type of matchedStyles.pseudoTypes?.() ?? []) {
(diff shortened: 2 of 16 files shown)


src/tools/css.ts
```diff
@@ -0,0 +1,359 @@
...
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {zod} from '../third_party/index.js';
+
+import {ToolCategory} from './categories.js';
+import {definePageTool, type MatchedStyles} from './ToolDefinition.js';
+
+const INHERITED_PROPERTIES = new Set([
```
OrKoN: I think this should exist in DevTools? we should re-use it, or introduce it there.

src/tools/css.ts
```diff
@@ -0,0 +1,359 @@
...
+    return false;
+  }
+  return (
+    typeof rule['selectorText'] === 'function' &&
+    typeof rule['isUserAgent'] === 'function' &&
+    typeof rule['lineNumberInSource'] === 'function'
+  );
+}
+
+function formatCssStyles(matchedStyles: MatchedStyles, uid: string): string {
```
OrKoN: can we re-use any formatters on the DevTools side? If not (even by refactoring presenters), we should extract this in a separate formatter in this repo.

src/McpPage.ts
```diff
@@ -671,6 +695,63 @@ export class McpPage implements ContextPage {
...
+
+    const devtools = await this.ensureDevToolsUniverse();
+    const domModel = devtools.target.model(DevTools.DOMModel.DOMModel);
+    const cssModel = devtools.target.model(DevTools.CSSModel.CSSModel);
+    if (!domModel || !cssModel) {
+      throw new Error('DevTools DOMModel or CSSModel is not available.');
+    }
+
+    await domModel.requestDocument();
+    const nodeMap = await domModel.pushNodesByBackendIdsToFrontend(
```
OrKoN: we should not need to push nodes to the frontend. Can we just use the CSS logic relying on backend node IDs only?

OrKoN (changes requested):
Thanks! Left a few comments on the overall approach.

src/McpPage.ts
```diff
@@ -671,6 +695,63 @@ export class McpPage implements ContextPage {
...
+      backendNodeId = await handle.backendNodeId();
+    }
+    if (!backendNodeId || !isBackendNodeId(backendNodeId)) {
+      throw new Error(
+        `Failed to resolve backend node ID for element with uid "${uid}".`,
+      );
+    }
+
+    const devtools = await this.ensureDevToolsUniverse();
+    const domModel = devtools.target.model(DevTools.DOMModel.DOMModel);
```
OrKoN: uid might belong to iframes for we need to find the right target instance

src/McpPage.ts
```diff
@@ -671,6 +695,63 @@ export class McpPage implements ContextPage {
...
+      using handle = await this.#resolveElementHandle(node, uid);
+      backendNodeId = await handle.backendNodeId();
+    }
+    if (!backendNodeId || !isBackendNodeId(backendNodeId)) {
+      throw new Error(
+        `Failed to resolve backend node ID for element with uid "${uid}".`,
+      );
+    }
+
+    const devtools = await this.ensureDevToolsUniverse();
```
OrKoN: do we need to ensure the universe here? I think it should automatically exist when the page is available. Let's throw an error if it does not exist instead?

src/McpPage.ts
```diff
@@ -671,6 +695,63 @@ export class McpPage implements ContextPage {
...
+      );
+    }
+    const node = this.textSnapshot.idToNode.get(uid);
+    if (!node) {
+      throw new Error(`Element uid "${uid}" not found on page ${this.id}.`);
+    }
+
+    let backendNodeId = node.backendNodeId;
+    if (!backendNodeId) {
+      using handle = await this.#resolveElementHandle(node, uid);
```
OrKoN: could we resolve without creating a handle?

src/tools/css.ts
samiyac: using `DevTools.CSSMetadata.cssMetadata().isPropertyInherited` instead

src/tools/css.ts
samiyac: Extracted into a separate formatter

src/McpPage.ts
samiyac: Updated to check

src/McpPage.ts
samiyac: Done, only throws an error

src/McpPage.ts
samiyac: Done

src/formatters/CssFormatter.ts
```diff
@@ -0,0 +1,663 @@
...
+          properties: u.properties.map(p => this.#formatStructuredProperty(p)),
+        })),
+      });
+    }
+
+    return result;
+  }
+
+  toString(): string {
+    const meta = getNodeMetadata(this.#matchedStyles.node());
```
OrKoN: toString() should use toJSON() output to avoid duplication

src/formatters/CssFormatter.ts
```diff
@@ -0,0 +1,645 @@
...
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {DevTools} from '../third_party/index.js';
+import type {MatchedStyles} from '../tools/ToolDefinition.js';
+import type {PaginationOptions} from '../types.js';
+import type {PaginationResult} from '../utils/pagination.js';
+import {paginate} from '../utils/pagination.js';
```
OrKoN: I think we should perform pagination in McpResponse  (see #dataWithPagination).

tests/formatters/CssFormatter.test.ts
```diff
@@ -0,0 +1,342 @@
...
+    // Page 1 with pageSize 2
+    const formatter1 = new CssFormatter(matchedStyles, {
+      uid: 'btn-1',
+      pageSize: 2,
+      pageIdx: 1,
+    });
+    const output1 = formatter1.toString();
+    assert.ok(output1.includes('Showing 3-3 of 3 style chunks (Page 2 of 2).'));
+    assert.ok(output1.includes('.rule-3 (from app.css:31:1)'));
+    assert.ok(output1.includes('Previous page: 0'));
```
OrKoN: let's use snapshots for all formatter tests?

src/formatters/CssFormatter.ts
OrKoN: Also, i think we should paginate by rules and not by properties.

tests/tools/css.test.ts
```diff
@@ -0,0 +1,406 @@
...
+          id="child-frame"
+          src="/iframe_content.html"
+        ></iframe>
+      `,
+    );
+
+    await withMcpContext(async (_, context) => {
+      const page = context.getSelectedMcpPage().pptrPage;
+      await page.goto(server.getRoute('/iframe_host.html'));
+      await page.waitForSelector('iframe');
```
OrKoN: This could be racy potentially, let's use waitForFrame plus waitForSelector on the frame to make sure element is loaded.

tests/formatters/CssFormatter.test.ts
samiyac: Done

src/formatters/CssFormatter.ts
samiyac: Yeah, the reason `#dataWithPagination` was not being used was because i was trying to paginate on rules and properties. Now that we are only paginating by rules, `#dataWithPagination` is being used.

src/formatters/CssFormatter.ts
samiyac: Updated

src/formatters/CssFormatter.ts
```diff
@@ -0,0 +1,337 @@
...
+  }
+  if (parsed.isBlobURL()) {
+    return '(blob)';
+  }
+  return (
+    DevTools.Common.ParsedURL.ParsedURL.extractName(sheetUrl) || '(inline)'
+  );
+}
+
+function getSourceLocation(rule: StyleRule): string {
```
OrKoN: I feel like this code is also available in DevTools

tests/tools/css.test.ts
```diff
@@ -0,0 +1,512 @@
...
+  );
+  for (const rule of matchedStyles.rules) {
+    if (isMatchedRule(rule) && (!predicate || predicate(rule))) {
+      return rule;
+    }
+  }
+  throw new Error('Expected to find a matching rule');
+}
+
+describe('css', () => {
```
OrKoN: I think a lot of the tests here need to be in McpPage.test.ts as they are basically testing getMatchedStylesForUid (given that we test formatting separately).

OrKoN: I think for tool tests we should just use recently introduced mocks (see AGENTS.md) and just verify that the tool code correctly passes the data to McpResponse (1 or 2 tests probably).

src/formatters/CssFormatter.ts
samiyac: As discussed offline, there is code available in DevTools but it is part of elements panel codebase, so not replacing this and other such instances for now

tests/tools/css.test.ts
samiyac: Updated

OrKoN:
A few comments from AI-review:

Code Style & Guidelines Compliance (AGENTS.md)

  #### 1. TypeScript Rules: as Keyword Usage

  AGENTS.md specifies:

  │ • Do not use as keyword for type casting.

  In CssFormatter.test.ts, as unknown as ... is used repeatedly:

  • createMockNode: as unknown as DevTools.DOMModel.DOMNode
  • createMockProperty: as unknown as StyleProperty
  • createMockStyle: as unknown as NodeStyle
  • createMockMatchedStyles: as unknown as MatchedStyles

  #### 2. Mocking Guidelines

  AGENTS.md specifies:

  │
  │ • Use sinon.createStubInstance(Class): Do not hand-roll mock objects or define custom mock interfaces.
  │ • Centralize mocks in tests/mocks.ts.

  In CssFormatter.test.ts, mock factories (createMockMatchedStyles, createMockRule, createMockStyle, createMockProperty) are hand-rolled using Object.create(Prototype) and contain simulated state logic (e.g.
  isInherited: (style) => inheritedStylesSet.has(style)). They should leverage sinon.createStubInstance().
  ──────
  ### Test Suite Performance

  In McpPage.test.ts:

  • There are 4 separate withMcpContext tests:
      1. retrieves matched styles for a page element
      2. retrieves matched styles for nodes inside open shadow roots
      3. retrieves matched styles for nodes inside closed shadow roots
      4. retrieves matched styles for nodes inside iframes
  • Each test launches a separate Chromium process, opens a route, and creates a snapshot.
  • Suggestion: Combine these test cases into a single HTML route (containing regular elements, open shadow root, closed shadow root, and an iframe). Testing all 4 scenarios in one browser session could save
  multiple seconds on test execution.

OrKoN (approved):

PR #2623, opened 2026-08-27
feat: add Agent Plugins 1.0 package

777genius:
## Summary

- Add an Agent Plugins 1.0 package at the repository root.
- Expose the existing stdio server through a version-pinned `mcp.json`.
- Keep the package version and MCP npm pin synchronized through release-please.
- Use `npx --prefix ${PLUGIN_DATA}` to avoid resolving the repository's unbuilt local npm package instead of the published server. This keeps compatibility with clients that cannot represent an explicit working directory.
- Document installation through Agent Plugins CLI.

## Verification

- `plugin.json` and `mcp.json` pass the Agent Plugins 1.0 schemas.
- Exact head `02372c4d47ad257773b2b80c7d7fd056d7067be0` was tested using local package bytes with the released Agent Plugins CLI 0.1.24.
- Isolated add, doctor, repair, and remove with data cleanup passed for Claude Code, Gemini CLI, OpenCode, Cline, and Windsurf. Configuration-only clients are not claimed as live model tests.
- OpenCode 1.18.25 with its built-in `opencode/big-pickle` model successfully called `new_page` and `take_snapshot`, recovering a random marker from a loopback test page. The marker was absent from the prompt, and other tools were denied.
- Gemini CLI 0.57.0 reported the MCP server as connected. Its model call was blocked by Google account eligibility, so Gemini model runtime is not claimed.
- Package tree: `sha256:50145f0cfd0e4095cfb00fc74755c0c930dc8258952c868de5dc81b54765c1a0`.
- Manifest: `sha256:b34a4dcd71cd536a7f5a3a51d76d53ae5af3d0ce0f18783e71c7f01da865b867`.
- Sanitized runtime evidence: https://github.com/777genius/universal-agent-plugins/blob/a022a505df4e9b6627d911d935c5fff511451ecd/tests/e2e/results/chrome-devtools-live-clients-2026-08-31.json

The earlier exact-head lifecycle evidence, including real Claude CLI plugin discovery, remains available at https://github.com/777genius/universal-agent-plugins/blob/main/tests/e2e/results/agentplugins-chrome-devtools-multiclient-2026-08-30.json. It applies to the previously recorded head, not a new runtime claim for this revision.

No real user project or browser profile was used. The community directory short alias is being expanded separately; the README keeps the direct repository command until that publication is available.


OrKoN:
@natorion do you know about https://agent-plugins.org/? should we add this plugin configuration?

mcp.json
```diff
@@ -0,0 +1,10 @@
+{
```
OrKoN, line 1: Should this be in the root of the repository so that skills are picked up?

natorion:
> @natorion do you know about https://agent-plugins.org/? should we add this plugin configuration?

I think we should!

mcp.json
777genius: Yes, good catch. I moved plugin.json and mcp.json to the repository root in cb39d1d835c3baa3eff87501cd8c1de020604789 and updated the release-please paths. The root package passes Agent Plugins 1.0 validation and an isolated add/remove lifecycle for Codex, Cursor, and Kiro.

README.md
```diff
@@ -67,6 +67,15 @@ You can disable these update checks by setting the `CHROME_DEVTOOLS_MCP_NO_UPDAT
 
 ## Getting started
 
+> [!TIP]
+> Install the [Agent Plugins 1.0](https://agent-plugins.org/specification)
+> package with one command for Claude Code, Codex, Cursor, Copilot CLI/VS Code,
+> Kiro, Gemini CLI, OpenCode, Cline, or Windsurf:
+>
+> ```bash
+> npx universal-agent-plugins add ChromeDevTools/chrome-devtools-mcp
```
OrKoN: We would not recommend third party installation packages from our readme, let's remove this and update docs for each MCP client separate instead (using that client's documented workflows).  Is there a registry for Agent Plugins 1.0 plugins?

777genius: Makes sense. I removed the third-party installer recommendation in 5b5325c and left the existing per-client setup flow as the README entry point. I could not find an official registry operated by the Agent Plugins project. The 1.0 documentation defines the portable package format and explicitly leaves distribution, installation, permissions, and UX to clients. Independent community directories exist, but I will not recommend one from this README.

OrKoN: Thanks!

777genius:
Correction: the Windows timing failure is unrelated to the Agent Plugins package. I removed the test stabilization from this PR in 59380dd so the change is scoped again to plugin.json, mcp.json, and release configuration. The standalone test fix is now #2635, with 20 consecutive focused repetitions passing.

OrKoN (approved):
LGTM thanks!

PR #2625, opened 2026-08-28
fix: do not enable the DevTools frontend Audits subscription

### LINKED ISSUE #2556: Large Audits backlog starves all page-facing tool calls
### Description of the bug

When `chrome-devtools-mcp` connects to a long-lived Chrome session whose page targets have accumulated a large DevTools Audits backlog, page-facing MCP tools consistently time out even though the browser-level CDP endpoint and every individual page target remain responsive.

In the affected session, the MCP process repeatedly writes:

```text
No handler registered for issue code PerformanceIssue
```

This looks like a regression or broader consequence of #1850 / #1852: the warning is present again in v1.7.0, and a sufficiently large initial `Audits.issueAdded` replay appears to starve normal tool requests. It is also related to the eager all-tab work discussed in #1921, but reproduces with only 32 page targets rather than hundreds or thousands.

The following page-facing tools were tested and all timed out in the affected MCP session:

- `list_pages`
- `evaluate_script`
- `list_network_requests`

An MCP-level ping can still succeed while these tools are unavailable.

#### Diagnostic evidence

- Chrome exposed 32 normal page targets.
- Direct browser-level CDP `Target.getTargets`, `Target.attachToTarget`, and `Runtime.evaluate` succeeded for every page; each page probe completed in under one second.
- Enabling the `Audits` domain directly replayed 17,357 `Audits.issueAdded` events, mostly `CookieIssue`, `GenericIssue`, and `PerformanceIssue`.
- After a five-second warm-up, only five new audit events arrived during the next ten seconds. This indicates that the large burst was primarily retained history rather than continuous page traffic.
- A public trigger was verified separately on `https://example.com`: 500 synchronous `document.cookie` reads produced exactly 500 `PerformanceIssue/DocumentCookie` events.

No page target was crashed, discarded, or unresponsive during the direct-CDP control test.

### Reproduction

Public-only reproduction shape:

1. Start Chrome with remote debugging enabled.
2. Open multiple tabs at `https://example.com`.
3. In each tab, run the following in DevTools Console to accumulate `PerformanceIssue/DocumentCookie` entries:

   ```js
   for (let i = 0; i < 1000; i++) {
     void document.cookie;
   }
   ```

   Repeat across enough tabs to build an Audits backlog of roughly 15,000-20,000 events. The backlog size can be confirmed through a separate raw-CDP session by enabling `Audits` and counting `Audits.issueAdded` events.

4. Connect `chrome-devtools-mcp` to that Chrome instance using either `--autoConnect` or a browser WebSocket endpoint.
5. Call `list_pages`.
6. Call a page-specific tool such as `evaluate_script` with a trivial function.

Observed behavior:

- Tool calls do not return before the client's 60-second timeout.
- stderr repeatedly reports the unhandled `PerformanceIssue` warning.
- Direct CDP calls to the same browser and pages continue to complete normally.
- Restarting only the MCP process does not remove the retained Audits backlog, so reconnecting replays it again.

### Expectation

A large retained Audits backlog, including unknown or unsupported issue codes, should not block unrelated MCP tool calls.

Possible safeguards include bounding/deduplicating initial issue replay, dropping unsupported issue codes before mapping/logging, and ensuring Audits processing cannot starve tool requests.

### MCP configuration

```json
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": [
        "-y",
        "chrome-devtools-mcp@1.7.0",
        "--autoConnect"
      ]
    }
  }
}
```

The same behavior was observed with an explicitly supplied loopback `--wsEndpoint`; the endpoint identifier is intentionally omitted.

1.7.0

### Chrome version

151.0.7922.76

### Coding agent version

N/A. The failure was observed at the MCP tool-call boundary and compared against raw CDP calls.

### Model version

N/A

### Chat log

N/A

### Node version

22.22.2

### Operating system

macOS 26.5.1

### Extra checklist

mvanhorn:
## Summary

Puppeteer already subscribes to `Audits.issueAdded`. `overrideDevToolsGlobals` did not stop the DevTools frontend from enabling its own `Audits` domain, so a second consumer of the same stream was created on session setup. Against a target holding a large retained issue backlog, that second subscription replays the entire backlog through `IssuesManager`, logging `No handler registered for issue code PerformanceIssue` per unsupported entry and delaying unrelated page work.

This stubs `Audits.invoke_enable` on the agent prototype exactly the way the Network emulation commands are already stubbed, so the redundant subscription is never enabled. The four identical no-op command bodies are collapsed into one `noopAgentCommand` helper while adding the fourth.

## Why this matters

Closes #2556

The guard added in #2535 covers `PageEventSubscriber.#onIssueAdded`, the collector's own subscription. It does not cover the frontend's `IssuesManager` path, which is the other consumer of the same CDP event, so the warning and the backlog replay both survive it. Fixing it at the `invoke_enable` boundary means the frontend never receives the replay at all, rather than filtering each issue after the fact.

## Testing

`node scripts/test.js tests/devtools/DevtoolsUtils.test.ts` -> 20 passing. `npm run build` and `npm run typecheck` both clean.

The added case builds a real 500-issue `PerformanceIssue` backlog on a page, detaches, then constructs the target universe and asserts the warning is never emitted while `Runtime.evaluate` still works.

It pins the defect rather than just covering the path. Reverting only `src/devtools/DevtoolsUtils.ts`, rebuilding, and re-running:

```
warn('No handler registered for issue code PerformanceIssue') at createIssuesFromProtocolIssue
  (build/third_party/devtools-frontend/front_end/models/issues_manager/IssuesManager.js:164:13)
  ... (repeated)
at assert2.<computed> [as neverCalledWithMatch]
✖ createTargetUniverse (1533.363458ms)
  '1 subtest failed'
```

Not tested: no measurement of the tool-call timeout the report describes on a long-lived session. The test asserts the backlog is no longer replayed, which is the mechanism, not the end-to-end latency.

AI was used for assistance.

### DIFF
--- a/src/devtools/DevtoolsUtils.ts
+++ b/src/devtools/DevtoolsUtils.ts
@@ -36,6 +36,12 @@ export function overrideDevToolsGlobals({
   // DevTools CDP errors can get noisy.
   DevTools.ProtocolClient.InspectorBackend.test.suppressRequestErrors = true;
 
+  const noopAgentCommand = () => {
+    return Promise.resolve({
+      getError: () => undefined,
+    });
+  };
+
   // Stub out Network emulation commands on the DevTools Agent prototype globally.
   // This prevents the DevTools Frontend from ever resetting/clearing Puppeteer's
   // active network blocking/throttling rules during target setup or session lifetime.
@@ -63,42 +69,42 @@ export function overrideDevToolsGlobals({
       networkAgentPrototype,
       'invoke_overrideNetworkState',
       {
-        value: () => {
-          return Promise.resolve({
-            getError: () => undefined,
-          });
-        },
+        value: noopAgentCommand,
         writable: true,
         configurable: true,
         enumerable: true,
       },
     );
     Object.defineProperty(networkAgentPrototype, 'invoke_enable', {
-      value: () => {
-        return Promise.resolve({
-          getError: () => undefined,
-        });
-      },
+      value: noopAgentCommand,
       writable: true,
       configurable: true,
       enumerable: true,
     });
     Object.defineProperty(networkAgentPrototype, 'invoke_disable', {
-      value: () => {
-        return Promise.resolve({
-          getError: () => undefined,
-        });
-      },
+      value: noopAgentCommand,
       writable: true,
       configurable: true,
       enumerable: true,
     });
     Object.defineProperty(networkAgentPrototype, 'invoke_setBlockedURLs', {
-      value: () => {
-        return Promise.resolve({
-          getError: () => undefined,
-        });
-      },
+      value: noopAgentCommand,
+      writable: true,
+      configurable: true,
+      enumerable: true,
+    });
+  }
+
+  // Puppeteer already collects issues from its own Audits subscription. Avoid
+  // enabling the DevTools Frontend's redundant subscription, which can replay
+  // a large retained issue backlog and delay unrelated page work.
+  const auditsAgentPrototype =
+    DevTools.ProtocolClient.InspectorBackend.inspectorBackend.agentPrototypes.get(
+      'Audits',
+    );
+  if (auditsAgentPrototype) {
+    Object.defineProperty(auditsAgentPrototype, 'invoke_enable', {
+      value: noopAgentCommand,
       writable: true,
       configurable: true,
       enumerable: true,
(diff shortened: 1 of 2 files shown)


OrKoN (approved):
lgtm thanks!

PR #2627, opened 2026-08-28
feat: add an option to turn off js execution tools

OrKoN:
While we keep the evaluate script available by default since it is useful for many users, we give an option to disable it.


Lightning00Blade (approved):
LGTM, but revert the change to the sub module

PR #2628, opened 2026-08-28
feat: option to disable source maps

OrKoN:
allowing to disable DevTools source maps similar to what DevTools settings allow in the UI.

### DIFF
--- a/src/devtools/DevtoolsUtils.ts
+++ b/src/devtools/DevtoolsUtils.ts
@@ -137,8 +137,13 @@ export interface TargetUniverse {
   session: CDPSession;
 }
 
+export interface CreateTargetUniverseOptions {
+  sourceMaps?: boolean;
+}
+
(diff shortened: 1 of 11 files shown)


src/devtools/DevtoolsUtils.ts
```diff
@@ -137,8 +137,29 @@ export interface TargetUniverse {
   session: CDPSession;
 }
 
+export const jsSourceMapsSettingDescriptor: DevTools.Common.Settings.SettingDescriptor<boolean> =
```
szuend: Can we use the exported descriptor from SDK instead?

OrKoN: https://chromium-review.git.corp.google.com/c/devtools/devtools-frontend/+/8314067

szuend (approved):

PR #2629, opened 2026-08-29
feat(cli): persist startup defaults from XDG config

### LINKED ISSUE #2388: Persist CLI defaults for automatic daemon startup

joebasrawi:
## Summary

The standalone `chrome-devtools` CLI auto-starts its daemon when the first tool command runs, but that path had no way to reuse user-specific startup options (`executablePath`, `headless`, `userDataDir`, and other `start` flags). After a reboot, crash, or `stop`, the next command launched the daemon with built-in defaults.

This adds an XDG-compatible config file:

```text
$XDG_CONFIG_HOME/chrome-devtools/config.json
```

falling back to `~/.config/chrome-devtools/config.json`. For example:

```json
{
  "executablePath": "/usr/bin/chromium",
  "headless": false
}
```

Precedence is:

```text
explicit command-line options > configuration file > built-in defaults
```

The same file is applied to automatic daemon startup and to explicit `chrome-devtools start` when an option is omitted. Config keys are validated with the existing `start` option parser (including coerce/conflict rules). `chrome-devtools status` continues to report the effective daemon arguments.

Fixes #2388

## Design notes

Implemented against current `main` rather than resurrecting closed PR #2389.

- Config loading lives in `src/config/cli-config.ts` next to the current option modules.
- Auto-start and `start` share `getCliStartOptions()`, `applyCliStartDefaults()`, and yargs `.config()`, so CLI-specific defaults such as `isolated` stay consistent.
- Auto-start no longer serializes the tool command’s argv into daemon flags. Tool commands do not register `start` options, and overlapping keys (for example `viewport` on `emulate`) could leak into the daemon. Persistent defaults belong in the config file; explicit `chrome-devtools start --…` still overrides them.
- Unknown or invalid config keys fail with the same validation as `chrome-devtools start`.

## Testing

- Unit tests for config path resolution, missing/invalid files, option validation, and CLI > config > built-in precedence (`tests/cli-config.test.ts`)
- E2E: automatic `list_pages` startup uses config values and `status` shows them
- E2E: explicit `start` uses config when flags are omitted
- E2E: command-line flags override config

Locally: `npm run format`, `npx tsc --noEmit`, `npm run test tests/cli-config.test.ts`, and the chrome-devtools start/commands e2e files.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -33,6 +33,27 @@ chrome-devtools take_screenshot 1 --filePath screenshot.png
 chrome-devtools stop
 ```
 
+## Startup defaults
+
+You can persist options used to start the daemon in an XDG-compatible configuration file at `$XDG_CONFIG_HOME/chrome-devtools/config.json`, or `~/.config/chrome-devtools/config.json` when `XDG_CONFIG_HOME` is not set.
+
+The file contains a JSON object whose keys match the options accepted by `chrome-devtools start`. Run `chrome-devtools start --help` for the supported options.
+
+```json
+{
+  "executablePath": "/usr/bin/chromium",
+  "headless": false
+}
+```
+
+These defaults apply both when a tool command automatically starts the daemon and when `chrome-devtools start` is invoked explicitly. Precedence is:
+
+```text
+explicit command-line options > configuration file > built-in defaults
+```
+
+`chrome-devtools status` reports the effective daemon arguments after this merge.
+
 ## Command Usage
 
 The CLI only supports tools available in the MCP server without additional arguments (see [Tool reference](./tool-reference.md)).
--- a/skills/chrome-devtools-cli/SKILL.md
+++ b/skills/chrome-devtools-cli/SKILL.md
@@ -190,3 +190,5 @@ chrome-devtools start --headless=false # Start with visible browser window
 chrome-devtools status  # Checks if chrome-devtools-mcp is running
 chrome-devtools stop    # Stop chrome-devtools-mcp if any
 ```
+
+Persist daemon startup defaults in `$XDG_CONFIG_HOME/chrome-devtools/config.json` (or `~/.config/chrome-devtools/config.json`). Keys match `chrome-devtools start` options. Command-line flags override the file; the file overrides built-in defaults. The same file is used when a tool command auto-starts the daemon.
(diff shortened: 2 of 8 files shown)


OrKoN:
Please see https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2389#issuecomment-5374108217 there is no need to re-publish this PR because that has not changed. Thanks!

PR #2633, opened 2026-09-01
refactor: wrap Server into a class

Lightning00Blade:
I tried to keep the lines of things the same to ease the review process. I will see if any follow up patches will be helpful after this is merged.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,26 +6,28 @@
 
 import type fs from 'node:fs';
 
-import {type ParsedArguments} from './config/mcp-options.js';
 import type {Channel} from './browser.js';
 import {ensureBrowserConnected, ensureBrowserLaunched} from './browser.js';
+import {type ParsedArguments} from './config/mcp-options.js';
 import {loadIssueDescriptions} from './devtools/issueDescriptions.js';
 import {McpContext} from './McpContext.js';
 import {ClearcutLogger} from './telemetry/ClearcutLogger.js';
 import {FilePersistence} from './telemetry/persistence.js';
 import {
-  McpServer,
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -42,174 +44,214 @@ puppeteer.setFollowSymlinks(false);
...
+
+  /**
+   * Roots are client state rather than browser state, so the last listing stays
+   * valid across browser reconnects and only the client can invalidate it, via
+   * the `roots/list_changed` notification handled below
+   */
+  #lastRoots?: Root[];
+  #toolMutex = new Mutex();
+
+  private constructor(
```
OrKoN, line 65: should we have close/dispose method on the new class?

src/index.ts
```diff
@@ -42,174 +44,214 @@ puppeteer.setFollowSymlinks(false);
  */
 const ROOTS_REQUEST_TIMEOUT = 5_000;
 
-export async function createMcpServer(
```
OrKoN, line 45: this was exported on purpose as gemini-cli planned to consume it (and we do not know who else might). Can we create this public API?

src/index.ts
```diff
@@ -42,174 +44,214 @@ puppeteer.setFollowSymlinks(false);
...
+    }
+    await loadIssueDescriptions();
+  }
+
+  /**
+   * `timeout` is only passed where a tool call is waiting on the result – the
+   * background refreshes below block nobody, so bounding them would just discard
+   * roots a slow client was about to send
+   */
+  #updateRoots = async (timeout?: number): Promise<void> => {
```
OrKoN: can this be a regular method?

OrKoN (approved):

Issue #261, opened 2025-10-02
Headless isolated launch fails as root without --no-sandbox


mamoru127:
## Summary
  Running `chrome-devtools-mcp` as the root user causes Chrome to exit immediately with `Running as root without --no-sandbox is not supported`, so the MCP client receives `Target closed / Connection closed`
  errors.

  ## Environment
  - chrome-devtools-mcp 0.6.0
  - Node.js 22.20.0
  - OS: Linux container (root user)
  - Command: `node build/src/index.js --headless --isolated`

  ## Steps to Reproduce
  1. Checkout v0.6.0
  2. Run `npm install && npm run build`
  3. Execute `node build/src/index.js --headless --isolated` as root
  4. Observe Chrome crashing with the log above and the MCP server exiting

  ## Expected Behavior
  The bundled Chrome should launch successfully even when the server runs as root (typical for containers/CI), without requiring extra flags from the user.

  ## Suggested Fix
  Automatically add sandbox flags when running as root. For example in `src/browser.ts`:

  ```ts
  const args = [
    ...(options.args ?? []),
    '--hide-crash-restore-bubble',
  ];

  if (process.getuid?.() === 0 && !args.some(arg => arg.startsWith('--no-sandbox'))) {
    args.push('--no-sandbox', '--disable-setuid-sandbox');
  }

  Then rebuild (npm run build) so the generated build/src/browser.js picks up the same logic.

wsxiaoys:
Encountering the same issue, can confirm the proposed solution works. it'll be nice to just allow passing additional args to chrome as launch args. e.g `--launch-arg="--no-sandbox --disable-setuid-sandbox"'

OrKoN:
Generally we cannot recommend running without Chrome sandboxes or turn them off automatically but we can allow passing additional args to Chrome. Started a PR here https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/261

PR #2634, opened 2026-09-01
fix: explain launch failures caused by running as root

elcodabra:
Follow-up to #261. That issue was closed by #338, which added `--chrome-arg` so the sandbox can be disabled explicitly. The flag works, but the failure it fixes is still undiscoverable — you cannot find out that `--chrome-arg=--no-sandbox` is the answer.

### Problem

`launch()` uses `pipe: true`. In pipe mode Puppeteer never calls `waitForLineOutput()`, which is the only place Chrome's stderr makes it into the thrown error. So when Chrome exits at startup with

```
Running as root without --no-sandbox is not supported. See https://crbug.com/638180.
```

that line is dropped, and what reaches the MCP client is:

```
Protocol error (Target.setDiscoverTargets): Target closed
```

Reproduced on `main` (v1.8.0) by pointing `executablePath` at a stub that prints Chrome's message to stderr and exits 1 — exactly the symptom reported in #261. The `catch` in `launch()` only recognised `The browser is already running`.

### Fix

Detect the case and rethrow with an explanation. Deliberately **not** disabling the sandbox automatically, per https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/261#issuecomment-3389871157:

```
Chrome failed to start: Protocol error (Target.setDiscoverTargets): Target closed

chrome-devtools-mcp is running as root and Chrome does not start as root unless its
sandbox is disabled (https://crbug.com/638180). Prefer running chrome-devtools-mcp as
a non-root user. If that is not possible, for example in a container, pass
--chrome-arg=--no-sandbox. That disables the Chrome sandbox, so only do it for content
you trust.
```

Notes on the shape of the check:

- It runs **only on a failed launch**, so root with a properly installed setuid sandbox is unaffected.
- The original error is kept in the message and as `cause`, so an unrelated failure under root is never masked or mislabelled.
- `--no-sandbox` present in the args short-circuits it: root is then not what stopped Chrome.
- `--disable-setuid-sandbox` and `--no-sandbox-and-elevated` do **not** count as opting out — Chrome's zygote check requires `--no-sandbox` specifically. Covered by a test.
- On Windows `process.getuid` is undefined, so the check is a no-op.

Also adds a "Running as root" section to `docs/troubleshooting.md`, next to the existing "Operating system sandboxes".

### Testing

Five unit tests in `tests/browser.test.ts` covering root / non-root / no-uid platforms, the already-opted-out args, and the near-miss args above.

The pre-existing integration tests in that file cannot run in my environment — Chrome reports `No usable sandbox! ... AppArmor userns restrictions` there, on a clean checkout as well as with this change.


elcodabra:
@googlebot I signed it!

docs/troubleshooting.md
```diff
@@ -81,6 +81,22 @@ either disable sandboxing for `chrome-devtools-mcp` in your MCP client or use
...
 of the MCP client sandbox.
 
+### Running as root
+
+Chrome does not start as root unless its own sandbox is disabled, which is a
+common issue in containers and CI images that run everything as root. Chrome
+exits immediately and `chrome-devtools-mcp` reports that Chrome failed to start.
+
+The recommended fix is to run `chrome-devtools-mcp` as a non-root user. If that
+is not possible, disable the Chrome sandbox explicitly:
```
OrKoN, line 91: Thanks for the PR. Running as non-root should always be possible (also in containers), is there scenario where it is not? I am a bit hesitant to recommend --no-sandbox as it is likely that users could try to use it for unrelated issues resulting in decreased security. Can we perhaps link to https://pptr.dev/troubleshooting#setting-up-chrome-linux-sandbox instead of giving specific examples of running with --no-sandbox?

elcodabra: Good point — dropped the `--no-sandbox` recommendation entirely.

On the scenario question: I don't have one where non-root is genuinely impossible. What I hit (and what #261 describes) is images that just happen to run everything as root, and there the fix is to add a user, not to disable the sandbox. So both the runtime error and the docs section now say to run as a non-root user and link https://pptr.dev/troubleshooting#setting-up-chrome-linux-sandbox for the sandbox setup; `--no-sandbox` is no longer mentioned in either.

The detection itself is unchanged: it still only runs on a failed launch, keeps the original error as the message and `cause`, and short-circuits when `--chrome-arg=--no-sandbox` was passed (root then isn't what stopped Chrome). The value it adds is turning `Protocol error (Target.setDiscoverTargets): Target closed` into something that names root as the cause.

OrKoN (approved):

PR #2637, opened 2026-09-01
refactor: move transform logic to Clearcut level

Lightning00Blade:
I think this makes a little bit cleaner structures. As now Clearcut holds the responsibility of how the data is formatted.


yulunz (approved):
thanks for the cleanup!

PR #2638, opened 2026-09-02
feat: extend --no-javascript-evaluation to cover navigations and initScripts

OrKoN:


src/tools/pages.ts
```diff
@@ -195,8 +202,17 @@ export const navigatePage = definePageTool(() => {
         request.params.type = 'url';
       }
 
+      if (request.params.url) {
+        validateUrl(request.params.url, args?.javascriptEvaluation);
+      }
+
       let initScriptId: string | undefined;
       if (request.params.initScript) {
+        if (args?.javascriptEvaluation === false) {
```
Lightning00Blade: Is our Parser strict or not? If it is we don't need this. And think it is not but we handle it on `this.unknownArgumentNames`

src/utils/url.ts
```diff
@@ -48,3 +48,30 @@ export function isLocalhost(url?: string): boolean {
...
+  let parsed: URL;
+  try {
+    parsed = new URL(url);
+  } catch {
+    throw new Error(
+      `Invalid URL: "${url}". URLs must be valid according to the URL standard.`,
+    );
+  }
+
+  if (javascriptEvaluation === false && parsed.protocol === 'javascript:') {
```
Lightning00Blade: What about `data:` urls?

src/ToolHandler.ts
```diff
@@ -285,6 +285,22 @@ export class ToolHandler {
       };
     }
 
+    if (typeof params.url === 'string') {
```
Lightning00Blade: IS this needed there seems like overreach. for what this class should handle

Lightning00Blade (approved):
LGTM with comments.

src/tools/pages.ts
OrKoN: I think zod is not strict but we handle unknown args by ourselves. Updated.

PR #2644, opened 2026-09-02
refactor: use Zod meta to annotate the inputSchema

Lightning00Blade:


karandeep2799-source (approved):

Lightning00Blade:
@OrKoN Before I invest time in rebasing and fixing issue, do you think this is something we will want and pursue?

OrKoN:
> @OrKoN Before I invest time in rebasing and fixing issue, do you think this is something we will want and pursue?

let's keep the current approach at least it makes it visible on each tool definition that file access needs to be considered at build time. I think with meta we would only be able to lint this or hack zod types to require meta? I would rather leave the current approach as is and perhaps see if we can move the logic from tool definitions/handlers into where the file access happens.

Lightning00Blade:
With Zod v4 Meta has first class support, so we can drop the type and hacks. https://zod.dev/metadata

I am not sure what you mean by `I think with meta we would only be able to lint this` we don't verify anything at this point so people can forget to add it and it's still valid.

With the Meta an EsLint rule would be much simpler, as it can look at the Callstack of the Zod Schema instead of getting split information.

OrKoN:
> I am not sure what you mean by I think with meta we would only be able to lint this we don't verify anything at this point so people can forget to add it and it's still valid.

we currently should have a compile error if verifySchemaFiles is not explicitly defined.

Lightning00Blade:
Closing per the comment above.

PR #2647, opened 2026-09-03
fix: treat timeout 0 as the default wait, not unlimited

gyanu2507:
## Summary
`timeoutSchema` documents that `timeout: 0` uses the default wait. The transform used `value && value <= 0`, which is falsy for `0`, so the number `0` was passed through to Puppeteer. In Puppeteer, `timeout: 0` means wait forever.

Agents that pass `0` expecting the normal default can hang `navigate_page` / `new_page` on a stuck load.

## Test plan
- [x] `timeout: 0` and `-1` parse to `undefined`
- [x] `timeout: 500` stays `500`
- [x] omitted timeout stays omitted

### DIFF
--- a/tests/tools/timeoutSchema.test.ts
+++ b/tests/tools/timeoutSchema.test.ts
@@ -1,0 +1,31 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'node:assert';
+import {describe, it} from 'node:test';
+
+import {zod} from '../../src/third_party/index.js';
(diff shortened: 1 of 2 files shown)


tests/tools/timeoutSchema.test.ts
```diff
@@ -0,0 +1,31 @@
+/**
```
wolfib, line 1: Existing test files are named after the file they are testing. Could you please rename the file to `ToolDefinition.test.ts`?

gyanu2507:
Main already treats `timeout <= 0` as the default in `timeoutSchema`, which is the case this branch was fixing. The branch now conflicts with that change, so I am closing it.

PR #2648, opened 2026-09-03
fix: filter out Chrome webui targets by default

OrKoN:


src/utils/url.ts
```diff
@@ -49,17 +49,111 @@ export function isLocalhost(url?: string): boolean {
...
+      !options.categoryExtensions &&
+      (lower.startsWith('chrome-extension://') ||
+        lower.startsWith('chrome-extension:'))
+    ) {
+      return false;
+    }
+    return true;
+  }
+
+  if (protocol === 'chrome:' || protocol === 'chrome-untrusted:') {
```
Lightning00Blade: This case should be part of the `new URL` as else it's empty.

src/utils/url.ts
```diff
@@ -49,17 +49,111 @@ export function isLocalhost(url?: string): boolean {
...
+ * @param url The URL string to test.
+ * @param options Configuration specifying whether extensions are enabled.
+ * @returns true if the URL is allowed.
+ */
+export function isAllowedUrl(
+  url: string,
+  options: IsAllowedUrlOptions,
+): boolean {
+  if (typeof url !== 'string') {
+    return false;
```
Lightning00Blade, line 81: This if should not be reachable/

src/utils/url.ts
```diff
@@ -49,17 +49,111 @@ export function isLocalhost(url?: string): boolean {
...
+        (host === '' &&
+          (path === 'newtab' ||
+            path === 'new-tab-page' ||
+            path.startsWith('inspect')))
+      ) {
+        return true;
+      }
+    }
+  } catch {
+    // If not a parseable URL, fallback to checking scheme prefixes.
```
Lightning00Blade: Can't we just return `false` in this case? I also don't see any test case for this.

src/utils/url.ts
```diff
@@ -49,17 +49,111 @@ export function isLocalhost(url?: string): boolean {
...
+  const lower = url.toLowerCase();
+  let protocol: string;
+  try {
+    const parsed = new URL(url);
+    protocol = parsed.protocol;
+    if (protocol === 'chrome:') {
+      const host = parsed.hostname.toLowerCase();
+      const path = parsed.pathname.toLowerCase();
+      if (
+        host === 'newtab' ||
```
Lightning00Blade: Can we split it up into two if blocks.

src/utils/url.ts
```diff
@@ -49,17 +49,111 @@ export function isLocalhost(url?: string): boolean {
...
+    ) {
+      return false;
+    }
+    return true;
+  }
+
+  if (protocol === 'chrome:' || protocol === 'chrome-untrusted:') {
+    return false;
+  }
+  if (!options.categoryExtensions && protocol === 'chrome-extension:') {
```
Lightning00Blade: Same as above

src/McpContext.ts
```diff
@@ -164,6 +167,15 @@ export class McpContext implements Context {
 
   #onTargetCreated = async (target: Target) => {
     try {
+      const url = target.url();
+      if (
+        url &&
+        !isAllowedUrl(url, {
+          categoryExtensions: this.#options.categoryExtensions,
+        })
```
Lightning00Blade, line 174:
```suggestion
      if (
        !isAllowedUrl(url, {
          categoryExtensions: this.#options.categoryExtensions,
        })
```

Lightning00Blade (approved):

PR #2649, opened 2026-09-03
fix: wait for scheduled navigations in slim navigate

xia-chao:
## Problem

`navigate_page` and `new_page` wait for a follow-up navigation (meta refresh, scheduled `location` change) through `waitForEventsAfterAction`. Slim `navigate` only called `page.goto` and returned as soon as the first document loaded.

On a URL that immediately meta-refreshes to another path, slim reported the interstitial URL. Agents using `--slim` then evaluated or screenshotted the wrong page.

## Fix

Wrap slim `navigate`'s `goto` in the same `waitForEventsAfterAction` helper. The existing beforeunload dialog handling and 30s timeout are unchanged. The "Navigated to …" line now uses the settled `page.url()`.

## Testing

- `tests/tools/slim/tools.test.ts`: meta refresh from `/meta` to `/final`; asserts the tool returns after `#done` is present and the reported URL ends with `/final`.
- Existing slim navigate / evaluate / screenshot cases still pass.

### DIFF
--- a/src/tools/slim/tools.ts
+++ b/src/tools/slim/tools.ts
@@ -69,7 +69,12 @@ export const navigate = definePageTool(args => {
       page.pptrPage.on('dialog', dialogHandler);
 
       try {
-        await page.pptrPage.goto(request.params.url, options);
+        await page.waitForEventsAfterAction(
+          async () => {
+            await page.pptrPage.goto(request.params.url, options);
+          },
+          {timeout: options.timeout},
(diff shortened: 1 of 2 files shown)


OrKoN:
Thanks for the PR but it is the intention to keep the slim mode minimal and fast. If you use the slim mode, I'd suggest adding waiting on the client side or switching to the full mode.

PR #2659, opened 2026-09-04
refactor: unify call type and ensure typesafety

Lightning00Blade:
Prefactoring for the SDK v2.


src/tools/ToolDefinition.ts
```diff
@@ -386,85 +401,66 @@ export type ContextPage = Readonly<{
...
+let defaultToolArgs: ParsedArguments | undefined;
+
+export function resolveToolArgs(args?: ParsedArguments): ParsedArguments {
+  if (args !== undefined) {
+    return args;
   }
-  return definition;
+  defaultToolArgs ??= parseArguments('0.0.0', DEFAULT_TOOL_ARGV, {
+    ...process.env,
+    CHROME_DEVTOOLS_MCP_NO_USAGE_STATISTICS: 'true',
```
OrKoN: the default for this is false? do we need to use parseArguments on this level?

src/tools/performance.ts
```diff
@@ -6,6 +6,7 @@
 
 import zlib from 'node:zlib';
 
+import type {ParsedArguments} from '../config/mcp-options.js';
```
OrKoN: let's create a dedicated interface for args in ToolDefinition.ts duplicating args that tools actually consume and avoid depending on the MCP server config directly. That would also help to keep it more manageable. WDYT?

Lightning00Blade: I think it may makes sense but for now I would prefer keeping it in one place. We can revisit after all the changes land.

OrKoN (approved):

PR #2661, opened 2026-09-04
feat: support config file

Lightning00Blade:
Utilizes the Yargs default. We may need to change this to resolve this ourself, but for now we can explicitly allow this.

### DIFF
--- a/src/config/mcp-options.ts
+++ b/src/config/mcp-options.ts
@@ -440,7 +440,26 @@ export function parser(
   return yargsInstance
     .config('config', 'Path to JSON configuration file', configPath => {
       try {
-        return JSON.parse(readFileSync(configPath, 'utf-8'));
+        const parsed = JSON.parse(readFileSync(configPath, 'utf-8'));
+        if (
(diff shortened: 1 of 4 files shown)


src/config/mcp-options.ts
```diff
@@ -546,6 +551,13 @@ export function parser(
     ]);
 
   return yargsInstance
+    .config('config', 'Path to JSON configuration file', configPath => {
```
OrKoN, line 554: does it understand all aliases and variants of CLI arguments? i.e., can the config file contain `{"no-category-network": true}`? I wonder if we want to have so many options for config files if that is the case.

Lightning00Blade: Added test to show case the current allowed. 
What we can do is re-run the parser on the JSON schema before we return in strict mode to only allow the capitalCase. That will also mean anything that is not on the config we throw an error. WDYT?

Lightning00Blade: Update to remove dashed properties as well.

OrKoN (approved):

PR #2667, opened 2026-09-04
test: skip symlink cases without Windows privileges

### LINKED ISSUE #2666: tests: symlink cases fail on Windows without symlink privileges
### Description of the bug

On Windows, the repository's test suite fails in `tests/utils/files.test.ts` when the current account cannot create symbolic links. Four symlink-specific cases throw `EPERM` before they exercise `resolveCanonicalPath`, while the non-symlink cases pass.

This is a test portability problem rather than a failure of the production path-resolution logic. A similar symlink-only block in `tests/McpContext.test.ts` already avoids running where symlinks are unavailable.

### Reproduction

1. Use Windows with Developer Mode disabled and without `SeCreateSymbolicLinkPrivilege`.
2. Check out `main` at `086299a69e6d322df43d7e54417fce25b3a2fc08`.
3. Run `npm ci`.
4. Run `npm test -- tests/utils/files.test.ts`.

The four tests that call `fs.symlink` fail with `EPERM: operation not permitted, symlink ...`. The three standard-path cases pass.

### Expectation

The symlink-specific tests should run when the host can create symlinks and be reported as skipped when Windows denies that capability. Other errors should still fail, and all non-symlink coverage should continue to run.

### MCP configuration

Not applicable; this is a repository test-suite failure.

1.8.0 (`main` at `086299a69e6d322df43d7e54417fce25b3a2fc08`)

### Chrome version

Not applicable.

### Coding agent version

Not applicable.

### Model version

Not applicable.

### Chat log

Not applicable.

### Node version

v24.14.1

### Operating system

Windows

### Extra checklist

dajiaohuang:
Fixes #2666.

This keeps the symlink-specific `resolveCanonicalPath` coverage on hosts that can create symbolic links. When Windows rejects `fs.symlink` with `EPERM` or `EACCES`, only that case is marked skipped; non-symlink cases and unexpected errors still run or fail normally.

Tests:

- `npm test -- tests/utils/files.test.ts` (3 passed, 4 skipped on Windows without symlink privileges)
- `npm test -- tests/McpContext.test.ts`
- `npm test`
- `npm run check-format`
- `git diff --check`


tests/utils/files.test.ts
```diff
@@ -60,14 +83,16 @@ describe('resolveCanonicalPath', () => {
...
     const targetDir = path.join(tmpDir, 'target');
     await fs.mkdir(targetDir);
     const targetFile = path.join(targetDir, 'file.txt');
     await fs.writeFile(targetFile, 'hello');
 
     const symlinkDir = path.join(tmpDir, 'symlink_dir');
-    await fs.symlink(targetDir, symlinkDir, 'dir');
+    if (!(await createSymlinkOrSkip(t, targetDir, symlinkDir, 'dir'))) {
+      return;
+    }
```
nattallius: I would make it more readable 

```suggestion
const created = await createSymlinkOrSkip(t, targetDir, symlinkDir, 'dir');
if (!created) {
  return;
}
```

nattallius (commented):
Hi, thank you for your contribution. Can you please sign the CLA?

yangfan-yf-yf (commented):
I checked the Windows no-symlink-privilege case by applying the test-file diff from 561e8777a4de93afbcccd5238f8c0d3c73b73ee4 to my existing d23d057b7144d56526fdbe6901d2134902b583f8 checkout. This is a compatibility check on that checkout, not a full-suite run of this PR's head.

Without this patch, the four symlink cases failed with EPERM in my full-suite run. With the patch, `npm run test -- tests/utils/files.test.ts` reports 3 passed, 0 failed, 4 skipped, and each skip gives the Windows privilege reason. The three ordinary-path cases still execute and pass. I restored the original source and rebuilt after the experiment.

The catch is limited to Windows EPERM/EACCES and rethrows other errors by inspection. I did not test an elevated/developer-mode Windows session or another operating system, so this result only confirms the no-privilege path; the four symlink assertions were skipped, not validated. I did not change Windows privileges or developer-mode settings.

tests/utils/files.test.ts
dajiaohuang: Applied in d97b8fc: each helper result is now assigned to `created` before the early return. `npm run test tests/utils/files.test.ts` and `npm run check-format` pass.

dajiaohuang:
Branch updated to `main` (`cdc365c`) in 46455e7; the test-file diff is unchanged.

Closing the OS gap noted above:

- Windows without symlink privileges (node v24.14.1): 3 passed, 4 skipped with the Windows privilege reason, 0 failed. Before the patch, the same run failed those four cases with `EPERM`.
- Linux (WSL Ubuntu 24.04, node v26.4.0): 7 passed, 4 symlink cases included, 0 skipped — the assertions still run where symlinks are available.
- `npm run check-format` passes.

I did not test an elevated or Developer-Mode Windows session.

nattallius (approved):

PR #2668, opened 2026-09-05
fix: accept multiple file paths for upload_file in the CLI

### LINKED ISSUE #2741: upload_file always fails from the CLI: Expected array, received string at filePaths
## Problem

`chrome-devtools upload_file <pageId> <uid> <path>` always fails, no matter how the paths are passed:

```
MCP error -32602: Input validation error: Invalid arguments for tool upload_file: Expected array, received string at filePaths
```

Every form fails — a single path, several paths, `--filePaths=x`, and repeated `--filePaths` flags. Reproduced on v1.8.0 and on current `main`.

## Cause

In `src/bin/chrome-devtools.ts` the yargs command string appends each optional arg as a `[--flag]` token after the required positionals:

```
upload_file <pageId> <uid> <filePaths> [--includeSnapshot]
```

yargs parses each `[--flag]` token as another *optional positional*. Since a variadic positional has to be the last one, `filePaths` can never be variadic, so yargs always hands the daemon a plain string and the tool's `z.array(z.string())` schema rejects it. Setting `type: 'array'` on the positional does not change this.

`upload_file` is currently the only tool with a required array argument, so it is the only command affected.

## Expected

```
$ chrome-devtools upload_file 2 1_1 /tmp/a.txt /tmp/b.txt
File uploaded from /tmp/a.txt, /tmp/b.txt.
```

## Fix

PR #2668 marks array-typed required args as variadic (`<filePaths..>`) and moves the `[--flag]` tokens into the command's `usage()` line, which is all they were for. `--help` output is unchanged apart from the `..` marker on `upload_file`.

u9g:
Fixes #2741

## Problem

`chrome-devtools upload_file <pageId> <uid> <path>` always fails:

```
MCP error -32602: Input validation error: Invalid arguments for tool upload_file: Expected array, received string at filePaths
```

Every form fails (one path, several paths, `--filePaths=x`, repeated flags). Reproduced on v1.8.0 and current `main`.

## Cause

The generated yargs command string appends the optional flags as `[--flag]` tokens after the required positionals, e.g. `upload_file <pageId> <uid> <filePaths> [--includeSnapshot]`. yargs parses each `[--flag]` token as another optional positional. A variadic positional has to be the last one, so `filePaths` can never be variadic and yargs always hands the daemon a plain string, which the tool's `z.array(z.string())` schema rejects. Setting `type: 'array'` on the positional does not change this.

## Fix

- Mark array-typed required args as variadic in the command string (`<filePaths..>`).
- Move the `[--flag]` tokens into the command's `usage()` line, which is all they were for. `--help` output for every command is unchanged except for the `..` marker on `upload_file`.

## Verification

Built CLI against a headless daemon on a `data:` page with `<input type=file multiple>`:

```
$ chrome-devtools upload_file 2 1_1 /tmp/a.txt /tmp/a.txt
File uploaded from /private/tmp/a.txt, /private/tmp/a.txt.
$ chrome-devtools evaluate_script --pageId 2 "() => Array.from(document.getElementById('f').files).map(f=>f.name)"
["a.txt","a.txt"]
```

`upload_file` is the only tool with a required array argument, so no other command's parsing changes. `tests/cli.test.ts` and `tests/daemon/client.test.ts` pass.

The command-string construction moves into `src/config/cli-commands.ts` so `tests/cli.test.ts` can cover it. Alongside the `upload_file` cases there is a check over every generated command asserting no command string contains a `--` token and that any required array arg is rendered variadic and last — all four fail against the previous behaviour.


u9g:
I signed it

drotavio-web:
- u9g propôs uma correção na CLI para aceitar múltiplos caminhos de
   arquivo no upload_file.
   - Google-cla[bot] solicitou o Acordo de Licença de Contribuidor, que u9g
   confirmou ter assinado.



Otávio P

Em sáb., 5 de set. de 2026 às 18:16, u9g ***@***.***>
escreveu:

src/config/cli-commands.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Optional args are rendered as `[--flag]` in the usage line only: as part of
+ * the command string yargs parses each one as a trailing positional, and a
+ * variadic positional has to be the last one.
+ */
+export function buildCommand(
+  commandName: string,
+  args: Record<string, ArgDef>,
+): {command: string; usage: string} {
+  let command = commandName;
+  for (const [name, arg] of Object.entries(args)) {
```
samiyac, line 21: Thank you for sending the PR.

Instead of iterating twice - we can build `command` and `usage` strings in a single loop?

samiyac (changes requested):

src/config/cli-commands.ts
u9g: done in daa36df

samiyac: Thanks! Can you please squash and rebase?

u9g: Sure, done

samiyac (approved):

PR #2673, opened 2026-09-06
Add initial devcontainer configuration

vdiassantiago5-star:
Ok


nattallius:
We are going to pass on this change. A generic universal:2 devcontainer configuration without project-specific setup (such as targeting our required Node/TypeScript environment, workspace settings, or lifecycle scripts like npm install) adds unnecessary maintenance overhead without providing significant value over a standard GitHub Codespace/container defaults.

If there's a demonstrated need for devcontainer support in the repository, we'd prefer discussing requirements in an issue first to ensure it's tailored and properly maintained. Closing for now.

PR #2679, opened 2026-09-07
feat: report pages opened by input actions

### LINKED ISSUE #367: return all pages when click web element

lisa0314:
Fixes #367

When an input action opens one or more pages, capture the source page's Puppeteer `popup` events and report only those newly opened pages in the action response.

Each new page is listed under `## New pages` with its MCP page ID, title, and URL. The same entries are exposed through `structuredContent.newPages`. Pages that were already open are not included, and the complete page list is not enabled implicitly.

Tests cover a single popup with an unrelated pre-existing page, multiple popups in event order with distinct MCP page IDs, and popup listener cleanup.

## Testing

- `npm run build`
- `npm run check-format`
- `npm run test -- tests/utils/WaitForHelper.test.ts tests/tools/input.test.ts`
- `npm run test -- tests/tools/extensions.test.ts tests/tools/pages.test.ts`

### DIFF
--- a/src/McpResponse.ts
+++ b/src/McpResponse.ts
@@ -675,6 +675,10 @@ export class McpResponse implements Response {
     content: Array<TextContent | ImageContent>;
     structuredContent: object;
   }> {
+    if (this.#attachedWaitForResult?.newPages?.length && !this.#includePages) {
+      await context.createPagesSnapshot();
+    }
+
     const [
       snapshot,
       detailedNetworkRequest,
@@ -784,6 +788,7 @@ export class McpResponse implements Response {
         defaultValue?: string;
       };
       pages?: object[];
+      newPages?: object[];
       pagination?: object;
       heapSnapshot?: {
         stats?: object;
@@ -855,6 +860,29 @@ export class McpResponse implements Response {
         );
         structuredContent.navigatedToUrl =
           this.#attachedWaitForResult.navigatedToUrl;
+      }
+      if (this.#attachedWaitForResult.newPages?.length) {
+        const mcpPagesByPptrPage = new Map(
+          context.getPages().map(mcpPage => [mcpPage.pptrPage, mcpPage]),
+        );
+        const newPages = this.#attachedWaitForResult.newPages.flatMap(page => {
+          const mcpPage = mcpPagesByPptrPage.get(page);
+          return mcpPage ? [mcpPage] : [];
+        });
+
+        if (newPages.length) {
+          response.push('## New pages');
+          const structuredPages = [];
+          for (const mcpPage of newPages) {
+            const title = await fetchPageTitle(mcpPage.pptrPage);
+            const pageLabel = title
+              ? `${truncateTitle(title)} (${mcpPage.pptrPage.url()})`
+              : mcpPage.pptrPage.url();
+            response.push(`${mcpPage.id}: ${pageLabel}`);
+            structuredPages.push(createStructuredPage(mcpPage, context, title));
+          }
+          structuredContent.newPages = structuredPages;
+        }
       }
     }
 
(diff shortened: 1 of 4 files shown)


lisa0314:
PTAL. Thanks!

src/utils/WaitForHelper.ts
```diff
@@ -32,6 +34,7 @@ export class WaitForHelper {
     this.#stableDomFor = 100 * cpuTimeoutMultiplier;
     this.#expectNavigationIn = 100 * cpuTimeoutMultiplier;
     this.#navigationTimeout = 3000 * networkTimeoutMultiplier;
+    this.#sourcePage = page;
```
Lightning00Blade, line 37: This should not be needed we have `#page` CdpPage extends Page, so all methods should be there.

src/utils/WaitForHelper.ts
```diff
@@ -295,6 +309,8 @@ export interface WaitForEventsResult {
    * Whether a dialog was automatically handled during the action.
    */
   dialogHandled?: boolean;
+  /** Pages opened by the action. */
+  newPages?: Page[];
```
Lightning00Blade, line 313: Let's call it `openedPages` as new can imply that this is just a diff and is unrelated to the action.
Let's also change this to a WeakSet.

src/McpResponse.ts
```diff
@@ -856,6 +861,29 @@ export class McpResponse implements Response {
         structuredContent.navigatedToUrl =
           this.#attachedWaitForResult.navigatedToUrl;
       }
+      if (this.#attachedWaitForResult.newPages?.length) {
+        const mcpPagesByPptrPage = new Map(
```
Lightning00Blade, line 865: This should be replaced with 
```ts
const openedMcpPages = context.getPages().filter(mcpPage => this.#attachedWaitForResult.openPages.has(mcpPage.pptrPage));
```

src/McpResponse.ts
```diff
@@ -856,6 +861,29 @@ export class McpResponse implements Response {
...
+        const newPages = this.#attachedWaitForResult.newPages.flatMap(page => {
+          const mcpPage = mcpPagesByPptrPage.get(page);
+          return mcpPage ? [mcpPage] : [];
+        });
+
+        if (newPages.length) {
+          response.push('## New pages');
+          const structuredPages = [];
+          for (const mcpPage of newPages) {
+            const title = await fetchPageTitle(mcpPage.pptrPage);
```
Lightning00Blade, line 877: I see that the logic here is repeated from other places, let's extract it into a helper.

tests/tools/input.test.ts
```diff
@@ -89,6 +89,105 @@ describe('input', () => {
...
+
+        const newPages = getStructuredNewPages(result.structuredContent);
+        assert.deepStrictEqual(
+          newPages.map(({url, title}) => ({url, title})),
+          [{url: popupUrl, title: 'Opened by click'}],
+        );
+      });
+    });
+
+    it('reports multiple pages opened by one click in event order', async () => {
```
Lightning00Blade, line 140: I am not sure if we can guarantee order, and there is not a clear use case from my understanding.

src/McpResponse.ts
```diff
@@ -675,6 +675,10 @@ export class McpResponse implements Response {
     content: Array<TextContent | ImageContent>;
     structuredContent: object;
   }> {
+    if (this.#attachedWaitForResult?.newPages?.length && !this.#includePages) {
+      await context.createPagesSnapshot();
```
Lightning00Blade, line 679: I assume we need this because we need the latest snapshot to populate the McpPages.

PR #2693, opened 2026-09-08
chore: add comments and open_devtools tools without implementation

OrKoN:
This is part 2 of stacked PRs for DevTools comments support.

Defines the `open_devtools`, `get_devtools_comments`, `resolve_devtools_comment`, and `reveal_in_devtools` tools with complete schemas, categories, and descriptions, throwing `Not implemented` until implemented. Gated behind the hidden flag `devtoolsComments`.

### Stack
1. #2692 chore: add hidden devtoolsComments flag
2. **#this PR**: chore: add comments and open_devtools tools without implementation
3. chore: implement bridge and DevTools comments tools

### DIFF
--- a/src/tools/comments.ts
+++ b/src/tools/comments.ts
@@ -1,0 +1,106 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {zod} from '../third_party/index.js';
+
+import {ToolCategory} from './categories.js';
+import {definePageTool} from './ToolDefinition.js';
+
+export const openDevtools = definePageTool({
+  name: 'open_devtools',
+  description: 'Open a DevTools window for the selected page.',
+  annotations: {
+    category: ToolCategory.DEBUGGING,
+    readOnlyHint: true,
+    conditions: ['devtoolsComments'],
+  },
+  schema: {},
+  blockedByDialog: false,
+  verifyFilesSchema: [],
+  handler: async () => {
+    throw new Error('Not implemented');
+  },
+});
+
+export const getDevtoolsComments = definePageTool({
+  name: 'get_devtools_comments',
+  description: 'Retrieve user comments from the DevTools window for the page.',
+  annotations: {
+    category: ToolCategory.DEBUGGING,
+    readOnlyHint: true,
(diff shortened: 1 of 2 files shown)


src/tools/comments.ts
```diff
@@ -0,0 +1,106 @@
...
+  name: 'reveal_in_devtools',
+  description:
+    'Navigate DevTools to a specified panel and highlight a target DOM node or network request.',
+  annotations: {
+    category: ToolCategory.DEBUGGING,
+    readOnlyHint: true,
+    conditions: ['devtoolsComments'],
+  },
+  schema: {
+    panelName: zod
```
yulunz, line 83: looking at the implementation: third_party/devtools-frontend/src/front_end/models/comment_manager/CD4ABridge.ts

I think the expectation is that either uid or reqid should be provided, both empty is fine, but not both. Maybe we should note this in describe()? Also, the panel name can probably be optional as well.

yulunz (approved):

src/tools/comments.ts
```diff
@@ -0,0 +1,106 @@
...
+  name: 'resolve_devtools_comment',
+  description:
+    'Append an agent reply to a DevTools comment thread and mark it as resolved.',
+  annotations: {
+    category: ToolCategory.DEBUGGING,
+    readOnlyHint: false,
+    conditions: ['devtoolsComments'],
+  },
+  schema: {
+    threadId: zod
```
yulunz, line 54: do we need targetId here too?

yulunz (approved):

src/tools/comments.ts
OrKoN: no, threadIds should be globally unique.

src/tools/comments.ts
OrKoN: Good point, made panelName optional for now and noted that reqid and uid are mutually exclusive.

src/tools/comments.ts
OrKoN: i.e., DevTools UI is already multitarget (in terms of the CDP target under inspection). The tool is page-scoped so it routes to a single DevTools instance.

yulunz: yeah, and that was also clear from the child PR. thanks for the explanation!

PR #2694, opened 2026-09-08
chore: implement bridge and DevTools comments tools

OrKoN:
This is part 3 of stacked PRs for DevTools comments support.

Implements `DevToolsCommentBridge`, `McpPage` and `McpContext` bridge integration, full handlers for comment tools.

### Stack
1. #2692 chore: add hidden devtoolsComments flag
2. #2693 chore: add comments and open_devtools tools without implementation
3. **#this PR**: chore: implement bridge and DevTools comments tools
4. Skill PR

### DIFF
--- a/src/McpPage.ts
+++ b/src/McpPage.ts
@@ -55,6 +55,7 @@ export function replaceHtmlElementsWithUids(schema: JSONSchema7Definition) {
   }
 }
 
+import {DevToolsCommentBridge} from './devtools/DevToolsCommentBridge.js';
 import {
   createTargetUniverse,
   type TargetUniverse,
@@ -139,6 +140,8 @@ export class McpPage implements ContextPage {
   #locatorClass: typeof Locator;
   #navigationTimeout: number;
   #sourceMaps: boolean;
+  #commentBridge?: DevToolsCommentBridge;
+  #onNotification?: (message: string) => void;
 
   constructor(
     page: Page,
@@ -149,12 +152,14 @@ export class McpPage implements ContextPage {
       isolatedContextName?: string;
       navigationTimeout?: number;
       sourceMaps?: boolean;
+      onNotification?: (message: string) => void;
     },
   ) {
     this.#hasNetworkBlockOrAllowlist = options.hasNetworkBlockOrAllowlist;
     this.#locatorClass = options.locatorClass;
     this.#navigationTimeout = options.navigationTimeout ?? NAVIGATION_TIMEOUT;
     this.#sourceMaps = options.sourceMaps ?? true;
+    this.#onNotification = options.onNotification;
     this.pptrPage = page;
     this.id = id;
     this.isolatedContextName = options.isolatedContextName;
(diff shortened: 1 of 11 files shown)


src/McpPage.ts
```diff
@@ -676,6 +711,34 @@ export class McpPage implements ContextPage {
...
 
+  async resolveBackendNodeId(
+    backendNodeId: number,
+  ): Promise<string | undefined> {
+    if (!this.textSnapshot) {
+      this.textSnapshot = await TextSnapshot.create(this);
+    }
+    let id = this.textSnapshot.resolveCdpElementId(backendNodeId);
+    if (!id) {
+      this.textSnapshot = await TextSnapshot.create(this);
```
yulunz, line 722: why do we need to repeat TextSnapshot.create and resolveCdpElementId?

src/tools/comments.ts
```diff
@@ -36,8 +54,62 @@ export const getDevtoolsComments = definePageTool({
...
+      response.appendResponseLine('No open DevTools comments found.');
+      return;
+    }
+
+    response.appendResponseLine(
+      `Found ${threads.length} DevTools comment thread(s):`,
+    );
+    for (const thread of threads) {
+      response.appendResponseLine(`\n### Thread: ${thread.id}`);
+      response.appendResponseLine(`- Comment: ${thread.text}`);
```
yulunz, line 81: so if the comment is not on a DOM node, not on a network request, and not on an editor, the model only sees the text from the user.

Where should we pass the context?

src/tools/comments.ts
```diff
@@ -65,8 +137,39 @@ export const resolveDevtoolsComment = definePageTool({
   },
   blockedByDialog: false,
   verifyFilesSchema: {},
-  handler: async () => {
-    throw new Error('Not implemented');
+  handler: async (request, response) => {
```
yulunz, line 132: one test case that I'd be interested to try on is opening devtools on multiple tabs, and leave comments on all of them, and see if the agent can select the correct page when resolving the comments. It'd also be interesting if these tabs are all loading the same web page.

yulunz (approved):

src/McpPage.ts
OrKoN: The current snapshot might be stale so it would create a new snapshot to just if that is the case.

src/tools/comments.ts
OrKoN: I think we should pass it as part of the text field. WDYT?

src/tools/comments.ts
OrKoN: all these tools are scoped to the inspected page and require a pageId. Usually the agent is good at at picking the right page (we also done some evals for that).

src/tools/comments.ts
yulunz: That makes sense. Maybe "===Context===" and then "===Comment===". Let me create a bug.

OrKoN:
@yulunz PTAL (I have made some changes around backendNodeId resolution and formatting)

yulunz (approved):

PR #2697, opened 2026-09-08
chore: bucketize days_since_last_active.

yulunz:


OrKoN (approved):

OrKoN:
Forgot to mention that we use feat: for user relevant features to be in the changelog. This sounds like it’s better as a chore.

PR #2699, opened 2026-09-08
fix: fail fast on tool calls stuck after a dead browser connection

bitbay:
Root cause: when the debugged browser's CDP transport dies mid-call — silently, without ever firing a `close`/`error`/`disconnected` event (as opposed to a clean disconnect) — the in-flight tool call just hangs forever waiting on a response that will never come. Because `ToolHandler` serializes every tool call behind a single shared `Mutex`, that one stuck call blocks all subsequent tool calls too, effectively taking the whole MCP server down until someone manually reconnects via `/mcp`.

Observed when an Android app being debugged over `chrome-devtools-mcp` was reinstalled/relaunched mid-session, the MCP server's tools went completely unavailable, even though the underlying CDP endpoint (http://127.0.0.1:9222/json) stayed reachable the whole time. This wasn't a startup/connection-time failure, which the existing "next-call self-heal" logic already handles fine — it was a hang mid-call, so there was never a clean "next call" for that self-heal to run on.

- `browser.ts`: export `forgetBrowser()` and attach a `disconnected` listener after every successful connect/launch, so a browser Puppeteer does detect as closed is dropped from the module cache immediately rather than relying solely on the next call's `browser?.connected` check.
- `ToolHandler.ts`: bound each tool handler invocation with a 60s timeout. On timeout, forget the cached browser handle and fail with a clear error instead of hanging until the client's own timeout gives up on the whole server. The next tool call then reconnects via the existing `ensureBrowserConnected`/`#getContext` self-heal path instead of reusing a handle that still looks connected. The forget call is now injectable via an optional constructor parameter (defaulting to the real `forgetBrowser`), so the timeout path can be unit tested without stubbing an ES module.
- `tests/ToolHandler.test.ts`: cover the timeout path directly — a tool handler that never resolves triggers the 60s timeout, `forgetBrowser` is called with the current browser, and a normal (non-timeout) handler rejection does not trigger it.
- `tests/browser.test.ts`: cover the disconnected-listener path against a real launched browser — a clean `disconnect()` is forgotten automatically so the next `ensureBrowserConnected()` reconnects instead of reusing the dead handle, and `forgetBrowser()` only clears the cache on an exact reference match.

Deliberately scoped to fail-fast + rely on the existing next-call self-heal, not automatic retry-and-replay of the timed-out call (a larger change to discuss with whoever verifies this against the real device). A full Android reproduction is out of scope for this change; verified here via the targeted unit tests above.

### DIFF
--- a/src/ToolHandler.ts
+++ b/src/ToolHandler.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import {forgetBrowser} from './browser.js';
 import type {ParsedArguments} from './config/mcp-options.js';
 import type {McpContext} from './McpContext.js';
 import type {McpPage} from './McpPage.js';
@@ -11,7 +12,7 @@ import type {DataFormat} from './McpResponse.js';
 import {McpResponse} from './McpResponse.js';
 import {SlimMcpResponse} from './SlimMcpResponse.js';
 import {ClearcutLogger} from './telemetry/ClearcutLogger.js';
-import type {CallToolResult} from './third_party/index.js';
+import type {Browser, CallToolResult} from './third_party/index.js';
 import {zod} from './third_party/index.js';
 import {labels} from './tools/categories.js';
 import {categoryToFlagName} from './config/category-options.js';
@@ -26,6 +27,21 @@ import {logger} from './utils/logger.js';
 import type {Mutex} from './third_party/index.js';
 import {fileURLToPath, pathToFileURL} from 'node:url';
 import {isLocalhost} from './utils/url.js';
+
+/**
+ * Upper bound on how long a single tool call may wait on the browser
+ * connection. Puppeteer normally rejects in-flight CDP calls when the
+ * underlying transport closes, but a transport that dies silently (e.g. an
+ * adb port-forward torn down mid-call, rather than closed cleanly) never
+ * fires `close`/`error`/`disconnected`, so the call would otherwise hang
+ * until an external (client-side) timeout gives up on the whole server. This
+ * bound turns that into a fast, clear error instead, and forgets the cached
+ * browser handle so the next call reconnects rather than reusing a handle
+ * that still looks connected.
+ */
+export const TOOL_CALL_TIMEOUT_MS = 60_000;
+
+class ToolCallTimeoutError extends Error {}
 
 function buildDisabledMessage(
   toolName: string,
@@ -175,6 +191,8 @@ export class ToolHandler {
     private readonly serverArgs: ParsedArguments,
     private readonly getContext: () => Promise<McpContext>,
     private readonly toolMutex: Mutex,
+    // Injectable for tests; production callers rely on the default.
+    private readonly forgetBrowserOnTimeout: (browser: Browser) => void = forgetBrowser,
   ) {
     const {disabled, reason} = getToolStatusInfo(tool, serverArgs);
     this.disabledReason = reason;
@@ -194,6 +212,38 @@ export class ToolHandler {
     return Object.keys(params).filter(
       key => !Object.hasOwn(this.inputSchema, key),
     );
+  }
+
+  /**
+   * Races a tool handler invocation against TOOL_CALL_TIMEOUT_MS. On timeout,
+   * forgets the cached browser handle (see forgetBrowser()) so the next tool
+   * call re-establishes the connection instead of hanging on the same dead
+   * one. The loser of the race (a genuinely hung handler) is left running;
+   * there is no way to cancel a pending Puppeteer call, but since nothing is
+   * left awaiting it, it cannot block subsequent tool calls.
+   */
+  async #runToolWithTimeout<T>(
+    context: McpContext,
+    handlerPromise: Promise<T>,
+  ): Promise<T> {
+    const timeoutError = new ToolCallTimeoutError(
+      `Tool "${this.tool.name}" timed out after ${TOOL_CALL_TIMEOUT_MS}ms waiting on the browser connection. The connection may have been lost (for example, the debugged browser or app restarted). It will be re-established automatically on the next tool call.`,
+    );
+    let timer: ReturnType<typeof setTimeout>;
+    const timeout = new Promise<never>((_, reject) => {
+      timer = setTimeout(() => reject(timeoutError), TOOL_CALL_TIMEOUT_MS);
+      timer.unref?.();
+    });
+    try {
+      return await Promise.race([handlerPromise, timeout]);
+    } catch (err) {
+      if (err === timeoutError) {
+        this.forgetBrowserOnTimeout(context.browser);
+      }
+      throw err;
+    } finally {
+      clearTimeout(timer!);
+    }
   }
 
   async handle(params: Record<string, unknown>): Promise<CallToolResult> {
@@ -261,21 +311,27 @@ export class ToolHandler {
           if (this.tool.blockedByDialog) {
             page.throwIfDialogOpen();
           }
-          await this.tool.handler(
-            {
-              params,
-              page,
-            },
-            response,
+          await this.#runToolWithTimeout(
             context,
+            this.tool.handler(
+              {
+                params,
+                page,
+              },
+              response,
+              context,
+            ),
           );
         } else {
-          await this.tool.handler(
-            {
-              params,
-            },
-            response,
+          await this.#runToolWithTimeout(
             context,
+            this.tool.handler(
+              {
+                params,
+              },
+              response,
+              context,
+            ),
           );
         }
       } catch (err) {
--- a/tests/ToolHandler.test.ts
+++ b/tests/ToolHandler.test.ts
@@ -17,7 +17,7 @@ import {McpContext} from '../src/McpContext.js';
 import {McpPage} from '../src/McpPage.js';
 import {ClearcutLogger} from '../src/telemetry/ClearcutLogger.js';
 import {zod} from '../src/third_party/index.js';
-import {ToolHandler} from '../src/ToolHandler.js';
+import {TOOL_CALL_TIMEOUT_MS, ToolHandler} from '../src/ToolHandler.js';
 import {ToolCategory} from '../src/tools/categories.js';
 import type {
   DefinedPageTool,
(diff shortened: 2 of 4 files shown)


src/ToolHandler.ts
```diff
@@ -27,6 +28,21 @@ import type {Mutex} from './third_party/index.js';
...
+ * connection. Puppeteer normally rejects in-flight CDP calls when the
+ * underlying transport closes, but a transport that dies silently (e.g. an
+ * adb port-forward torn down mid-call, rather than closed cleanly) never
+ * fires `close`/`error`/`disconnected`, so the call would otherwise hang
+ * until an external (client-side) timeout gives up on the whole server. This
+ * bound turns that into a fast, clear error instead, and forgets the cached
+ * browser handle so the next call reconnects rather than reusing a handle
+ * that still looks connected.
+ */
+export const TOOL_CALL_TIMEOUT_MS = 60_000;
```
samiyac, line 41: I think we should not set it statically - since the user can define a higher timeout for some of the tool calls using `request.params.timeout`

src/ToolHandler.ts
```diff
@@ -196,6 +214,38 @@ export class ToolHandler {
...
+    });
+    try {
+      return await Promise.race([handlerPromise, timeout]);
+    } catch (err) {
+      if (err === timeoutError) {
+        this.forgetBrowserOnTimeout(context.browser);
+      }
+      throw err;
+    } finally {
+      clearTimeout(timer!);
```
samiyac: Style issue (as per `AGENTS.md`): Do not use `!` operator for type assertion.

tests/ToolHandler.test.ts
```diff
@@ -1117,4 +1117,105 @@ describe('ToolHandler', () => {
...
+
+      assert.strictEqual(result.isError, true);
+      assert.match(
+        result.content[0].type === 'text' ? result.content[0].text : '',
+        /timed out/,
+      );
+      assert.strictEqual(
+        forgetBrowserSpy.calledOnceWith(mockContext.browser),
+        true,
+      );
```
samiyac, line 1281: Here and elsewhere

Issue (as per `AGENTS.md`): Use `sinon.assert` methods

samiyac (commented):
Thank you for the PR. I have added some initial review comments.

src/ToolHandler.ts
OrKoN: I think `request.params.timeout` is not part of MCP.

OrKoN: could we make this depend on actual MCP client timeout?

OrKoN (changes requested):
### Critical Bugs & Architectural Issues

#### 1. Timeout Only Wraps `tool.handler()`, Leaving `response.handle()` and `getContext()` Unbounded Inside `toolMutex`
* **Location**: `src/ToolHandler.ts:238-297`
* **Problem**:
  `#runToolWithTimeout()` only wraps `this.tool.handler(...)`. However, inside `this.toolMutex.acquire()`, three major phases talk to CDP:
  1. **Tools whose CDP calls happen inside `response.handle()` will still hang forever**:
     Many tools do not make CDP calls inside `tool.handler()`; instead, their handler synchronously sets a flag on `McpResponse` and resolves in `<1ms`, while the actual CDP calls execute in `await response.handle(context, dataFormat)` (line 294, **outside** `#runToolWithTimeout`):
     * `take_snapshot`: `handler` only calls `response.includeSnapshot()`; `TextSnapshot.create()` (`Accessibility.getFullAXTree`) runs in `response.handle()`.
     * `list_pages`: `handler` only calls `response.setIncludePages(true)`; `context.createPagesSnapshot()` (`browser.pages()`) runs in `response.handle()`.
     * `get_network_request`: `handler` only calls `response.attachNetworkRequest(reqid)`; `httpResponse.buffer()` (`Network.getResponseBody`) runs in `response.handle()`.
     * `list_extensions`: `handler` only calls `response.setListExtensions(true)`; `context.listExtensions()` runs in `response.handle()`.
     If the CDP transport is silently dead, all of these tools will bypass `#runToolWithTimeout` and hang indefinitely in `response.handle()`, deadlocking `toolMutex`.
  2. **Post-timeout fallthrough into `response.handle()`**:
     When `this.tool.handler()` *does* time out after 60s, the `catch (err)` block at line 281 catches `ToolCallTimeoutError`, calls `response.setError(err)`, and **continues execution** to:
     * `devToolsData = await context.getDevToolsData(page)` (which adds an extra 500ms timeout wait on the dead browser), and
     * `await response.handle(context, dataFormat)` (line 294).
     If the timed-out handler had already called `response.includeSnapshot()` or `response.setIncludePages(true)` before hanging, `McpResponse.handle()` will attempt to query CDP on the dead `context`/`page` and hang forever *after* the 60s timeout has already fired.
  3. **Unbounded `this.getContext()`**:
     `await this.getContext()` (line 238) runs inside `this.toolMutex.acquire()` but outside `#runToolWithTimeout()`. If `puppeteer.connect()` or `McpContext.from()` (`browser.pages()`) hangs on a half-open socket during initial connect or reconnect, `getContext()` hangs forever while holding `toolMutex`.
* **Recommendation**:
  * Either wrap the entire CDP-interacting section inside `toolMutex.acquire()` (including `getContext()`, `tool.handler()`, and `response.handle()`) with the timeout, or
  * Immediately return the error `CallToolResult` from `handle()` when `ToolCallTimeoutError` is caught (do not fall through to `getDevToolsData` and `response.handle()`), AND include `response.handle()` within `#runToolWithTimeout()`.

#### 2. Resource Leak & Uncancelled Hung Promises in `forgetBrowser()`
* **Location**: `src/browser.ts:216-221` and `src/ToolHandler.ts:130-132`
* **Problem**:
  * In `src/browser.ts`, `forgetBrowser(candidate)` sets `browser = undefined; browserMode = undefined;` without disconnecting or closing `candidate`.
  * If `browserMode === 'launched'`, dropping the reference without calling `candidate.close()` leaves an **orphaned Chrome subprocess** (`candidate.process()`) running in the background.
  * If `browserMode === 'connected'`, dropping the reference without calling `candidate.disconnect()` leaves the WebSocket transport and event listeners open in memory.
  * Furthermore, the comment in `src/ToolHandler.ts:130-132` states:
    > `The loser of the race (a genuinely hung handler) is left running; there is no way to cancel a pending Puppeteer call...`
    In fact, calling `candidate.disconnect()` (or `candidate.close()`) **does cancel all pending Puppeteer CDP calls**: Puppeteer's `Connection` disposal synchronously rejects all in-flight CDP callbacks with `TargetCloseError` / `Connection closed`. Actively disconnecting/closing the forgotten browser on timeout will immediately settle the hung handler promise and allow its `finally` blocks / `DisposableStack` (e.g. element handles) to clean up instead of leaking in memory.
* **Recommendation**: In `forgetBrowser(candidate)` (or when handling a timeout), actively tear down `candidate` in the background: call `void candidate.close().catch(() => {})` if launched, or `void candidate.disconnect().catch(() => {})` if connected.

---

### `AGENTS.md` Rule Violations

#### 3. Non-Null Assertion Operator (`!`) in `src/ToolHandler.ts`
* **Location**: `src/ToolHandler.ts:154` (`clearTimeout(timer!);`)
* **Rule**: `AGENTS.md` -> *Rules for TypeScript*: `"Do not use ! operator for type assertion."`
* **Fix**: Declare `let timer: ReturnType<typeof setTimeout> | undefined;` and call `clearTimeout(timer);` (`clearTimeout(undefined)` is a safe no-op in Node.js).

#### 4. Type Casting (`as`) and Hand-Rolled Mock in `tests/browser.test.ts`
* **Location**: `tests/browser.test.ts:442` (`forgetBrowser({} as Browser);`)
* **Rule**: `AGENTS.md` -> *Rules for TypeScript*: `"Do not use as keyword for type casting."` and *Mocking Guidelines*: `"Do not hand-roll mock objects... Use sinon.createStubInstance()"` / `"Centralize mocks in tests/mocks.ts"`.
* **Fix**: Use `getMockBrowser()` from `tests/utils.js` / `tests/mocks.js`.

#### 5. Node `assert` Instead of `sinon.assert` for Spy Verification in `tests/ToolHandler.test.ts`
* **Location**: `tests/ToolHandler.test.ts:315-318` and `365`
  ```ts
  assert.strictEqual(
    forgetBrowserSpy.calledOnceWith(mockContext.browser),
    true,
  );
  ...
  assert.strictEqual(forgetBrowserSpy.called, false);
  ```
* **Rule**: `AGENTS.md` -> *Assertions and Sinon Best Practices*:
  > **Use `sinon.assert` methods**: Do not use Node's `assert.ok(stub.calledOnce)` or `assert.deepStrictEqual(stub.firstCall.args[0], ...)` to verify stub calls.
  > - Use `sinon.assert.calledOnceWithExactly(stub, ...args)` for single calls with exact arguments.
  > - Use `sinon.assert.notCalled(stub)` to assert a method was not invoked.
* **Fix**: Replace with:
  ```ts
  sinon.assert.calledOnceWithExactly(forgetBrowserSpy, mockContext.browser);
  ...
  sinon.assert.notCalled(forgetBrowserSpy);
  ```

src/ToolHandler.ts
bitbay: After looking into this, `request.params.timeout` isn't part of the MCP spec. The follow-up
> could we make this depend on actual MCP client timeout?

turns out to hit the same wall: the client's `timeout` option is purely local state (`Protocol.request()`, [`shared/protocol.js#L1215-L1218`](https://github.com/modelcontextprotocol/typescript-sdk/blob/1.30.0/src/shared/protocol.ts#L1215-L1218)). It is used to schedule a `setTimeout` inside the *client's* own `Protocol` instance to abort waiting on its side. It's never serialized into the outgoing JSON-RPC request (no `params.timeout`, no `_meta` field carrying it). The server has no visibility into whatever timeout the calling client is using - there's nothing on the wire to read.

So there's no spec-compliant way for the server to either honor a client-supplied timeout or infer the client's own timeout value. Given this, a server-defined `TOOL_CALL_TIMEOUT_MS` seems to be the only option available here for now.

tests/ToolHandler.test.ts
bitbay: Fixed in bitbay/chrome-devtools-mcp@3d9dd88

src/ToolHandler.ts
bitbay: Fixed in bitbay/chrome-devtools-mcp@3d9dd88

OrKoN (changes requested):
Could you please rebase the PR?

bitbay:
Rebased onto current main using the new BrowserManager, see the 2      updated commits.

PR #2705, opened 2026-09-09
feat(telemetry): persist date of the last tool call.

yulunz:
The state file is now loaded into memory and stored within ClearcutLogger. At each tool call it checks whether it's been more than one day since the recorded value in the state file, and if so, log the last tool active, and update the state file.

Regarding the state file's format, lastToolCall is not exactly the very last one, but rather the last one that was **logged**. This is the same with the existing lastActive timestamp (i.e. it's not the last timestamp when the server is active, but rather the last timestamp when the active status is logged). So here we keep the definitions consistent.

A caveat for future improvement: arguably the loadState() and writeState() and everything in between should be a atomic unit (there might be multiple mcp servers running), but I think it's probably good enough for now. If we see weirdness we can make that happen.

I also added format guards to loadState() so we don't have to try and catch everywhere else when de-serializing the string. We should actually just deserialize in loadState() but let me do this in a follow-up CL.

### DIFF
--- a/src/telemetry/ClearcutLogger.ts
+++ b/src/telemetry/ClearcutLogger.ts
@@ -44,6 +44,33 @@ function detectOsType(): OsType {
   }
 }
 
+function isSameDay(d1: Date, d2: Date): boolean {
+  return (
+    d1.getUTCFullYear() === d2.getUTCFullYear() &&
+    d1.getUTCMonth() === d2.getUTCMonth() &&
+    d1.getUTCDate() === d2.getUTCDate()
+  );
+}
+
+function shouldLogDailyActive(state: LocalState): boolean {
+  if (!state.lastActive) {
+    return true;
+  }
+  return !isSameDay(new Date(state.lastActive), new Date());
+}
+
+function calculateDaysSince(
+  lastDateString?: string,
+  now: Date = new Date(),
+): number {
+  if (!lastDateString) {
+    return -1;
+  }
+  const lastDate = new Date(lastDateString);
+  const diffTime = Math.abs(now.getTime() - lastDate.getTime());
+  return Math.ceil(diffTime / MS_PER_DAY);
+}
+
 export interface ClearcutLoggerOptions {
   appVersion: string;
   persistence: Persistence;
@@ -61,6 +88,7 @@ export class ClearcutLogger {
   #persistence: Persistence;
   #watchdog: WatchdogClient;
   #mcpClient: McpClient;
+  #state?: LocalState;
 
   static initialize(options: ClearcutLoggerOptions): ClearcutLogger {
     if (_clearcut_logger_instance) {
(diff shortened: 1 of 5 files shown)


tests/telemetry/ClearcutLogger.test.ts
```diff
@@ -318,6 +314,161 @@ describe('ClearcutLogger', () => {
     });
   });
 
+  describe('tool_active logging', () => {
+    it('logs tool active with -1 on first tool call when lastToolCall is not set', async () => {
+      mockPersistence.loadState.resolves({});
```
OrKoN, line 325: let's use sinon assertions for all assertions (see AGENTS.md for instructions).

src/telemetry/ClearcutLogger.ts
```diff
@@ -92,6 +120,14 @@ export class ClearcutLogger {
         clearcutIncludePidHeader: options.clearcutIncludePidHeader,
       });
     this.#mcpClient = McpClient.MCP_CLIENT_UNSPECIFIED;
+    void this.#persistence.loadState().then(
+      state => {
+        this.#state = state;
```
OrKoN, line 126: may this lead to telemetry being dropped while the state is being loaded?

src/telemetry/ClearcutLogger.ts
```diff
@@ -132,6 +168,8 @@ export class ClearcutLogger {
     devToolsData?: DevToolsData;
     pageUrl?: string;
   }): Promise<void> {
+    void this.#logToolActiveIfNeeded();
```
OrKoN: can this race with logDailyActiveIfNeeded overwriting the state?

src/telemetry/ClearcutLogger.ts
```diff
@@ -92,6 +120,14 @@ export class ClearcutLogger {
         clearcutIncludePidHeader: options.clearcutIncludePidHeader,
       });
     this.#mcpClient = McpClient.MCP_CLIENT_UNSPECIFIED;
+    void this.#persistence.loadState().then(
```
OrKoN: please attach a promise catch handler to all void promises

yulunz: done.

tests/telemetry/ClearcutLogger.test.ts
yulunz: done.

src/telemetry/ClearcutLogger.ts
yulunz: logDailyActiveIfNeeded is called at server start and before any mcp clients are connected (we know this because none of the daily active logs have mcp_client populated). Tool active log can only happen after a client has connected, so the two paths are mutually exclusive.

src/telemetry/ClearcutLogger.ts
yulunz: Yes, theoretically.

We can await on this.#state being available when logging the tool active, but I don't want to block tool call because of telemetry (say there is a file access issue and the load hangs forever). The ClearcutLogger is constructed at server start, and that's some long time before the first tool call is made, so we are reasonably confident that this.#state will become available at the first tool call. 

And this why I choose to skip tool active logging if this.#state is still undefined by then. I think we can afford some slight inaccuracies to the telemetry to preserve a clean unblock execution path for tool call.

OrKoN (approved):
LGTM!

PR #2720, opened 2026-09-10
fix(performance): prevent memory leak by scoping trace engine model per parse

jackfranklin:
## Rationale

DevTools' `TraceEngine.TraceModel.Model` permanently retains parsed traces in an internal recordings array (`#traces`). Calling `engine.resetProcessor()` between runs only resets the underlying processor and does not clear `#traces`.

Because `engine` was previously instantiated as a module-scoped singleton, every performance trace recorded or parsed in a session remained permanently held in heap memory for the lifetime of the process.

## Changes

- In `src/processors/PerformanceTrace.ts`, instantiate `DevTools.TraceEngine.TraceModel.Model.createWithAllHandlers()` per invocation of `parseRawTraceBuffer()`.
- Remove the module-scoped `engine` singleton and the obsolete `engine.resetProcessor()` call.
- Add a unit test in `tests/trace-processing/parse.test.ts` verifying that consecutive trace parses construct isolated `Model` instances and keep internal trace retention bounded at 1.
- Add a profiling regression scenario in `scripts/profile/scenarios/performance_trace.ts` to guard against memory retention across consecutive traces during `npm run test:memory`.

### DIFF
--- a/scripts/profile/scenarios/performance_trace.ts
+++ b/scripts/profile/scenarios/performance_trace.ts
@@ -1,0 +1,47 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type {ScenarioArgs, ScenarioIterations, ToolCall} from '../types.ts';
+
+/**
+ * Defines iteration counts for the performance trace profiling scenario.
+ *
+ * Keeps iteration counts bounded so profiling remains fast while still providing
+ * enough repetitions to detect memory retention across consecutive traces.
+ */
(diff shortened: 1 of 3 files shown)


jackfranklin:
PTAL. I found this when debugging the buffer overflow issue. Until we support >1 traces at once, let's ensure we clean up the memory.

OrKoN (commented):
Could you please add a regression-scenario to https://github.com/ChromeDevTools/chrome-devtools-mcp/tree/main/scripts/profile/scenarios?

OrKoN (approved):

PR #2721, opened 2026-09-10
feat(performance): chunked trace buffer parser for large recordings

jackfranklin:
> [!NOTE]
> This PR is marked as **WIP (Work in Progress)** / Draft to gather feedback and extra opinions on the streaming / chunked parsing strategy for large performance traces.

## Problem Statement

When recording performance traces on busy sites or over long recording durations, the resulting trace buffer often exceeds 512 MB (and can reach 1 GB+).

Currently, `parseRawTraceBuffer` decodes the entire raw buffer into a single string via `new TextDecoder().decode(buffer)` and parses it with `JSON.parse(asString)`. In V8 (64-bit), strings are strictly limited to $2^{29} - 24$ characters (~512 MB). Any trace exceeding ~512 MB throws a fatal error:
```
RangeError: Cannot create a string longer than 0x1fffffe8 characters
```
This prevents the MCP server from parsing and analyzing large traces even when sufficient heap memory is available.

## Proposed Solution: Chunked Trace Buffer Scanner

This change introduces `ChunkedTraceParser.ts` to extract trace events directly from the raw binary buffer in bounded batches without materializing the full trace JSON into a single string:

1. **Byte-Level Delimiter & Escape Scanning**:
   - Scans the `Uint8Array` directly using ASCII byte codes (`0x7B` `{`, `0x7D` `}`, `0x22` `"`, `0x5C` `\`).
   - Tracks brace nesting depth while properly handling escaped quotes within strings.
   - Supports both bare JSON arrays (`[ {...}, ... ]`) and top-level object envelopes (`{ "traceEvents": [ ... ], "metadata": { ... } }`).

2. **Batched Native `JSON.parse`**:
   - Accumulates slices of events into bounded batches (default: 5,000 events) wrapped in `[` and `]`.
   - Parses each batch using native V8 `JSON.parse`, keeping intermediate string allocations small (< 5 MB) and short-lived.
   - Pushes parsed events iteratively to avoid V8 function call argument stack limits.

3. **Metadata Preservation**:
   - Extracts top-level metadata from object envelopes (such as `cpuThrottling` and `networkThrottling`) and merges it with caller options.

4. **Strict Structural Validation**:
   - Enforces valid structural tokens (closing brackets, braces, colons, commas) and prevents silent truncation on incomplete buffers or unexpected EOF.

## Testing

- Unit tests in `tests/trace-processing/ChunkedTraceParser.test.ts` covering:
  - Both top-level array and object formats.
  - Custom batch sizes and large batches exceeding function call argument limits (70,000+ events).
  - Malformed inputs: truncated arrays, unterminated strings, unquoted keys, trailing characters.
  - Multibyte UTF-8 characters split across byte boundaries.
- Full regression tests against existing trace fixtures (`basic-trace.json.gz`, `web-dev-with-commit.json.gz`).


jackfranklin:
The buffer limit in Chrome Tracing does not align with the final JSON output, because it limits the size of the binary protobuf that is captured. So I also think separately we might want to consider reducing the limit from 1.2gb.

jackfranklin:
Please note also this PR is a WIP and is not fully ready for code review, sharing to get early thoughts on if we want to pursue it.

jackfranklin:
@OrKoN PTAL. Not fully ready yet but keen to get your early opinion on this fix.

OrKoN:
Some AI-advice that seems to make sense:

#### 1. Batching Only by Event Count (`eventsPerBatch = 10_000`) Can Still Hit V8's Max String Length Limit
* **Location**: `src/processors/ChunkedTraceParser.ts:293` (`if (batchCount >= eventsPerBatch) flushBatch();`)
* **Problem**:
  * V8's string limit (`RangeError: Invalid string length`, `~512 MB`) is a **byte/character length limit**, not an event count limit.
  * Chrome traces frequently include events with very large payloads, such as base64-encoded screenshots (`disabled-by-default-devtools.screenshot`), inline script source texts/source maps (`v8.compile`), or heap profiler chunks. A batch of 10,000 events containing screenshots or large script sources can easily exceed 256–512 MB.
  * Furthermore, `flushBatch()` constructs `` `[${chunkText}]` `` (line 241), allocating a second copy of the entire batch string in memory and doubling peak string allocation per batch.
  * Additionally, if `options.eventsPerBatch` is passed as `NaN`, `batchCount >= eventsPerBatch` is always `false`, causing the parser to buffer the entire file into a single `flushBatch()` call at the end.
* **Recommendation**:
  * Enforce a **byte-size limit** per batch alongside `eventsPerBatch` (e.g. `const MAX_BATCH_BYTES = 32 * 1024 * 1024;` and flush when `batchCount >= eventsPerBatch || (lastEventEnd - batchStart) >= MAX_BATCH_BYTES`).
  * Validate `eventsPerBatch` (`if (!Number.isFinite(eventsPerBatch) || eventsPerBatch <= 0)`).

#### 2. `undefined` Properties in Caller `metadata` Clobber Extracted File Metadata in Production
* **Location**: `src/processors/PerformanceTrace.ts:534-536`
  ```ts
  const combinedMetadata = fileMetadata
    ? {...fileMetadata, ...metadata}
    : metadata;
  ```
* **Problem**:
  * In production (`src/tools/performance.ts:224-227`), `parseRawTraceBuffer` is called as:
    ```ts
    const result = await parseRawTraceBuffer(traceEventsBuffer, {
      cpuThrottling: page.cpuThrottlingRate,
      networkThrottling: page.networkConditions ?? undefined,
    });
    ```
  * When network throttling is not active on `page`, `page.networkConditions` is `null`, so `metadata` is `{ cpuThrottling: 1, networkThrottling: undefined }`.
  * In JavaScript, `{ ...fileMetadata, ...metadata }` overwrites `fileMetadata.networkThrottling` with `undefined` because `networkThrottling` is an own property on `metadata`.
  * Similarly, `page.cpuThrottlingRate` defaults to `1` even when never explicitly configured via `emulate`, which clobbers any `cpuThrottling` value stored in `fileMetadata`.
  * Note: The snapshot diffs in `tests/McpResponse.test.js.snapshot` and `tests/trace-processing/parse.test.js.snapshot` changed to `1x` / `No throttling` only because those tests call `parseRawTraceBuffer(rawData)` with `metadata === undefined`.
* **Recommendation**:
  * Filter out `undefined` properties from `metadata` before merging so they do not overwrite defined properties in `fileMetadata`:
    ```ts
    const combinedMetadata = {
      ...fileMetadata,
      ...(metadata?.cpuThrottling !== undefined ? {cpuThrottling: metadata.cpuThrottling} : {}),
      ...(metadata?.networkThrottling !== undefined ? {networkThrottling: metadata.networkThrottling} : {}),
    };
    ```

#### 3. `parseEventsArray` Silently Ignores Tokens at `depth === 0` (Missing Commas, Trailing Commas, Non-Object Items, and Nested Brackets)
* **Location**: `src/processors/ChunkedTraceParser.ts:251-308`
* **Problem**:
  Because `flushBatch()` only slices from `batchStart` (the opening `{` of the first event in a batch) to `lastEventEnd` (the closing `}` of the last event in a batch), any bytes at `depth === 0` before the first `{`, between batches, or after the last `}` before `]` are never passed to `JSON.parse` and are ignored by the scanner:
  1. **Missing commas at batch boundaries**: If `eventsPerBatch` is 2, `[{"name":"e1"},{"name":"e2"} {"name":"e3"},{"name":"e4"}]` (missing comma between `e2` and `e3`) parses without error.
  2. **Trailing commas**: `[{"name":"e1"}, ]` parses without throwing `SyntaxError`.
  3. **Non-object array items silently dropped**: In `[null, "string", 123, {"name":"e1"}]`, the non-object items before `{"name":"e1"}` are silently skipped and omitted from `events`.
  4. **Premature termination on array elements (`[]`)**: `parseEventsArray` does not track `BYTE_OPEN_BRACKET` (`[`). If `traceEvents` contains an array element (e.g. `[ [], {"name":"e1"} ]`), the closing `]` of `[]` is encountered at `depth === 0` and triggers line 301 (`if (byte === BYTE_CLOSE_BRACKET && depth === 0)`), prematurely terminating `parseEventsArray`.

#### 4. `skipValue` Mismatched Delimiter Bug & Missing Value Handling
* **Location**: `src/processors/ChunkedTraceParser.ts:155-205`
* **Problem**:
  1. **Single `depth` counter for `{}` and `[]`**: `skipValue` increments `depth` on both `{` and `[`, and decrements on both `}` and `]`. If a skipped property precedes `traceEvents` and contains mismatched delimiters (e.g., `{"other": {"a": ]}, "traceEvents": [...]}`), `depth` reaches `0` at `]`, causing `skipValue` to return early. The outer object loop then sees the remaining `}` of `"other"` and mistakes it for the closing brace of the root JSON object, terminating parsing before `"traceEvents"` is read.
  2. **Empty primitive values**: In primitive scanning (lines 190-205), if a property value is missing before a comma (`{"other": , "traceEvents": []}`), the loop immediately breaks on `BYTE_COMMA` after advancing `0` bytes, silently accepting invalid JSON.
  3. **Out-of-bounds index on trailing backslash**: If a skipped compound object/array ends with a trailing backslash `\` at `buffer[buffer.length - 1]`, `pos += 2` (line 164) advances `pos` to `buffer.length + 1` and returns an index strictly greater than `buffer.length`.

#### 5. Unescaped Object Keys in `parseTraceEventsFromBuffer`
* **Location**: `src/processors/ChunkedTraceParser.ts:390-393`
* **Problem**:
  `const key = decoder.decode(buffer.subarray(keyStart, keyEnd));` decodes the raw bytes between quotes without `JSON.parse`. If a JSON property key contains escape sequences (e.g. `\"`, `\\`, or `\uXXXX`), `key` will contain the raw unescaped sequence rather than the parsed string value.
* **Recommendation**: Parse the key slice using `JSON.parse(decoder.decode(buffer.subarray(keyStart - 1, pos)))`.

jackfranklin:
@OrKoN PTanotherL

src/processors/ChunkedTraceParser.ts
```diff
@@ -0,0 +1,464 @@
...
+      }
+      if (byte === BYTE_CLOSE_BRACE) {
+        depth--;
+        if (depth === 0) {
+          lastEventEnd = i + 1;
+          batchCount++;
+          state = 'expect_comma_or_end';
+          if (
+            batchCount >= config.eventsPerBatch ||
+            lastEventEnd - batchStart >= config.maxBatchBytes
```
OrKoN: I think this can overshoot the limit and we should check before we add the batchCount++?

OrKoN (approved):

src/processors/ChunkedTraceParser.ts
jackfranklin: Given that we can't know the byte size of an event until we have parsed it all, I think without a lot of extra work we cannot guarantee that this case doesn't happen where we go 1 event's worth over the max bytes.

However, given we set the bytes at 32mb, and the V8 string alloc limit is ~512mb, I think going over here by one event's worth of space, isn't worth any extra complexity in the code. wdyt? 

(I will add an inline comment I think)

Issue #2233, opened 2026-06-19
Improve configuration section in readme with (implicit) defaults


mathiasbynens:
### Task to do:

https://github.com/ChromeDevTools/chrome-devtools-mcp#configuration includes e.g. “Default: `true`” for some entries like `--categoryPerformance`, but not for others like `--memoryDebugging` (which is “Default: `false`”). We should make it explicitly clear what the defaults are for every option.

PR #2722, opened 2026-09-10
fix: stop inventing a `false` default for options without one

dajiaohuang:
## Summary

`generateConfigOptionsMarkdown()` falls back to the literal string `'false'` when an option declares neither `defaultDescription` nor `default`:

```js
markdown += `  - **Default:** \`${optionConfig.defaultDescription ?? optionConfig.default ?? 'false'}\`\n`;
```

As a result, every string, number and array config option is documented as having a default of `false`. Of the 45 documented options, 30 declare no default and are affected. Only 15 declare one (8 `true`, 6 `false`, and `--filesystemRoot` with `defaultDescription: 'OS temp directory'`).

### Relation to #2233 / #2234 — please read before merging

This fallback was introduced **deliberately** by #2234 (merged 2026-06-19, closing #2233 "Improve configuration section in readme with (implicit) defaults"). #2234 replaced a guarded emission with an unconditional one:

```diff
-    if (optionConfig.default !== undefined) {
-      markdown += `  - **Default:** \`${optionConfig.default}\`\n`;
-    }
+    markdown += `  - **Default:** \`${optionConfig.default ?? 'false'}\`\n`;
```

So this PR removes lines that #2234 intentionally added, and I'd rather flag that than present it as an unambiguous bug fix.

The reason I believe it is still a defect: #2233's motivating example was `--memoryDebugging`, a **boolean** whose implicit default genuinely is `false`. Generalising `?? 'false'` to every option makes the output factually wrong for non-boolean options. For example `--screenshotMaxWidth` (type `number`) currently reads:

```
Maximum width in pixels for screenshots. ... Unset means no resize.
- **Type:** number
- **Default:** `false`
```

and `--screenshotQuality` and `--experimentalScreencastFps` are numbers documented as defaulting to a boolean.

**If you would prefer to keep a `Default:` line on every option**, the alternative is to emit each option's true implicit default (e.g. `unset`, or wording derived from the option's existing description) rather than omitting the line. I did not attempt that here because it needs per-option judgement — happy to rework the PR that way instead.

## Verification

- `npm run gen` — completes; the only diff is the 30 removed doc lines plus the generator edit.
- Re-running `npm run gen` is idempotent (`sha256` of `docs/configuration.md` unchanged), so `[Required] Check docs updated` stays green.
- `npm run check-format` — passed.
- `npm run typecheck` — passed.
- `git diff --check` — passed.

No issue is linked; reproduced on the current `main` (`d05cbc05`).

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -22,17 +22,14 @@ The Chrome DevTools MCP server supports the following configuration option:
 - **`--categoryExtensions`/ `--category-extensions`**
   Set to true to include tools related to extensions. Note: This feature is currently only supported with a pipe connection. autoConnect, browserUrl, and wsEndpoint are not supported with this feature until 149 will be released.
   - **Type:** boolean
-  - **Default:** `false`
 
 - **`--categoryExperimentalThirdParty`/ `--category-experimental-third-party`**
   Set to true to enable third-party developer tools exposed by the inspected page itself
   - **Type:** boolean
-  - **Default:** `false`
 
 - **`--categoryPwa`/ `--category-pwa`**
   Set to true to include tools for automating Progressive Web Apps (install, launch, uninstall, and OS state). This feature is only supported with a pipe connection; autoConnect, browserUrl, and wsEndpoint are not supported.
   - **Type:** boolean
-  - **Default:** `false`
 
 - **`--autoConnect`/ `--auto-connect`**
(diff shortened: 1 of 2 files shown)


nattallius:
Hi, can you fix sign the CLA check? That  happened because you co-authored commit with noreply@anthropic.com

PR #2728, opened 2026-09-11
fix: block javascript links when evaluation is disabled

### LINKED ISSUE #2726: --no-javascript-evaluation can be bypassed by clicking javascript: links
### Description of the bug

When chrome-devtools-mcp is started with `--no-javascript-evaluation`, direct JavaScript evaluation and direct navigation to `javascript:` URLs are blocked, but clicking an existing `javascript:` hyperlink still executes the script.

That makes the flag inconsistent: the protected side effect (JavaScript execution) is still reachable through the click path.

This was reviewed through Google's security process and was assessed as not meeting the bar for a standalone security vulnerability because an attacker would already need control of page content. The recommendation was to file it here as a regular functional issue, so I am reporting this as a functional/policy-consistency bug rather than a security issue.

### Reproduction

1. Start chrome-devtools-mcp with `--no-javascript-evaluation`.
2. Open a page you control containing a link such as:
   ```html
   <a id="js-link" href="javascript:document.body.dataset.mcpJs='executed'">Run</a>
   ```
3. Confirm the direct path is blocked (for example, direct navigation to a `javascript:` URL is rejected while the flag is enabled).
4. Use the MCP click action to click `#js-link`.
5. Observe that the JavaScript executes and the page state changes (`document.body.dataset.mcpJs === 'executed'`).

I reproduced the behavior on chrome-devtools-mcp 1.8.0. The original current-main verification was on commit `b2007ab86639a0ae4e4370484561158773112b93`.

### Expectation

With `--no-javascript-evaluation` enabled, MCP-exposed actions should not be able to trigger JavaScript evaluation through `javascript:` links. The click should be rejected, or the `javascript:` navigation should be blocked before execution, consistent with the direct-navigation protection.

### MCP configuration

Local build / equivalent MCP launch command:

```text
node build/src/bin/chrome-devtools-mcp.js --no-javascript-evaluation
```

No special page privileges are required for the reproduction; the test page is locally controlled.

1.8.0

### Chrome version

153.0.8010.36

### Coding agent version

N/A - reproduced with a direct MCP test harness

### Model version

N/A

### Chat log

N/A

### Node version

v24.15.0

### Operating system

Windows

### Extra checklist

ITSMERNB:
Fixes #2726

## Summary
- reuse the existing JavaScript-evaluation URL validation for click-triggered navigations
- inspect the clicked element (or link ancestor) before dispatching the click
- block javascript:, data:, and vbscript: link navigations when --no-javascript-evaluation is active
- keep normal clicks and regular event-handler interactions unchanged

## Tests
- npm run typecheck
- npm run build
- npm run test:no-build -- tests/tools/input.test.ts
- npm run test:no-build -- tests/utils/url.test.ts
- git diff --check

### DIFF
--- a/src/tools/input.ts
+++ b/src/tools/input.ts
@@ -10,6 +10,7 @@ import type {ElementHandle, KeyInput} from '../third_party/index.js';
 import type {TextSnapshotNode} from '../types.js';
 import {parseKey} from '../utils/keyboard.js';
 import {logger} from '../utils/logger.js';
+import {validateJavaScriptEvaluationUrl} from '../utils/url.js';
 import type {WaitForEventsResult} from '../utils/WaitForHelper.js';
 
 import {ToolCategory} from './categories.js';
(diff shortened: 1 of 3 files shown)


OrKoN:
Closing per https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2726#issuecomment-5631648885

PR #2732, opened 2026-09-12
feat: allow new pages to default to background

brandonp2412:
## Summary

- add `--defaultBackground` / `--default-background`
- use it when `new_page.background` is omitted
- preserve an explicit per-call `background` value

This keeps the existing foreground default unchanged while allowing side-agent setups to opt into background tabs without repeating `background: true` on every `new_page` call. It intentionally does not change `select_page` or other focus behavior.

## Tests

- `npm run gen`
- `npm run check-format`
- `node scripts/test.js tests/tools/pages.test.ts`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -110,6 +110,11 @@ The Chrome DevTools MCP server supports the following configuration option:
   - **Type:** boolean
   - **Default:** `false`
 
+- **`--defaultBackground`/ `--default-background`**
+  Open new pages in the background by default.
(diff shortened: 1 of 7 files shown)


samiyac:
Hi, before a feature PR can be created, we require a feature request to be opened, discussed with the team and marked as confirmed.

Can you please file a feature request? For now, I am marking this PR as closed. Please feel free to open it once the request is confirmed. Thank you!

PR #2734, opened 2026-09-12
fix(logger): puppeteerLogger should check DEBUG before checking --log-file

### LINKED ISSUE #2731: --log-file unconditionally writes all puppeteer:protocol traffic since v1.8.0 (PR #2562), ignoring DEBUG, causing log files to grow to GB scale
### Description of the bug

Since v1.8.0, when --log-file is set, the puppeteerLogger function in src/utils/logger.js unconditionally writes all puppeteer:protocol:SEND and puppeteer:protocol:RECV messages to the log file, regardless of the DEBUG environment variable. This is a regression introduced by [#2562](https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2562) ("refactor: replace debug with NodeJS native").

In v1.7.0, logger.js used the debug npm package. The saveLogsToFile() function called debug.enable() with the user's DEBUG namespaces + mcp:log, so only explicitly enabled namespaces were written to the file.

In v1.8.0+, puppeteerLogger() has a branching bug — when logFileStream is truthy (i.e. --log-file is set), it returns a write function unconditionally without checking util.debuglog(prefix).enabled or DEBUG:

```javascript
// v1.8.0+ src/utils/logger.js
export const puppeteerLogger = (prefix) => {
    if (logFileStream) {
        //  ALWAYS returns a logger — DEBUG is never checked
        return (...args) => {
            logFileStream.write(`${new Date().toISOString()} ${prefix} ${util.format(...args)}\n`);
        };
    }
    // DEBUG is only checked when there is NO log file
    const dbg = util.debuglog(prefix);
    return dbg.enabled
        ? (...args) => { dbg('%s %s', new Date().toISOString(), util.format(...args)); }
        : undefined;
};
```

The result: every CDP wire message (network requests, responses, DOM events, accessibility tree queries, etc.) is written to disk. In real-world use, this produces gigabytes of log files, compared to kilobytes previously for the same amount of traffic.

### Reproduction

Testing identical operations on v1.7.0 vs v1.9.0 with `DEBUG=mcp:log` (should exclude `puppeteer:*`):

1. Configure MCP with `--log-file /path/to/log` and `DEBUG=mcp:log`
2. Truncate the log file
3. Reload the MCP server
4. Call `list_pages`, then `navigate_page` to `https://example.com`, then `take_snapshot`
5. Count log lines by namespace

**Results:**

| | v1.7.0 | v1.9.0 |
|---|---|---|
| **Total lines** | 26 | 209 |
| **`puppeteer:*` lines** | **0** | **174** |
| **`mcp:log` lines** | 16 | 17 |
| **`DEBUG` respected?** | Yes | No |

v1.7.0 correctly writes only `mcp:log` entries. v1.9.0 dumps 174 additional `puppeteer:protocol:SEND/RECV` lines containing full CDP JSON payloads (accessibility trees, network events, etc.) despite `DEBUG=mcp:log`.

### Expectation

`puppeteerLogger` should respect `DEBUG` namespace filtering when writing to `--log-file`, matching v1.7.0 behavior:

- `DEBUG` unset → only `mcp:log` lines in the log file
- `DEBUG=mcp:log` → only `mcp:log` lines
- `DEBUG=puppeteer:*` → puppeteer lines included (opt-in)
- `DEBUG=*` → everything (explicit user choice)

A minimal fix would be to check `util.debuglog(prefix).enabled` before returning a file-writing logger:

```javascript
export const puppeteerLogger = (prefix) => {
    const dbg = util.debuglog(prefix);
    if (!dbg.enabled) return undefined;
    if (logFileStream) {
        return (...args) => {
            logFileStream.write(`${new Date().toISOString()} ${prefix} ${util.format(...args)}\n`);
        };
    }
    return (...args) => {
        dbg('%s %s', new Date().toISOString(), util.format(...args));
    };
};
```

### MCP configuration

```json
{
  "mcpServers": {
    "chrome-devtools": {
      "command": "npx",
      "args": [
        "-y",
        "chrome-devtools-mcp@latest",
        "--log-file",
        "/path/to/chrome-devtools-mcp.log"
      ],
      "env": {
        "DEBUG": "mcp:log"
      }
    }
  }
}
```

v1.8.0 and v1.9.0 affected; v1.7.0 works correctly

### Chrome version

Chromium 150.0.7871.181

### Coding agent version

Cursor 3.18.9

### Model version

_No response_

### Chat log

_No response_

### Node version

v24.19.0

### Operating system

Linux

### Extra checklist

mittalpk:
Fixes #2731.

### Problem

Since v1.8.0 (#2562, replacing the `debug` npm package with Node's native `util.debuglog`), `puppeteerLogger()` in `src/utils/logger.ts` checks `logFileStream` truthiness *before* checking whether the namespace is DEBUG-enabled:

```ts
export const puppeteerLogger = (prefix: string) => {
  if (logFileStream) {
    return (...args) => { logFileStream!.write(...); };  // unconditional
  }
  const dbg = util.debuglog(prefix);
  return dbg.enabled ? (...) => dbg(...) : undefined;      // gated
};
```

So once `--log-file` is set, every `puppeteer:protocol:SEND`/`RECV` message (full CDP JSON payloads — network events, accessibility trees, etc.) gets written to the log file regardless of `DEBUG`, producing GB-scale log files instead of the KB-scale v1.7.0 behavior for the same traffic.

In v1.7.0, `saveLogsToFile()` called `debug.enable()` with the user's `DEBUG` namespaces plus `mcp:log`, so only explicitly-enabled namespaces were ever written — `mcp:log` always (since it was always added to the enabled set), everything else only if the user opted in.

### Fix

Reordered `puppeteerLogger` to check `dbg.enabled` first, so both the file-writing and `util.debuglog`-writing branches are gated the same way — matching v1.7.0 semantics. `logger()` (the separate function backing the always-on `mcp:log` namespace) is untouched; its unconditional file-write is correct, not a bug — `mcp:log` was always in the enabled set even in v1.7.0.

### How did you test it

Added `tests/utils/logger.test.ts` + a `tests/utils/fixtures/puppeteer-logger-fixture.ts` helper. `util.debuglog(...).enabled` reflects `NODE_DEBUG` as of process start, not a value togglable at runtime, so the fixture is spawned as a fresh child process per case (4 cases: namespace disabled + log file → file stays empty; namespace enabled + log file → file gets the message; `NODE_DEBUG=*` + log file → file gets the message; namespace disabled + no log file → returns `undefined`, matching existing behavior).

Verified fail-before/pass-after by running the equivalent logic directly against both the pre-fix and post-fix source (via `git stash`) — pre-fix, the "disabled + log file" case wrongly writes to the file; post-fix, it doesn't.

`eslint` and `prettier --check` both clean on all 3 changed/added files. I was not able to run the full `npm test` (`tsc && node scripts/post-build.ts` needs the `third_party/devtools-frontend` git submodule, which failed to fetch in my environment — appears to be a shallow-submodule/pinned-commit mismatch, unrelated to this change) or the full project `tsc` (blocked by the same missing submodule via `src/third_party/index.ts`'s import of it). I isolated my 3 changed files with a standalone `tsc --noEmit` invocation instead — confirmed zero type errors originate from any of them (the errors present in that run are 100% pre-existing, from unrelated `puppeteer-core`/`rxjs`/submodule type declarations, and disappear if those files are excluded). Happy to re-verify against a full build if a maintainer can point me at a working submodule fetch, or if this is a known environment quirk.

### DIFF
--- a/tests/utils/logger.test.ts
+++ b/tests/utils/logger.test.ts
@@ -1,0 +1,112 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import assert from 'node:assert';
+import {execFileSync} from 'node:child_process';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import process from 'node:process';
+import {describe, it} from 'node:test';
+import {fileURLToPath} from 'node:url';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const FIXTURE = path.join(__dirname, 'fixtures', 'puppeteer-logger-fixture.js');
+
+/**
+ * Runs the puppeteer-logger-fixture in a fresh child process with the given
+ * NODE_DEBUG value. A fresh process is required because
+ * `util.debuglog(...).enabled` reflects NODE_DEBUG as of process start, not
+ * a value that can be toggled at runtime (see the fixture's own comment).
+ */
+function runFixture(
+  prefix: string,
+  logFilePath: string | undefined,
+  nodeDebug: string | undefined,
+): string {
+  const env = {...process.env};
+  if (nodeDebug === undefined) {
+    delete env.NODE_DEBUG;
+  } else {
+    env.NODE_DEBUG = nodeDebug;
+  }
+  return execFileSync(
+    process.execPath,
+    [FIXTURE, prefix, logFilePath ?? '--'],
+    {env, encoding: 'utf8'},
+  ).trim();
+}
+
+describe('puppeteerLogger', () => {
+  it('does not log to a file when the namespace is not DEBUG-enabled', () => {
+    const logFilePath = path.join(
+      os.tmpdir(),
+      `puppeteer-logger-test-${process.pid}-${Date.now()}.log`,
+    );
+    try {
+      const result = runFixture(
+        'puppeteer:protocol:SEND',
+        logFilePath,
+        'mcp:log', // DEBUG is set, but does not include this prefix.
+      );
+      assert.strictEqual(result, 'logger:undefined');
+      // saveLogsToFile() creates the file unconditionally (--log-file is a
+      // global option, independent of any specific namespace), so the
+      // regression check is on CONTENT, not existence: the file must stay
+      // empty rather than receiving this namespace's traffic.
+      const contents = fs.existsSync(logFilePath)
+        ? fs.readFileSync(logFilePath, 'utf8')
+        : '';
+      assert.strictEqual(
+        contents,
+        '',
+        'puppeteerLogger must not write to the log file for a namespace ' +
+          'that is not enabled via NODE_DEBUG, even when --log-file is set',
+      );
+    } finally {
+      fs.rmSync(logFilePath, {force: true});
+    }
+  });
+
+  it('logs to a file when the namespace is DEBUG-enabled', () => {
+    const logFilePath = path.join(
+      os.tmpdir(),
+      `puppeteer-logger-test-${process.pid}-${Date.now()}.log`,
+    );
+    try {
+      const result = runFixture(
+        'puppeteer:protocol:SEND',
+        logFilePath,
+        'puppeteer:protocol:SEND',
+      );
+      assert.strictEqual(result, 'logger:defined');
+      const contents = fs.readFileSync(logFilePath, 'utf8');
+      assert.match(contents, /puppeteer:protocol:SEND fixture message/);
+    } finally {
+      fs.rmSync(logFilePath, {force: true});
+    }
+  });
+
+  it('logs to a file when NODE_DEBUG=* enables every namespace', () => {
+    const logFilePath = path.join(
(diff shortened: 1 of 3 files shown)


tests/utils/logger.test.ts
```diff
@@ -0,0 +1,112 @@
...
+import {execFileSync} from 'node:child_process';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import process from 'node:process';
+import {describe, it} from 'node:test';
+import {fileURLToPath} from 'node:url';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const FIXTURE = path.join(__dirname, 'fixtures', 'puppeteer-logger-fixture.js');
```
Lightning00Blade:
```suggestion
const FIXTURE = path.join(import.meta.dirname, 'fixtures', 'puppeteer-logger-fixture.js');
```

Lightning00Blade:
Thank for the PR but there are simpler ways to test this. Please refer to https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2743

PR #2738, opened 2026-09-13
feat: add cookie management tools

steffen-heil-secforge:
> [!IMPORTANT]
> **Depends on #2737.** This branch is stacked on it, so the diff currently shows both commits; once #2737 lands, only the cookie commit remains.
>
> The dependency is real, not just ordering: `set_cookie` resolves `{{secret:NAME}}` placeholders, which #2737 introduces. A cookie value is usually a session token, so passing it as literal text would write it into the conversation transcript -- exactly the leak #2737 exists to close. Without that PR, `set_cookie` would reintroduce it on the write side while `get_cookies` avoids it on the read side.

Adds `get_cookies`, `set_cookie` and `clear_cookies`, covering the inspect / edit / delete cases raised in #408.

## Why a tool rather than the skill

The `cookie-debugging` skill documents what is reachable today, and is explicit that two things are not:

- **`HttpOnly` cookies** are invisible to `cookieStore` and `document.cookie`. The skill's workaround is to find them in the `Cookie` header of a captured request.
- **Cookies for domains the page never requested** never appear in the network tools at all, so a profile cannot be enumerated.

That leaves real gaps. #408 asked directly *"How important is it to support HttpOnly cookies?"*, and the answer in the thread was *"Would want this... Or else, we'll still have to use playwright for HttpOnly for extraction."* These tools go through CDP (`Storage.getCookies`, `Network.setCookie`, `Network.deleteCookies`), so both cases are covered.

### `get_cookies`

Enumerates every cookie in the browser's default context, optionally filtered by domain.

Cookie values are **written to a JSON file and never returned inline**. The response reports only each cookie's name, domain, path, expiry and security flags. This matters: on a normal profile this tool finds hundreds of cookies, including live session tokens for every site the user is signed in to, and returning those inline would put them into the conversation transcript (and any log of it) wholesale. `filePath` is therefore required rather than optional.

### `set_cookie`

Sets a cookie; `url` or `domain` is required so the browser knows where it applies. The value accepts `{{secret:NAME}}` (see #2737), so a session token never has to be passed as literal text. Supports `path`, `expires`, `httpOnly`, `secure` and `sameSite`. The value is not echoed back in the response. A rejection by the browser is reported with the likely causes (domain mismatch, or `sameSite: "None"` without `secure`).

### `clear_cookies`

Deletes cookies matching `name` and/or `domain`.

Clearing everything signs the user out of every site, so the tool **refuses to do it implicitly**: with no filter given it errors and asks for an explicit `all: true`. This is deliberate — a model reaching for "clear cookies" to reset one site should not silently destroy every session in the profile.

## Testing

`tests/tools/cookies.test.ts` covers: writing values to the file while keeping them out of the response, the domain filter, `set_cookie` rejecting a call with neither `url` nor `domain`, `set_cookie` actually setting a cookie without echoing its value, `clear_cookies` refusing an unfiltered wipe, and `clear_cookies` deleting only the matching cookie.

These require a browser, so they have not been run in my environment — they follow the existing patterns in that directory and I would appreciate CI confirming them. `npm run check-format`, `tsc --noEmit` and the tool registration (62 tools, all three present with the expected schemas) were all verified locally.

## Details worth flagging

- **Partitioned cookies**: `clear_cookies` passes each cookie's `partitionKey` to `Network.deleteCookies`. Without it a CHIPS cookie is not matched and silently survives a delete that reported success.
- **Scope**: all three act on the **default browser context**. Cookies in an isolated context created via `new_page({isolatedContext})` are not seen or cleared. The skill now says so.
- **Skill updated**: `cookie-debugging` currently states that HttpOnly cookies cannot be read or deleted and points at network headers as the workaround. That is no longer the whole picture, so its capability table, the HttpOnly warning and two troubleshooting entries now mention these tools. The changes are deliberately minimal.


OrKoN:
Thanks for the PR but we do not plan to add explicit cookie tools at this point. Please express your interest in the feature on a feature request.

PR #2747, opened 2026-09-15
fix(config): preserve raw values for config coercion

### LINKED ISSUE #2746: JSON config fails for viewport and wsHeaders
### Description of the bug

The JSON config support added in #2661 fails for options whose coercion converts a string into an object.

On main at `17c671899b557d63e76e21f18af323616dd2ee8e`, using this config:

```json
{"viewport":"1280x720"}
```

produces `arg.split is not a function`. Supplying the same value directly as `--viewport 1280x720` works. Passing `--viewport 800x600` alongside the config also fails rather than overriding the configured value.

Similarly, a config with `wsEndpoint` and string-encoded `wsHeaders` fails with `Invalid JSON for wsHeaders: "[object Object]" is not valid JSON`.

The config callback in `src/config/mcp-options.ts` validates the JSON through a nested yargs parser and returns its parsed result to the outer parser. The inner parser has already converted `viewport` and `wsHeaders` into objects; the outer parser then attempts their string-based coercion again.

A small fix can keep the strict inner validation, but return the original config values for final CLI/config merging and coercion. This preserves the existing key restrictions and validation behavior.

I verified this with the repository's CLI tests: adding cases for viewport, wsHeaders, and viewport CLI overrides produces three failures on the original implementation. With the fix, all 42 CLI tests pass, including checks for invalid viewport values, unknown keys, dashed/no-prefixed keys, conflicting options, and missing wsEndpoint.

Would you be open to a focused PR for this?

### Reproduction

_No response_

### Expectation

_No response_

### MCP configuration

_No response_

main at 17c671899b557d63e76e21f18af323616dd2ee8e (package version 1.9.0)

### Chrome version

_No response_

### Coding agent version

_No response_

### Model version

_No response_

### Chat log

_No response_

### Node version

v24.18.0

### Operating system

None

### Extra checklist

dajiaohuang:
Fixes #2746. Returning the inner yargs parser result applies option coercers twice, so JSON-configured viewport and wsHeaders fail. Returning the original JSON values after strict validation lets the outer parser coerce them once and preserves CLI overrides. Tests: npm run test -- tests/cli.test.ts and npm run check-format pass. The full npm run test reaches five unrelated PWA failures because the test server cannot bind 0.0.0.0:10236 with EACCES on this Windows host.


Lightning00Blade (approved):
Thanks!

PR #2748, opened 2026-09-15
docs: update page routing option name

dajiaohuang:
The concurrent sessions guide documented a removed option. Update it to the current page routing option, explain that routing is enabled by default, and document the selected-page-only opt-out. Tests: npm run check-format.


docs/advanced-usage.md
```diff
@@ -3,21 +3,17 @@
...
 {
   "mcpServers": {
     "chrome-devtools": {
       "command": "npx",
-      "args": [
-        "-y",
-        "chrome-devtools-mcp@latest",
-        "--experimentalPageIdRouting"
-      ]
+      "args": ["-y", "chrome-devtools-mcp@latest", "--pageIdRouting"]
```
samiyac, line 16: Since `pageIdRouting` is enabled by default, we don't need to pass `"--pageIdRouting"` as `args`

samiyac (changes requested):

docs/advanced-usage.md
OrKoN: we should update docs/advanced-usage.md

dajiaohuang: Updated docs/advanced-usage.md in fa222ff to rely on the default pageIdRouting behavior, so the example no longer passes --pageIdRouting. npm run check-format and npm run build both pass.

samiyac (approved):

Issue #2188, opened 2026-06-08
Support http transport


natorion:
### Is your feature request related to a problem? Please describe.

Umbrella FR to track value and feasability of dedicated per-machine server with http transport.

Relevant issues: 
https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2154
https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1985
https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1232
https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/316

### Describe the solution you'd like

n/a

### Describe alternatives you've considered

proxy servers

Austin519:
Adding a concrete use case from production for the umbrella:

I run `chrome-devtools-mcp` against a long-lived Chrome on a dedicated VM (Linux, headed XFCE session, CDP on `127.0.0.1:9222`), with multiple Claude Code clients on a separate VM that all want to drive the same browser. Today the only path is per-client `ssh <vm> chrome-devtools-mcp --browserUrl http://127.0.0.1:9222` — one MCP process per Claude session, stdio over SSH.

Two failure modes this produces:

1. **Per-session orphan leak.** When a Claude session ends abnormally (autoheal kill, context-limit, OOM-killed `claude.exe`), the SSH client reparents to PID 1 but `sshd` doesn't notice the dead TCP peer for hours (default `ClientAliveInterval=0`), so the remote `chrome-devtools-mcp` Node process keeps running. Each holds ~100MB–1.7GB RSS. Over 4 days I accumulated 62 stale procs (5.4 GiB) on an 8 GB VM, which then OOM-killed Chrome itself. Mitigated with `ClientAliveInterval=60` + a code-side orphan reaper, but the structural cure would be a single long-lived MCP server.

2. **Tools/list registration race.** Short `claude --print` runs occasionally settle before the SSH+Node cold-start finishes, so `mcp__chrome-vm-cdp__*` tools never register even though `claude mcp list` shows `✓ Connected`. Mitigated by installing the binary globally (skipping `npx -y …@latest` resolution), but cold-start is still ~1.5 s — a long-lived daemon would have no cold-start at all.

[...]

PR #2733, opened 2026-09-12
feat: support shared MCP server sessions

lmn451:
## Why

Addresses https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2188. Multiple MCP clients need to drive one long-lived Chrome without spawning one Chrome-connected server per client, leaking orphaned processes, contending for profiles, or sharing selected-page and filesystem state.

## What changed

- add a loopback-only Streamable HTTP endpoint at `http://127.0.0.1:<port>/mcp` via `--http-port`
- create an independent `McpServer` and `McpContext` for every HTTP session while sharing one process-level Chrome connection and page inventory
- isolate selected pages, roots, named isolated contexts, DevTools resource loading, tracing/recording state, telemetry attribution, and per-client request mutexes
- add `--server-url` / `CHROME_DEVTOOLS_MCP_SERVER_URL` as a transparent stdio-to-HTTP proxy for stdio-only clients
- preserve existing stdio mode when neither option is supplied
- support server-launched headed/headless Chrome, `--auto-connect`, and explicit remote-debugging `--browser-url` / `--ws-endpoint` modes

## Lifecycle and security

- bind only to `127.0.0.1`; reject non-loopback `Host` and `Origin`
- cap request bodies at 4 MiB and bound pending/established sessions
- close only the addressed session on accepted `DELETE`; inactive sessions expire after 30 minutes while active requests/streams suspend expiry
- drain active tools and stop session-owned traces, screencasts, and owned isolated browser contexts before disposal
- document SSH port forwarding as the only remote-access path

## Verification

- `npm run format`
- focused shared HTTP/proxy/browser/context/telemetry run: 169 passed
- `npm run test tests/http-server.test.ts tests/proxy.test.ts tests/http-multi-client.test.ts`: 10 passed
- `npm run test -- --retry`: final attempt 1,036 passed, 3 platform skips; earlier attempts hit existing extension teardown/network ordering flakes
- live smoke: two named `chrome-devtools-axi` sessions opened separate tabs through one service, saw the same page inventory, and retained different selected pages

## Companion

AXI upstream: https://github.com/kunchenguid/chrome-devtools-axi/pull/139

Fork review PRs: https://github.com/lmn451/chrome-devtools-mcp/pull/3 and https://github.com/lmn451/chrome-devtools-axi/pull/1

Supersedes the narrower per-session-browser approach in https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2730.

PR #2750, opened 2026-09-15
feat: add multi-session Streamable HTTP transport

artgas1:
## Summary

Add a native Streamable HTTP mode for long-lived Chrome DevTools MCP processes while preserving stdio as the default.

- create one independent McpServer and McpContext per negotiated HTTP session
- reuse one concurrency-safe Puppeteer browser connection within the process
- bind only to loopback and validate Host and Origin headers
- clean up sessions on DELETE, idle expiry, and process shutdown
- keep telemetry initialization process-wide while attributing tool events per client session
- document the HTTP endpoint and deployment model

This avoids the state-sharing problem of a generic stdio-to-HTTP bridge: roots, capabilities, selected pages, trace state, and tool mutexes remain session-local. A deployment should run one process per Chrome endpoint or profile.

## Verification

- npm run gen
- npm run check-format
- npm test
- npm run test:notices
- npm run profile -- --threshold-percentage=1
- integration tests cover two isolated HTTP clients sharing one browser, the real executable lifecycle, Host and Origin rejection, malformed JSON, explicit and idle cleanup, and concurrent first browser launch

Addresses #2188. Internal tracking: INFRA-1236.


artgas1:
Closing this duplicate in favor of #2733, which already implements the same per-session server architecture with broader lifecycle and proxy coverage. Internal tracking: INFRA-1236.

PR #2766, opened 2026-09-18
fix(daemon): keep fallback socket in private runtime dir

mcc0nnell:
## Summary

- keep the non-XDG Unix daemon socket inside the daemon's existing private runtime directory
- reject symlinked runtime directories before trusting ownership or permissions
- preserve the current XDG and Windows paths
- add regression coverage for both fallback placement and runtime-directory symlink tampering

## Why

The daemon already creates and validates getRuntimeHome(sessionId) as a user-owned, non-group/world-writable directory. However, when XDG_RUNTIME_DIR is unset, getSocketPath() currently binds the Unix socket directly in /tmp.

Node's readableAll: false / writableAll: false do not force an owner-only Unix socket. The socket mode follows the daemon's umask. In testing:

- 0022 -> 0755
- 0002 -> 0775
- 0000 -> 0777

Because the daemon protocol does not perform a separate peer-authentication handshake, a socket writable by another local account becomes a cross-user IPC boundary.

Putting server.sock under the already-validated private runtime directory makes the parent directory the access-control boundary and avoids relying on socket mode.

The runtime directory itself is predictable in the /tmp fallback. The prior validation used statSync(), which follows symbolic links. This PR therefore also switches that check to lstatSync() and rejects symlinked runtime directories before checking ownership. That prevents a pre-created /tmp symlink from redirecting daemon.pid or server.sock operations into another directory.

## Verification

Real-daemon before/after under umask 0000, with XDG_RUNTIME_DIR unset:

Before:

```text
mode=777 uid=0 gid=0 path=/tmp/chrome-devtools-mcp-0.sock
```

An unprivileged UID could connect and issue a harmless status request.

After:

```text
parent mode=700 uid=0 gid=0 path=/tmp/chrome-devtools-mcp-0
socket mode=777 uid=0 gid=0 path=/tmp/chrome-devtools-mcp-0/server.sock
```

The same cross-UID connection fails with EACCES.

Also verified:

- daemon symlink security tests
- daemon utils tests
- ESLint / Prettier on changed files
- TypeScript build/typecheck
- git diff --check

Security context has been reported through Google OSS VRP as Issue 563280592.


OrKoN (changes requested):
Thanks for the PR! I think we should do the following changes:

- Guard `fs.unlinkSync(pidFilePath)` in `startDaemon` (`src/daemon/client.ts`, lines 122–124) with `fs.lstatSync` checks on `path.dirname(pidFilePath)` and `pidFilePath` to reject symbolic links before unlinking. Because the CLI invokes `startDaemon` before spawning `daemon.ts`, a pre-created symlink at `pidDir` currently causes `startDaemon` to follow the symlink and delete `daemon.pid` in the target directory before `daemon.ts`'s new `lstatSync(pidDir)` check ever runs.
- Simplify `getSocketPath` in `src/daemon/utils.ts` (lines 50–59) to return `path.join(getRuntimeHome(sessionId), 'server.sock')` directly for all non-Windows platforms, matching `getPidFilePath`. `getRuntimeHome(sessionId)` already returns `path.join(process.env.XDG_RUNTIME_DIR, appName)` when `XDG_RUNTIME_DIR` is set, making the separate `if (process.env.XDG_RUNTIME_DIR)` branch in `getSocketPath` redundant.

PR #2774, opened 2026-09-18
perf: remove daemon pid-file polling

mediumWellness:
## Summary

Remove redundant pid-file polling and filesystem watchers from the daemon lifecycle. Startup already has an authoritative socket readiness loop, and shutdown can clean up the pid file directly after the stop command completes.

This avoids repeated filesystem polling, removes watcher/timer bookkeeping, and eliminates up to 500 ms of avoidable lifecycle latency.

## Performance opportunities identified

1. Remove daemon pid-file polling/watching and rely on socket readiness plus direct cleanup (`src/daemon/client.ts`) — implemented here.
2. Replace hot-path linear MCP page lookup with a `WeakMap<Target, McpPage>`.
3. Pre-build and memoize tool flag strings instead of rebuilding them on every request.
4. Avoid `unshift` plus `splice` array churn in `PageCollector`.
5. Add an ID side-index `Map` for constant-time `getById()` lookups.
6. Cache repeated path-root resolution in `validatePath()`.
7. Cache target-filter functions by browser mode.
8. Use binary search instead of a linear scan in `bucketizeLatency()`.
9. Replace unnecessary `structuredClone()` calls with safe shallow copies in formatters.
10. Memoize recursive Zod type resolution.

## Validation

- The original implementation commit passed TypeScript typechecking.
- Revalidation in this fresh worktree could not start because dependencies are not installed: `npm run build` reports `'tsc' is not recognized as an internal or external command`.
- Diff is limited to `src/daemon/client.ts`: 3 insertions and 47 deletions.


OrKoN (changes requested):
If the goal is to eliminate the 500ms latency caused by `fs.watchFile(filePath, {interval: 500})` without breaking shutdown synchronization:
1. **Do not unlink `pidFilePath` in `stopDaemon()` while the daemon is still shutting down.** Let `daemon.ts`'s `cleanup()` remain the sole owner that unlinks `pidFilePath` (using `fs.rmSync(pidFilePath, {force: true})`) after closing `mcpClient`, `mcpTransport`, `server`, and `socketPath`.
2. **In `stopDaemon()`**, replace `waitForFile(pidFilePath, true)` (which used a 500ms `fs.watchFile` interval) with a fast polling loop (e.g. every `20ms` up to `FILE_TIMEOUT`) waiting until `!isDaemonRunning(sessionId)` (confirming both `pidFilePath` removal and process termination), and only `fs.rmSync(pidFilePath, {force: true})` if the process has terminated without removing its PID file:
   ```ts
   export async function stopDaemon(sessionId: string) {
     if (!isDaemonRunning(sessionId)) {
       logger?.('Daemon is not running');
       return;
     }

     await sendCommand({method: 'stop'}, sessionId);

     const deadline = Date.now() + FILE_TIMEOUT;
     while (Date.now() < deadline) {
       if (!isDaemonRunning(sessionId)) {
         fs.rmSync(getPidFilePath(sessionId), {force: true});
         return;
       }
       await delay(20);
     }
     throw new Error(`Timeout: daemon did not stop within ${FILE_TIMEOUT}ms`);
   }
   ```
3. This reduces shutdown wait overhead from up to `500ms` (`fs.watchFile`) to `~20ms` while preserving full teardown ordering and restart safety across POSIX and Windows.

PR #2775, opened 2026-09-19
fix: avoid cumulative page title timeouts

dennisimoo:
Fetch page titles concurrently so the existing one-second timeouts overlap instead of adding up per unresponsive page. Reuses the current timeout and fallback, preserves output order, and skips hidden extension pages.

The mock-based regression reproduces three seconds of title-lookup delay before the fix and one second afterward. Tests cover extension visibility, out-of-order completion, and rejected lookups. This fixes cumulative title delay, not every cause of `list_pages` hangs.

Testing: `npm run test -- tests/McpResponse.pageTitles.test.ts` (3 passed) and `npm run check-format` passed. The full suite was attempted, then stopped after browser startup failures: this host has no usable Chrome sandbox and lacks stable Chrome for some tests.

AI-assisted with OpenAI Codex.


dennisimoo:
Could a maintainer approve the pending CI runs? The three regression tests and local lint/format checks pass.

PR #2787, opened 2026-09-20
refactor: extract browser in a BrowserManager class

Lightning00Blade:


src/config/browser-options.ts
```diff
@@ -121,11 +121,13 @@ export const browserOptions = {
   },
   chromeArg: {
     type: 'array',
+    string: true,
     describe:
       'Additional arguments for Chrome. Only applies when Chrome is launched by chrome-devtools-mcp.',
   },
   ignoreDefaultChromeArg: {
     type: 'array',
+    string: true,
```
OrKoN, line 130: what does string: true do?

Lightning00Blade: Resolves the Array to `Array<string>`. If not set it resolves to `Array<string|number>`

OrKoN (changes requested):
Please check these:

## Summary of Unintended Changes & Regressions

While the goal of PR #2787 is a pure refactoring to extract `src/browser.ts` into `src/BrowserManager.ts`, the 4 commits introduce several **unintended behavioral changes, race conditions, and telemetry/lifecycle regressions**:

---

### 1. Race Conditions in `BrowserManager.ensureBrowser()` vs. `BrowserManager.close()`
* **Where:** `src/BrowserManager.ts` (`ensureBrowser`, `close`, `#mutex`, `#closingCount`)
* **What changed:** Commit 3 replaced the explicit `#pendingBrowser` / `#closingPromise` coordination from Commit 2 with `#mutex = new Mutex()` and `#closingCount`.
* **Unintended behaviors:**
  1. **Queued `ensureBrowser()` calls bypass `#closingCount > 0` and return a browser that is about to be closed:**
     ```ts
     async ensureBrowser(): Promise<Browser> {
       using _guard = await this.#mutex.acquire();
       if (this.#browser?.connected) {
         return this.#browser;
       }
       const browser = await this.#initBrowser();
       if (this.#closingCount > 0) {
         throw new Error('Browser was closed while initializing.');
       }
       return browser;
     }
     ```
     Suppose Call 1 (`ensureBrowser()`) is awaiting `#initBrowser()`, Call 2 (`ensureBrowser()`) queues on `#mutex`, and then Call 3 (`close()`) increments `#closingCount` to `1` and queues on `#mutex` behind Call 2:
     * Call 1 finishes `#launch()`, sets `this.#browser = browser`, checks `this.#closingCount > 0`, and throws `'Browser was closed while initializing.'`.
     * Call 2 acquires `#mutex` **before** Call 3 (`close()`). Call 2 checks `if (this.#browser?.connected)` — which is `true` because Call 3 hasn't acquired the mutex yet! — and **returns `this.#browser` without checking `this.#closingCount > 0`**.
     * Immediately afterward, Call 3 (`close()`) acquires `#mutex` and closes `this.#browser` while Call 2 is actively using it.
  2. **`ensureBrowser()` rejects *before* `browser.close()` has actually executed:**
     When `close()` is called mid-launch, `close()` increments `#closingCount` and blocks on `#mutex.acquire()`. `ensureBrowser()` throws `'Browser was closed while initializing.'` while still holding `#mutex`—meaning at the time `ensureBrowser()` rejects, `close()` has **not yet run** and the Chrome process is still alive.

---

### 2. Reversed Disposal Order in `McpServer.close()`
* **Where:** `src/index.ts` (`McpServer.close`)
* **What changed:**
  * **Before:**
    ```ts
    async close(): Promise<void> {
      this.#context?.dispose();
      this.#context = undefined;
      await this.server.close();
    }
    ```
  * **After:**
    ```ts
    async close(): Promise<void> {
      await Promise.allSettled([
        this.#browserManager.close(),
        this.server.close(),
      ]);
      this.#context?.dispose();
      this.#context = undefined;
    }
    ```
* **Unintended behavior:**
  1. Because `this.#browserManager.close()` now runs **before** `this.#context?.dispose()`, `McpContext` and `ServiceWorkerConsoleCollector` still have their `targetdestroyed` listeners attached to `browser` while Chrome is closing all targets, triggering unnecessary per-target teardown handlers during shutdown.
  2. When `this.#context?.dispose()` finally runs, the browser connection/pipe is already closed, causing `devtoolsUniverse?.session.detach()` in `McpPage.dispose()` and `this.#session?.detach()` in `ServiceWorkerSubscriber.unsubscribe()` to send CDP commands over a dead connection and log spurious `'Failed to detach DevTools session'` errors.

---

### 3. Telemetry Regression from Adding `default: false` to `experimentalDevtools`
* **Where:** `src/config/mcp-options.ts` (`experimentalDevtools`) & `src/telemetry/flagUtils.ts` (`computeFlagUsage`)
* **What changed:** Commit 3 added `default: false` to `experimentalDevtools` in `src/config/mcp-options.ts` to avoid `?? false` in TypeScript.
* **Unintended behavior:**
  In `src/telemetry/flagUtils.ts`:
  ```ts
  if (!('default' in config) || value !== config.default) {
    usage[`${snakeCaseName}_present`] = value !== undefined && value !== null;
  }
  if (config.type === 'boolean' && typeof value === 'boolean') {
    usage[snakeCaseName] = value;
  }
  ```
  * **Before:** When `--experimentalDevtools` was omitted by the user, `args.experimentalDevtools` was `undefined`. `computeFlagUsage` logged `experimental_devtools_present: false` and left `experimental_devtools` as `undefined` (matching other experimental flags like `experimentalVision`, `experimentalScreencast`, `experimentalIncludeAllPages`).
  * **After:** Because `default: false` is now in `mcpOptions`, Yargs populates `args.experimentalDevtools = false` on every run. `computeFlagUsage` now sets `experimental_devtools_present` to `undefined` and logs `experimental_devtools: false` on every server startup even when the user never passed the flag.

---

### 4. `enableExtensions` Now Passed to `puppeteer.connect()`
* **Where:** `src/BrowserManager.ts` (`#connect`) vs. `src/index.ts` (previously `ensureBrowserConnected`)
* **What changed:**
  * **Before (`src/index.ts`):** `ensureBrowserConnected({...})` did **not** pass `enableExtensions: this.#serverArgs.categoryExtensions` (only `ensureBrowserLaunched` passed `enableExtensions`).
  * **After (`src/BrowserManager.ts`):** `#connect()` extracts `categoryExtensions: enableExtensions` from `this.#serverArgs` and passes it to `BrowserManager.makeTargetFilter(enableExtensions)`.
* **Unintended behavior:** When connecting to an existing browser (`--browserUrl`, `--wsEndpoint`, or `--autoConnect`) with `--categoryExtensions`, `targetFilter` now allows `chrome-extension://` targets, whereas previously `makeTargetFilter` received `undefined` (`false`) in `ensureBrowserConnected`.

---

### 5. Delayed Signal & `stdin` Shutdown Listener Registration
* **Where:** `src/bin/chrome-devtools-mcp-main.ts`
* **What changed:** `const server = await McpServer.from(args, {logFile});` was moved **above** `process.stdin.on('end' | 'close', ...)` and `process.on('SIGTERM' | 'SIGINT' | 'SIGHUP', ...)`.
* **Unintended behavior:**
  1. `McpServer.from()` is asynchronous (it awaits `loadIssueDescriptions()`). If `stdin` closes immediately or `SIGTERM`/`SIGINT`/`SIGHUP` arrives while `McpServer.from()` is still initializing, the shutdown listeners are not yet attached.
  2. In `shutdown()`, `await closeBrowser()` was replaced with `await server.close()`, which now also awaits `this.server.close()` (MCP transport teardown) and `this.#context?.dispose()` inside the 5-second forced-exit timer.

---

### 6. Process-Wide Singleton Browser Replaced with Per-`McpServer` Instance State
* **Where:** `src/browser.ts` -> `src/BrowserManager.ts`
* **What changed:** Previously, `src/browser.ts` held a module-scoped singleton `let browser: Browser | undefined`. Now each `McpServer` instance creates its own `new BrowserManager(...)`.
* **Unintended behavior:** If two `McpServer` instances are created within the same Node process without `isolated: true`, the second instance will no longer reuse the connected `browser` singleton and will instead attempt to launch a second Chrome instance using the same default profile directory (`~/.cache/chrome-devtools-mcp/chrome-profile`), failing with `The browser is already running for ...`.

Lightning00Blade:
Addressed 1, 2, 4, 5.
For 3, we already to that for other places, we want to get the setup we should resolve the telemetry earlier.
For 6, now the server requires a BrowserManager instance, so one can be re-shared if needed, but no singleton is created.

OrKoN:
## Remaining & Newly Introduced Issues in Latest (`89611db4b`)

### 1. [HIGH] Breaking Public API Change in `createMcpServer()` and `McpServer.from()`
* **Where:** [`src/index.ts`](file:///Users/alexrudenko/src/chrome-devtools-mcp/src/index.ts) (`McpServerOptions`, `McpServer.from`, `createMcpServer`)
* **What changed in Commit 6 (`89611db4b`):**
  ```ts
  export interface McpServerOptions {
    browserManager: BrowserManager;
  }
  ...
  static async from(
    serverArgs: ParsedArguments,
    options: McpServerOptions,
  ): Promise<McpServer> { ... }

  /**
   * Creates and initializes a Chrome DevTools MCP server instance.
   *
   * Maintained as a public API for backwards compatibility because external
   * consumers and integrations rely on `createMcpServer()`. For new code,
   * prefer using `McpServer.from(serverArgs, options)`.
   */
  export async function createMcpServer(
    serverArgs: ParsedArguments,
    options: McpServerOptions,
  ): Promise<{server: SdkMcpServer}> {
    const server = await McpServer.from(serverArgs, options);
    return {server: server.server};
  }
  ```
* **Impact:**
  1. As documented in the JSDoc above `createMcpServer`, external consumers and integrations call `createMcpServer(serverArgs)` or `createMcpServer(serverArgs, {logFile})` (and `McpServer.from(serverArgs, {logFile})`).
  2. Making `options: {browserManager: BrowserManager}` a required parameter and removing `logFile?: fs.WriteStream` from `McpServerOptions` is a breaking runtime and TypeScript API change: calling `createMcpServer(serverArgs)` without a second argument now throws `TypeError: Cannot read properties of undefined (reading 'browserManager')`.
* **Recommendation:**
  Keep `browserManager` and `logFile` optional in `McpServerOptions` and keep `options: McpServerOptions = {}` defaulted in `McpServer.from` and `createMcpServer`:
  ```ts
  export interface McpServerOptions {
    browserManager?: BrowserManager;
    logFile?: fs.WriteStream;
  }
  ```
  And inside `McpServer`'s constructor:
  ```ts
  this.#browserManager =
    options.browserManager ?? new BrowserManager(serverArgs, options);
  ```

---

### 2. [MEDIUM] `BrowserManager.#ensureBrowserLocked()` Still Calls `#initBrowser()` Before Checking `this.#closingCount > 0`
* **Where:** `src/BrowserManager.ts` (`#ensureBrowserLocked`, `close`)
* **Current code in Commit 6 (`89611db4b`):**
  ```ts
  async #ensureBrowserLocked(): Promise<Browser> {
    using _guard = await this.#mutex.acquire();
    if (!this.#browser?.connected) {
      await this.#initBrowser();
    }
    if (this.#closingCount > 0 || !this.#browser) {
      await this.#closeBrowser();
      throw new Error('Browser was closed while initializing.');
    }
    return this.#browser;
  }

  async close(): Promise<void> {
    this.#initPromise = undefined;
    this.#closingCount++;
    using _guard = await this.#mutex.acquire();
    try {
      await this.#closeBrowser();
    } finally {
      this.#closingCount--;
    }
  }
  ```
* **Unintended behaviors:**
  1. **Launches Chrome when a close is already queued:**
     Because `close()` clears `this.#initPromise = undefined` before acquiring `#mutex`, a subsequent `ensureBrowser()` call made while `#mutex` is busy will start a new `#ensureBrowserLocked()` call and queue on `#mutex`. If `#closingCount > 0` when `#ensureBrowserLocked()` acquires `#mutex` (e.g., `ensureBrowser()` #1 -> `close()` #1 -> `ensureBrowser()` #2 -> `close()` #2), `#ensureBrowserLocked()` checks `if (!this.#browser?.connected)` **before** checking `if (this.#closingCount > 0)`. It will **launch a new Chrome process via `await this.#initBrowser()`**, only to immediately see `this.#closingCount > 0` on the next line, close the newly launched browser, and throw `'Browser was closed while initializing.'`.
  2. **`ensureBrowser()` called while `close()` is tearing down an active browser will relaunch Chrome instead of rejecting:**
     If a browser is already running (`#initPromise === undefined`) and `close()` is called, `close()` increments `#closingCount` to `1`, acquires `#mutex`, and awaits `browser.close()`. If `ensureBrowser()` is called while `browser.close()` is in flight, `ensureBrowser()` waits on `this.#mutex.acquire()`. When `browser.close()` completes, `close()` runs `finally { this.#closingCount--; }` (setting `#closingCount` back to `0`) *before* `_guard` is disposed at block exit. When `#mutex` is released to the waiting `ensureBrowser()`, `#closingCount` is already `0`, so `ensureBrowser()` launches a brand-new browser instead of rejecting.
* **Recommendation:**
  Check `if (this.#closingCount > 0)` at the beginning of `ensureBrowser()` (or `#ensureBrowserLocked()` before `this.#mutex.acquire()` and again after acquiring `#mutex` before `this.#initBrowser()`):
  ```ts
  async #ensureBrowserLocked(): Promise<Browser> {
    if (this.#closingCount > 0) {
      throw new Error('Browser was closed while initializing.');
    }
    using _guard = await this.#mutex.acquire();
    if (this.#closingCount > 0) {
      throw new Error('Browser was closed while initializing.');
    }
    if (!this.#browser?.connected) {
      await this.#initBrowser();
    }
    if (this.#closingCount > 0 || !this.#browser) {
      await this.#closeBrowser();
      throw new Error('Browser was closed while initializing.');
    }
    return this.#browser;
  }
  ```

---

### 3. [LOW] Telemetry Change from `default: false` on `experimentalDevtools` Remains Unresolved
* **Where:** [`src/config/mcp-options.ts`](file:///Users/alexrudenko/src/chrome-devtools-mcp/src/config/mcp-options.ts) (`experimentalDevtools`), [`tests/cli.test.ts`](file:///Users/alexrudenko/src/chrome-devtools-mcp/tests/cli.test.ts), & [`src/telemetry/flagUtils.ts`](file:///Users/alexrudenko/src/chrome-devtools-mcp/src/telemetry/flagUtils.ts) (`computeFlagUsage`)
* **What happened:**
  Commit 3 added `default: false` to `experimentalDevtools` in `src/config/mcp-options.ts`, and Commit 5 (`a265c2fcc`) updated `tests/cli.test.ts` to include `experimentalDevtools: false` in the expected default parsed arguments.
* **Impact:**
  In [`src/telemetry/flagUtils.ts`](file:///Users/alexrudenko/src/chrome-devtools-mcp/src/telemetry/flagUtils.ts#L31-L40):
  ```ts
  if (!('default' in config) || value !== config.default) {
    usage[`${snakeCaseName}_present`] = value !== undefined && value !== null;
  }
  if (config.type === 'boolean' && typeof value === 'boolean') {
    usage[snakeCaseName] = value;
  }
  ```
  Unlike the other experimental boolean flags (`experimentalVision`, `experimentalScreencast`, `experimentalIncludeAllPages`), which omit `default` so that `computeFlagUsage` reports `*_present: false` and `*: undefined` when not specified by the user, `experimentalDevtools` now has `default: false`. This causes `computeFlagUsage` to omit `experimental_devtools_present` (`undefined`) and always report `experimental_devtools: false` on every run.
* **Recommendation:**
  Remove `default: false` from `experimentalDevtools` in `src/config/mcp-options.ts` (and revert `tests/cli.test.ts`), and use `Boolean(this.#serverArgs.experimentalDevtools)` (or `this.#serverArgs.experimentalDevtools ?? false`) in `BrowserManager` and `McpServer`.


Let's revert changes to the experimental_devtools defaults. Even if we happened to other flags before we do not want to spike usage for this flag and we should not make this change during refactoring. Let's file an issue to address this as a follow-up (I think it is the same issue that yargs things default values are actually set by the user).

OrKoN (approved):

PR #2791, opened 2026-09-21
feat: Add analyze_heapsnapshot_contexts MCP tool

dinfuehr:


tests/formatters/HeapSnapshotFormatter.test.ts
```diff
@@ -597,4 +602,105 @@ describe('HeapSnapshotFormatter', () => {
...
+        '',
+        `#### Context @102 in \`createClosure (test.js)\` — ${formatBytesToKb(500)}`,
+        'Script @301, ScopeInfo @303, offsets 14-104, 2 context fields',
+        `- \`alsoDead\` (Array @204) — ${formatBytesToKb(500)}`,
+        '',
+        'Scripts without scope metadata could not be analyzed:',
+        '- `other.js` (@401) — 2 contexts',
+      ].join('\n');
+
+      assert.strictEqual(result, expected);
```
Lightning00Blade: Should we use a snapshot here to make it easier to view changes?

src/tools/memory.ts
```diff
@@ -533,3 +533,55 @@ export const queryHeapSnapshotObjects = defineTool(() => ({
     });
   },
 }));
+
+export const analyzeHeapSnapshotContexts = defineTool(() => ({
+  name: 'analyze_heapsnapshot_contexts',
+  description:
+    'Loads a memory heapsnapshot and reports JavaScript closure contexts containing dead captured fields, i.e. fields that no live closure can read anymore. Contexts are ranked across all scopes by the retained size of the values held in those fields, and each one is reported together with the scope that declares it. Returns 20 contexts per page by default. Retained size is a ranking heuristic, not the number of bytes that would be reclaimed.',
```
Lightning00Blade, line 540: Gemini suggested we can change. But for sure we should remove `Returns 20 contexts per page by default`.
```suggestion
    'Loads a memory heapsnapshot to identify and rank closure contexts holding dead captured fields—variables no remaining live closure can read. Scopes are ranked globally by the retained size of these dead values to provide a prioritizing heuristic, rather than an exact measure of reclaimable bytes.,
```

Lightning00Blade (approved):

dinfuehr:
Thanks a lot for the review! I've updated the PR, so PTAL again

tests/formatters/HeapSnapshotFormatter.test.ts
dinfuehr: Good point, thanks!

src/tools/memory.ts
dinfuehr: Makes sense, done!

Lightning00Blade (approved):
Still LGTM!

PR #2794, opened 2026-09-21
fix: handle JavaScript dialogs opened during input tool actions

Lightning00Blade:
Locator actions time out while the page is blocked by a JavaScript dialog, even though the action was dispatched and opened that dialog. That surfaced as an error contradicting the "Call handle_dialog to handle it before continuing" hint in the same response. These are now reported as an interruption instead of a failure, and `fill_form` stops rather than trying to fill the remaining elements against a blocked page.


src/tools/input.ts
```diff
@@ -418,21 +465,21 @@ export const fillForm = definePageTool(() => ({
...
+  handler: async (request, response) => {
+    const {result, interruptedUid} = await fillFormElements(
+      request.params.elements,
+      request.page,
+    );
+    if (!result) {
+      // The page is blocked by a dialog, so the remaining elements cannot be
+      // filled out until it is handled.
+      response.appendResponseLine(
+        `Filling out the element with uid ${interruptedUid} opened a dialog. The remaining elements were not filled out.`,
```
OrKoN: Only append `"The remaining elements were not filled out."` when `interruptedUid` is not the last element in `request.params.elements`. When the last (or only) element in `fill_form` opens a dialog, all requested elements have already been filled, so claiming remaining elements were skipped misleads the caller into re-filling the form.

src/tools/input.ts
```diff
@@ -244,46 +271,65 @@ function hasOptionChildren(aXNode: TextSnapshotNode) {
...
-      });
-
-      if (isToggle) {
-        if (['true', 'false'].includes(value)) {
-          await handle.asLocator().fill(value === 'true');
+    const result = await page.waitForEventsAfterAction(async signal => {
+      for (const {uid, value} of elements) {
+        currentUid = undefined;
+        using handle = await page.getElementByUid(uid);
+        currentUid = uid;
```
OrKoN, line 330: call `signal.throwIfAborted()` at the end of each loop iteration (while `currentUid` still points to the element just filled) and guard `handle.evaluate` / `selectOption` against `signal`. If filling an element opens a dialog just as `locator.fill` resolves (or on `blur`/microtask), the loop otherwise advances to the next element, overwrites `currentUid`, and hangs until `protocolTimeout` (180s) inside `handle.evaluate` or `childHandle.getProperty('value')` on the paused renderer.

src/utils/WaitForHelper.ts
OrKoN, line 102: return `Promise.resolve()` immediately if `this.#abortController.signal.aborted` is already `true` before calling `setTimeout`. Because `dialogHandler` now aborts `this.#abortController` during `action()`, any action that resolves without throwing on `signal` (such as `click_at`, `press_key`, `type_text`, or a `beforeunload` dialog after navigation starts) reaches `this.timeout(...)` with an already-aborted signal where `addEventListener('abort', ...)` never fires, leaking the timer or waiting the full navigation timeout.

OrKoN (approved):
LGTM thanks!

PR #2795, opened 2026-09-21
fix: set emulatedUserAgent and use finalDisplayedUrl in lighthouse_audit

Lightning00Blade:
`lighthouse_audit` never set `emulatedUserAgent`, so desktop audits ran with Lighthouse's default mobile UA. It now mirrors the UA constants Lighthouse's own presets use. It also reported `URL: undefined` in snapshot mode, because `mainDocumentUrl` is only populated for navigations; it now uses `finalDisplayedUrl`.


OrKoN (approved):
Nice!

PR #2796, opened 2026-09-21
fix: reject --blockedUrlPattern/--allowedUrlPattern hostname regexp groups

### LINKED ISSUE #2777: blockedUrlPattern: hostname patterns with a regexp group are enforced for direct navigation but not for redirects
### Description of the bug

A `--blocked-url-pattern` whose hostname uses a URLPattern regexp group is accepted silently, and `navigate_page` refuses a direct navigation that matches it. But when an allowed page answers with an HTTP redirect to a URL matching the same pattern, the redirect is followed and the blocked page loads. The same pattern written with an exact hostname is enforced in both cases.

So a user who writes a range (for example all of 127.0.0.0/8) gets protection for direct navigation only, with no warning that redirects are not covered.

### Measured results

| Pattern | Direct navigation | Redirect from an allowed page |
|---|---|---|
| `*://127.0.0.1:*/*` (exact host) | refused | refused |
| `*://(127\.\d+\.\d+\.\d+):*/*` (regexp group) | refused | **loads** |

`new URLPattern("*://(127\\.\\d+\\.\\d+\\.\\d+):*/*").test("http://127.0.0.1:8000/")` is `true` in Node 24, so the pattern itself is valid and matches.

Same result on chrome-devtools-mcp 1.9.0 (npm) and on a source build of main at 882f93e (2026-09-18).

### Notes

- Possibly related but different code path: #2767 (URL policy across redirects in the Node-side resource loader).
- Separate observation, not a bug in this project: a partial wildcard on a numeric host such as `*://127.*:*/*` matches nothing at all, because URLPattern canonicalizes `127.*` as an IPv4 address (`new URLPattern("*://127.*:*/*").hostname` is `0.0.0.127*`). A sentence in the option's documentation would save users from that one too.

### Reproduction

1. Serve any page locally: `python3 -m http.server 8000 --bind 127.0.0.1`
2. Start the server with: `--isolated --headless --blocked-url-pattern "*://(127\.\d+\.\d+\.\d+):*/*"`
3. `navigate_page` to `http://127.0.0.1:8000/`
   Result: "Unable to navigate in the selected page: Navigation to http://127.0.0.1:8000/ is blocked by blocklist/allowlist rules." The local server logs no request. (Correct.)
4. `navigate_page` to `https://httpbin.org/redirect-to?url=http%3A%2F%2F127.0.0.1%3A8000%2F`
   Result: "Successfully navigated", `location.href` is `http://127.0.0.1:8000/`, and the local server logs the request. (Unexpected.)
5. Restart with `--blocked-url-pattern "*://127.0.0.1:*/*"` and repeat step 4.
   Result: `net::ERR_INTERNET_DISCONNECTED`, the page stays on the error page, and the local server logs no request. (Correct.)

### Expectation

Either the regexp-group pattern is enforced on redirects and subresources as it is on direct navigation, or the server rejects (or warns about) patterns that the network-level blocking cannot enforce, and the option's documentation says which pattern forms are supported.

### MCP configuration

"args": ["-y", "chrome-devtools-mcp@1.9.0", "--isolated", "--headless", "--blocked-url-pattern", "*://(127\\.\\d+\\.\\d+\\.\\d+):*/*"]

1.9.0 (npm); also main at 882f93e

### Chrome version

153.0.8010.48

### Coding agent version

n/a (reproduced with a plain MCP SDK client)

### Model version

_No response_

### Chat log

_No response_

### Node version

v24.21.0

### Operating system

_No response_

### Extra checklist

Ashfaqbs:
## Problem

`--blockedUrlPattern`/`--allowedUrlPattern` accept any URLPattern string, including hostnames that use a regexp group (for example `*://(127\.\d+\.\d+\.\d+):*/*`, meant to cover a whole IP range). That pattern is enforced correctly by the target-attach check (`TargetManager#isUrlAllowed`, using the real `URLPattern.test()`), but the actual network-level blocking runs through the CDP command `Network.emulateNetworkConditionsByRule`, passing the raw pattern string as `NetworkConditions.urlPattern`. That command's native matching does not apply the pattern consistently on redirects, so a page allowed to load can redirect straight through a blocked host range while an exact-hostname pattern stays blocked in both cases.

Root-caused and written up in more detail on #2777.

This isn't something fixable from this repo's side (the mismatch is between the documented CDP `urlPattern` semantics and the underlying Chromium implementation for `emulateNetworkConditionsByRule`, not in `puppeteer-core` or `chrome-devtools-mcp`), so instead of leaving the gap silent, this rejects patterns this repo can't currently guarantee are enforced.

## Fix

- Added `findUnenforceableHostnamePattern` (`src/utils/url.ts`), which parses each pattern with `URLPattern` and flags one whose canonicalized `hostname` contains a regexp group (`(`). Wildcard (`*`) and named-group (`:name`) hostnames are unaffected and continue to work as before; a regexp group outside the hostname (e.g. in the pathname) is also left alone, since only the hostname case is demonstrated as broken.
- Wired it into the `blockedUrlPattern`/`allowedUrlPattern` CLI option `coerce`, so an offending pattern fails fast at startup with a clear error instead of silently only half-working.
- Updated both options' `describe` text and regenerated `docs/configuration.md` via `npm run gen`.

## Testing

- Added unit tests in `tests/utils/url.test.ts` covering: a hostname regexp group (flagged), multiple patterns (first offender returned), an exact hostname, a wildcard hostname, a named-group hostname, a regexp group outside the hostname, and a pattern that fails to construct.
- `npx tsc` and `npx eslint` both pass clean on the changed files.
- Verified the new function's behavior against Puppeteer's own vendored `URLPattern` polyfill (`node_modules/puppeteer-core/lib/third_party/urlpattern-polyfill`), matching every case in the new test suite - my local Node (22.22) predates Node's global `URLPattern` support that this repo's `.nvmrc` (v24) assumes, so I couldn't run the built test file directly against the global here, but the polyfill is the same spec implementation and gave identical results for all cases.

Fixes #2777


OrKoN:
@nattallius could you please review?

src/utils/url.ts
```diff
@@ -121,6 +121,36 @@ const DISALLOWED_PROTOCOLS = new Set(['javascript:', 'data:', 'vbscript:']);
...
+  for (const raw of patterns) {
+    let parsed: URLPattern;
+    try {
+      parsed = new URLPattern(raw);
+    } catch {
+      // Invalid patterns are reported by URLPattern's own constructor error
+      // when chrome-devtools-mcp starts; this check only looks at valid ones.
+      continue;
+    }
+    if (parsed.hostname.includes('(')) {
```
OrKoN: can we use https://developer.mozilla.org/en-US/docs/Web/API/URLPattern/hasRegExpGroups to detect regex groups?

Ashfaqbs: Good call — pushed a commit using `URLPattern.hasRegExpGroups` instead. Isolated the hostname into its own `URLPattern({hostname: ...})` first so it only reflects a regexp group in the hostname, keeping the existing scope (a regexp group elsewhere in the pattern is still intentionally left alone, per the existing test for that). Verified the full behavior matrix against the spec-identical `urlpattern-polyfill` puppeteer-core ships, since this Node install doesn't have the native global.

OrKoN (changes requested):
Let's make the following changes:

- Reject any `URLPattern` where `parsed.hasRegExpGroups` is true (and do not swallow `new URLPattern(raw)` syntax errors) in `src/utils/url.ts`, `src/config/mcp-options.ts`, `docs/configuration.md`, and `tests/utils/url.test.ts` instead of isolating `new URLPattern({hostname: parsed.hostname})`. Chromium's `ThrottlingController` compiles `Network.emulateNetworkConditionsByRule` patterns with `SimpleUrlPatternMatcher::Component::Create`, which rejects `HasRegexGroups()` across all URL components (`protocol`, `username`, `password`, `hostname`, `port`, `pathname`, `search`, `hash`) and silently drops the rule (`if (!maybe_pattern_matcher.has_value()) continue;`).
- Move `findUnenforceableHostnamePattern` (or the renamed helper) and its JSDoc block outside the gap between `validateUrl`'s JSDoc block and `export function validateUrl` in `src/utils/url.ts`. Inserting the new function at line 124 detached `validateUrl`'s JSDoc comment (lines 115–123) from `export function validateUrl`.
- Add CLI option tests in `tests/cli.test.ts` verifying that `parseArguments` rejects `--blocked-url-pattern` and `--allowed-url-pattern` values containing regexp groups. Currently only the helper in `tests/utils/url.test.ts` is tested directly, leaving the `coerce` callbacks in `src/config/mcp-options.ts` untested.

OrKoN (approved):

PR #2797, opened 2026-09-21
fix(cli): keep WebSocket credentials out of daemon argv

mcc0nnell:
`--wsHeaders` currently does not survive the `chrome-devtools` daemon path: yargs parses the JSON into an object, then the generic daemon serializer converts it to `[object Object]`, causing the daemon parser to reject it.

The obvious fix would be to JSON-stringify the value into the child argv, but that would also expose authentication headers through process inspection.

This change instead removes resolved WebSocket headers from daemon argv and carries them through the inherited environment. It preserves the existing direct CLI behavior and precedence, documents the safer non-argv path, and adds regression coverage ensuring an Authorization canary is absent from serialized daemon arguments.

Validation:
- `npm run typecheck`
- `npm run test -- tests/cli.test.ts tests/daemon/utils.test.ts`


OrKoN (changes requested):
let's fix the serilization issue without adding the env variable support. Could the `--config` option documented here https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/docs/configuration.md be used instead (it will be included in the next release https://github.com/ChromeDevTools/chrome-devtools-mcp/pull/2691)

mcc0nnell:
Updated per your suggestion — `--config` is the transport now.

I rebased onto current `main` and removed the environment-variable support entirely. The revised path does two things:

- config-backed `wsHeaders`: keep `--config` in the daemon argv and omit the already-resolved header value, so the daemon rereads the config;
- explicit `--wsHeaders`: keep the existing command-line semantics and JSON-serialize the parsed object correctly, including when it overrides a config value.

I also narrowed the security claim to just avoiding an unnecessary copy of config-held credentials into the daemon process argv. No claim that the config file itself is a secret store.

`npm run typecheck` and the focused CLI/daemon tests pass, including the Authorization canary regression.

OrKoN (changes requested):
Let's make these changes:

- Serialize object values (such as `wsHeaders`) directly inside `serializeArgs` in `src/daemon/utils.ts` (`typeof value === 'object' ? JSON.stringify(value) : String(value)`) and remove `serializeArgsForDaemon` and `hasExplicitWsHeaders`. Because `src/bin/chrome-devtools.ts` does not register Yargs `.config()`, `argv.wsHeaders` is never populated from `--config` in the CLI process, making `hasExplicitWsHeaders` dead code while leaving `serializeArgs` broken (`[object Object]`) for direct callers.
- Guard the `argv.isolated = true` and `argv.headless = true` defaults in `src/bin/chrome-devtools.ts` (lines 142–158) with `!argv.config`. When `chrome-devtools start --config=...` is used without CLI browser flags, injecting `--isolated` and `--headless` into the daemon argv overrides config values (`isolated: false`, `headless: false`) and causes a fatal Yargs conflict (`Arguments isolated and userDataDir/autoConnect are mutually exclusive`) when the daemon loads the config file.

